PR #192, opened 2026-03-08
Poc/import from dotmny

marksimpson:
**This is a proof of concept - not ready to merge**

Initial attempt at a migration pipeline. Put `source.mny` in `migration`, and run `migrate.sh`. It should take care of the rest.

This successfully imported my MS Money 2004 file, but I haven't done a full reconciliation to make sure all is correct. It looks very close though.

Submitting this for feedback and learning - I do not recommend merging it yet. I also have no idea how it will behave with .mny files from different versions.

### DIFF
--- a/docs/plans/2026-03-07-mny-import-pipeline-design.md
+++ b/docs/plans/2026-03-07-mny-import-pipeline-design.md
@@ -1,0 +1,146 @@
+# Microsoft Money Import Pipeline
+
+Import data from a Microsoft Money `.mny` file into the monize PostgreSQL database.
+
+## Pipeline
+
+Three stages, orchestrated by `migration/migrate.sh`:
+
+1. **Decrypt** -- sunriise JAR converts `.mny` to `.mdb` (downloaded automatically if missing)
+2. **Extract** -- `mdb-export` (from `mdbtools`) reads each table as CSV
+3. **Load** -- `migration/migrate.ts` transforms and inserts into Postgres via `pg`
+
+Entry point: `cd migration && ./migrate.sh`
+
+### Prerequisites
+
+- Java runtime (for sunriise)
+- `mdbtools` (`brew install mdbtools`)
+- Node.js / npm
+- Running Postgres with monize schema applied
+
+### Configuration
+
+`migration/.env`:
+
+```
+MONEY_FILE_PASSWORD=your-password-here
+MIGRATION_USER_EMAIL=user@example.com
+POSTGRES_DB=monize
+POSTGRES_USER=monize_user
+POSTGRES_PASSWORD=your-password
+POSTGRES_HOST=localhost
+POSTGRES_PORT=5432
+```
+
+## Self-contained structure
+
+Everything lives in `migration/` with its own `package.json` and `tsconfig.json`. No dependency on the NestJS backend or TypeORM. Dependencies: `pg`, `csv-parse`, `dotenv`.
+
+## Idempotency
+
+The script looks up the user by `MIGRATION_USER_EMAIL` and deletes all their existing data in FK-safe order before inserting. Safe to run repeatedly.
+
+### Deletion order
+
+1. `budget_period_categories`, `budget_periods`, `budget_alerts`, `budget_categories`, `budgets`
+2. `scheduled_transaction_overrides`, `scheduled_transaction_splits`, `scheduled_transactions`
+3. `investment_transactions`, `holdings`, `security_prices`, `securities`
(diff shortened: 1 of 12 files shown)


gerardfarrell11:
I've tested this with MSMoney 2001 and after a little help from _Claude_ it seemed to work.
I imported ~25 years of accounts and investment transactions.
I see closed accounts as closed, with a re-open option. Balances in multiple currencies seem to be correct.
Installed as a dedicated Proxmox LXC (Debian 12, 10GB disk, 2 cores, 2GB RAM) running Docker.
Notes below as they may help others.

I was initially unable to get the backend to start, which Claude identified as:

_Migration 056_monte_carlo_scenarios.sql tries to create a trigger that already exists in your database,
and the migration runner crashes instead of skipping it. It then keeps restarting and retrying in a loop._

_This happened because the migration was partially applied at some point.
The table/trigger got created, but the migration wasn't recorded as complete._

_Fix: manually mark the migration as done:_

_Connect to your postgres container and mark the migration as already applied:_

	docker exec -it monize-postgres psql -U $POSTGRES_USER -d $POSTGRES_DB

_Once inside psql, check how migrations are tracked:_

	\dt

_Look for a table like migrations, schema_migrations, or migration_history, then check what's in it:_

	SELECT * FROM migrations LIMIT 10;
	
_Then insert the missing record (the exact syntax depends on the table structure, but likely something like):_

	INSERT INTO schema_migrations (filename, applied_at) VALUES ('056_monte_carlo_scenarios.sql', NOW());

_Exit psql with \q, then restart the backend:_

	docker compose -f docker-compose.prod.yml restart backend
	docker compose -f docker-compose.prod.yml up -d frontend



I then copied my MSMoney file over to the LXC, renamed it to source.mny and ran the migration tool.

```
	cd ~/monize/migration
	npm install
	chmod +x migrate.sh
	./migrate.sh
```

This failed I was missing libatomic.so.1

```
	apt-get install -y libatomic1
	./migrate.sh
```

It got further, but was missing pg

```
	cd ~/monize/migration
	npm install pg
	./migrate.sh
```

I then had a few Postgres issues, mainly making sure the .env file was set up correctly with the docker IP address,
and my random password contained $T which Claude thought may have caused hashing issues if the shell tried to use it as variable. Once all those were figured out

`	./migrate.sh`

got further, but failed on table BILL.

Claude walked me through checking the tables and modifying the scripts:

`	nano ~/monize/migration/migrate.ts`
	
_First fix — line 347, change:_

	typescriptconst billRows = readTable(MDB_FILE, 'BILL')

_to:_

	typescriptconst billRows = tableExists(MDB_FILE, 'BILL') ? readTable(MDB_FILE, 'BILL') : []

_Second fix — line 726, same change:_

	typescriptconst billRows = tableExists(MDB_FILE, 'BILL') ? readTable(MDB_FILE, 'BILL') : []

Then:
	`nano ~/monize/migration/transform.ts`
	
_Add this function right after the closing } of readTable:_

	typescriptexport function tableExists(mdbFile: string, table: string): boolean {
	try {
		const tables = execSync(`mdb-tables "${mdbFile}"`, { maxBuffer: MAX_BUFFER }).toString('utf-8')
		return tables.split(/\s+/).includes(table)
	} catch {
		return false
	}
	}

Then:

	nano ~/monize/migration/migrate.ts
	
_Find the import line at the top that imports from transform.ts and add tableExists to it. It probably looks like:_

	typescriptimport { readTable, 
						tableExists,				<-- added this
						... } from './transform'



`	./migrate.sh`

and got:

Deleting existing data...
Ensuring currencies...
  Currency map: 45 entries
Reference data:
  Payees: 1032
  Categories: 392
  Securities: 117
Accounts:
  Accounts: 72 (6 investment-cash links)
Transactions:
  Transactions: 27584  (49 recurring phantoms, 0 orphaned transfers skipped)
  Splits: 115
  Transfers: 1740
Investments:
  Investment transactions: 281
Prices & rates:
  Security prices: 3168
  Exchange rates: 76
Scheduled transactions:
  Scheduled transactions: 0
Computing balances and holdings:
  Balances updated: 72 accounts
  Holdings: 9
Post-import cleanup:
  Deactivated past scheduled transactions: 0
Migration complete.
Import complete.

Scheduled transactions came across as 0, which is expected since Money 2001 didn't have a separate BILL table.

kenlasko:
Got this merged into Monize via #1003  Not sure how much of @marksimpson code was incorporated, but its done! Wish I did this approach at the beginning. Was much simpler to import the .mny file than the individual QIF files. Oh well!

PR #676, opened 2026-06-14
feat(payees): backfill uncategorized transactions from default category

WMP:
Setting a payee's default category only affects future transactions, so payees imported with a default category often keep large numbers of uncategorized historical transactions. The existing Auto-Assign dialog only surfaces payees whose category can be inferred from already categorized transactions, so these payees have no one-click fix.

Add three entry points on the Payees page that apply a payee's existing default category to its backfillable (uncategorized, non-transfer, non-split) transactions:

- A "No Category" status filter to find payees missing a default category.
- A clickable "N uncategorized" badge that backfills a single payee (only when it already has a default category).
- An "Apply Default Categories" toolbar button that backfills every payee that has a default category, with a confirmation showing the totals.

All three reuse the existing POST /payees/category-suggestions/apply endpoint (with backfillTransactions), so there are no backend changes. Fully internationalized across all locales with added tests.

### DIFF
--- a/frontend/src/app/payees/page.test.tsx
+++ b/frontend/src/app/payees/page.test.tsx
@@ -9,6 +9,18 @@ vi.mock('next/image', () => ({
   default: ({ priority, fill, ...props }: any) => <img alt="" {...props} />,
 }));
 
+// Callable toast mock: the bulk backfill uses the bare toast() for the
+// "nothing to do" notice, so the default export must be a function (the
+// global setup mock only exposes toast.success/.error).
+vi.mock('react-hot-toast', () => {
+  const fn: any = vi.fn();
+  fn.success = vi.fn();
+  fn.error = vi.fn();
+  fn.loading = vi.fn();
+  fn.dismiss = vi.fn();
+  return { default: fn, Toaster: () => null };
+});
+
 // Mock logger
 vi.mock('@/lib/logger', () => ({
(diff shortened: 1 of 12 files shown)


kenlasko:
It appears that we're working on the same thing at the same time. I've committed my approach to main already. I looked at yours and I still think we can use your "Apply Default Categories". Can you re-submit your PR with just that feature? That way, you will still get credit for contributing.

WMP:
Of course, please told me when you send your work to main (or to other branch), and then i will send pr with only this feature.

WMP:
Btw, i think in payee page we need dropdown menu for this eventy batch positions.

kenlasko:
All my stuff has been pushed to main

WMP:
Ready to merge, but I didn't build this commit, so I'm not sure if it works.

kenlasko:
Your PR still includes the "No Category" filter option, which isn't needed because I have a separate filter for All Payees/No Default Category/Uncategorized Txns

Can you remove the No Category option?

Also, make sure you create full tests for the added functionality

image

WMP:
Fixed, and i add tests. Build on 
```
ghcr.io/wmp/monize-backend:fork-deploy-ports-9fdf12e (and :latest)
ghcr.io/wmp/monize-frontend:fork-deploy-ports-9fdf12e (and :latest)
```

PR #679, opened 2026-06-14
feat(reports): drill into uncategorized transactions from monthly breakdown

WMP:
The monthly category breakdown already aggregates an "Uncategorized" bucket (transactions with no category), but its row, month cells, and total were rendered as plain text -- you could not click through to the underlying transactions like every other category row.

Make the Uncategorized row drillable using the existing "uncategorized" pseudo-filter the transactions list already supports (categoryId IS NULL, excluding splits and transfers) -- the same pattern the report's transfer section already uses with the "transfer" token. No backend, query, or schema changes: the report already computes the bucket and the transactions endpoint already understands the filter.


WMP:
Build and tested - ready to merge

PR #683, opened 2026-06-14
feat(auth): add OIDC_REQUIRE_VERIFIED_EMAIL to merge by IdP email without verification/SMTP

WMP:
## Problem

Monize is stricter than typical OIDC apps in two places, which together block some self-hosted setups:

1. It only trusts the OIDC provider's email when the provider asserts `email_verified: true`.
2. Merging an OIDC identity into an existing **local (password)** account requires an email-confirmation step, which needs **SMTP**.

So if your IdP does not send `email_verified` (e.g. a default Authentik OAuth2 provider) and you have no SMTP configured, an OIDC login for a user who already has a local account fails: the email isn't trusted, it falls through to user creation, and either hits the disabled-registration guard or a duplicate-email error. There is no way to link/merge that account without standing up SMTP.

## Change

Add a single environment variable, **`OIDC_REQUIRE_VERIFIED_EMAIL`** (default `true` — **no behavior change**).

Set it to `false` to:
- trust the IdP-provided email even without an `email_verified` claim, and
- merge the OIDC identity **directly** into an existing account matching that email — including a local password account — skipping the confirmation email (so it works without SMTP).

Local password login keeps working afterwards (the merge sets `oidcSubject`/`authProvider` but preserves the password; `login()` gates only on the password, not `authProvider`).

## Security

This intentionally lowers security: with it disabled, anyone who can get the IdP to issue a token for an email owns the matching Monize account. Only disable it when you trust your IdP to verify email ownership. Documented as such in `.env.example`, and the default keeps the strict, confirmation-based behavior.

## Scope / testing

- No DB or schema changes — pure config + logic in `findOrCreateOidcUser`.
- `backend/.env.example` documents the variable.
- Added unit tests: default still requires verification; `false` merges a password account directly without sending a confirmation email.
- `tsc` clean; full `auth.service`/`auth.controller` suites pass (219 tests).

Files: `backend/src/auth/auth.service.ts`, `backend/src/auth/auth.service.spec.ts`, `.env.example`.

### DIFF
--- a/backend/src/auth/auth.service.ts
+++ b/backend/src/auth/auth.service.ts
@@ -387,9 +387,21 @@ export class AuthService {
     const rawEmail = userInfo.email as string | undefined;
     // H7: Normalize email before lookups
     const email = rawEmail?.toLowerCase().trim();
-    // SECURITY: Only trust email if verified by the OIDC provider
+    // SECURITY: Only trust the IdP email if it is verified.
+    // OIDC_REQUIRE_VERIFIED_EMAIL (default true) gates this. Set it to "false"
+    // to drop the requirement: Monize then trusts the IdP-provided email even
+    // without an `email_verified` claim and merges directly into an existing
+    // account matching that email -- including a local password account,
+    // skipping the email-confirmation step (so it works without SMTP). This
+    // lowers security; only disable it when you trust your IdP to verify email
+    // ownership.
     const emailVerified = userInfo.email_verified === true;
-    const trustedEmail = emailVerified ? email : undefined;
+    const requireVerifiedEmail =
+      this.configService
+        .get<string>("OIDC_REQUIRE_VERIFIED_EMAIL")
+        ?.toLowerCase() !== "false";
+    const trustedEmail =
+      emailVerified || !requireVerifiedEmail ? email : undefined;
 
     // Handle name claims - try specific claims first, fall back to 'name'
     const fullName = userInfo.name as string | undefined;
@@ -425,8 +437,8 @@ export class AuthService {
         });
 
         if (existingUser) {
-          if (existingUser.passwordHash) {
-            // SECURITY: Local account requires user confirmation before linking.
+          if (existingUser.passwordHash && requireVerifiedEmail) {
+            // SECURITY: Local account requires user confirmation before merging.
             const linkToken = await this.initiateOidcLink(existingUser, sub);
             this.logger.warn(
               `OIDC link pending confirmation for user ${existingUser.id}`,
@@ -434,7 +446,8 @@ export class AuthService {
             await this.sendOidcLinkEmail(existingUser, linkToken);
             return { user: existingUser, linkPending: true };
           } else {
-            // OIDC-only account -- safe to link directly
+            // OIDC-only account, or OIDC_REQUIRE_VERIFIED_EMAIL=false bypasses
+            // the confirmation step -- merge the OIDC identity in directly.
             existingUser.oidcSubject = sub;
             existingUser.authProvider = "oidc";
             await this.usersRepository.save(existingUser);
@@ -486,7 +499,7 @@ export class AuthService {
               where: { email: trustedEmail },
             });
             if (existingUser) {
-              if (existingUser.passwordHash) {
+              if (existingUser.passwordHash && requireVerifiedEmail) {
                 // SECURITY: Local account requires confirmation
                 const linkToken = await this.initiateOidcLink(
                   existingUser,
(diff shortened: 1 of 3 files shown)


WMP:
Tested, build on:
```
ghcr.io/wmp/monize-backend:fork-deploy-webmcp-fda8773 (and :latest)
ghcr.io/wmp/monize-frontend:fork-deploy-webmcp-fda8773 (and :latest)
```

backend/src/auth/auth.service.ts
```diff
@@ -486,7 +499,7 @@ export class AuthService {
               where: { email: trustedEmail },
             });
             if (existingUser) {
-              if (existingUser.passwordHash) {
+              if (existingUser.passwordHash && requireVerifiedEmail) {
```
kenlasko, line 502: This duplicate-email catch path got the same `&& requireVerifiedEmail` change as the primary merge path (line ~437), but unlike that path it has no dedicated test exercising it with the flag off. The two modified branches aren't symmetric in coverage.

Suggest adding a test that forces the `23505` (duplicate email) path with `OIDC_REQUIRE_VERIFIED_EMAIL=false` and asserts it merges directly without calling `sendOidcLinkEmail` — e.g. by having `dataSource.transaction` reject with `{ code: "23505" }` and the by-email `findOne` resolve the existing password account. That locks in the behavior on both branches.

---

backend/src/auth/auth.service.ts
```diff
@@ -387,9 +387,21 @@ export class AuthService {
...
+    // OIDC_REQUIRE_VERIFIED_EMAIL (default true) gates this. Set it to "false"
+    // to drop the requirement: Monize then trusts the IdP-provided email even
+    // without an `email_verified` claim and merges directly into an existing
+    // account matching that email -- including a local password account,
+    // skipping the email-confirmation step (so it works without SMTP). This
+    // lowers security; only disable it when you trust your IdP to verify email
+    // ownership.
     const emailVerified = userInfo.email_verified === true;
-    const trustedEmail = emailVerified ? email : undefined;
+    const requireVerifiedEmail =
```
kenlasko, line 399: Worth noting a (mostly pre-existing) side effect this amplifies: for an already-linked OIDC user, the login path later does `user.email = trustedEmail` to sync the email. With this flag off, `trustedEmail` is now the *unverified* IdP email, so a misconfigured/compromised IdP that changes the email claim will mutate the stored email on every login. If the new value collides with another user's email, the `save()` on that update path can throw a unique-constraint error (`23505`) that isn't caught there (unlike the create path).

Not a blocker given the feature is opt-in and the tradeoff is documented, but a one-line mention in the PR description / `.env.example` that "email also follows unverified claims on subsequent logins" would make the risk explicit.

---

kenlasko (commented):
Reviewed — clean, well-scoped, and backward-compatible. The default (`OIDC_REQUIRE_VERIFIED_EMAIL` unset → `true`) preserves existing behavior, and the security tradeoff of disabling it is honestly documented in both `.env.example` and the inline comment.

**Correctness — looks good**
- Flag parse is safe: `?.toLowerCase() !== "false"` means unset → `true`, so default is strict.
- The `&& requireVerifiedEmail` guard is applied consistently to **both** merge sites (primary path and the duplicate-email catch path) — easy to miss one.
- Verified the PR's claim that local password login keeps working after the merge: `login()` gates only on `user.passwordHash`, never on `authProvider`, and the merge preserves the hash.
- No new user-facing strings, so no i18n work needed; single-entity save, so no QueryRunner needed.

**Non-blocking suggestions** (inline)
1. Add a test for the duplicate-email catch path with the flag off — only the primary merge path is covered by the new tests.
2. Optionally call out in the description / `.env.example` that, with the flag off, the stored email also follows unverified IdP claims on subsequent logins.

Approving in spirit — just the test-coverage symmetry is the one thing I'd like to see before merge.

---

kenlasko:
@WMP are you able to add the missing test as referenced in https://github.com/kenlasko/monize/pull/683#pullrequestreview-4493796740  Was going to push a release today and wanted to have this in there.

WMP:
Done

kenlasko:
Thanks for the quick turnaround. Both review points are addressed in `95affab`:

- The duplicate-email (23505) catch path now has a dedicated test with `OIDC_REQUIRE_VERIFIED_EMAIL=false`, mirroring the primary-path coverage (forces the 23505 path, asserts a direct merge with no confirmation email). Backend unit tests are green.
- `.env.example` now documents that, with the flag off, an already-linked user's stored email follows the unverified IdP email on subsequent logins.

Resolving both threads. Code-wise this looks good to merge; the only remaining red check is the "Verify PR checklist" governance gate (needs a linked approved discussion/issue with the `approved-to-build` label), which is a maintainer call, not a code issue.

---

WMP:
I forgot about readme, do you plan to add this with release, or I should add this to this pr now?

kenlasko:
I think its OK the way it is

PR #722, opened 2026-06-18
feat(ai): reverse MCP relay chat (read-only) — bring-your-own-subscription

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/712 (labelled `approved-to-build`)

## Summary

Adds a **reverse MCP relay** so the in-app AI chat can run on the user's own Claude/ChatGPT subscription via their local MCP agent, instead of a server-side LLM provider. The chat enqueues a prompt; the user's agent long-polls it with the new `get_next_prompt` MCP tool, does the work with the existing Monize tools, and returns the answer with `post_response`. Monize never needs to be public — only the user's agent needs to reach it (e.g. on a home network).

**Read-only for now.** Queries, reports and investment data work end to end. Writes (`create_transaction`, etc.) are intentionally not enabled through the relay yet: the MCP write confirmation is an in-client elicitation, which has no one to answer it when the human is in the web chat. Moving that confirmation into the chat (typed, batchable cards reusing the #686 signed-action flow) is the agreed next step, discussed in #712.

What's included:
- **MCP relay tools** `get_next_prompt` / `post_response` on the existing Streamable-HTTP server, an in-memory per-user broker, a browser SSE endpoint, and a 3-state tunnel indicator (listening / working / offline) with a copy-paste "how to connect" helper (Claude CLI + Codex).
- **MCP Relay as a provider type** — selectable in the existing *Add AI Provider* flow with priority and a live connection state, per the direction in #712. LLM resolution skips it; the chat routes to the relay when it's the top-priority active provider.
- **Rich paste** — pasted files and the richest clipboard form (HTML) are captured as attachments; plain text still lands in the box.
- i18n for every locale; unit tests for the relay service/tools, attachment signing/service/storage/download, and the new UI.

Note: one unrelated, pre-existing test (`insights-aggregator.service.spec` "average monthly spending") is date-sensitive and fails independently of this PR; no insights files are touched here.

## Checklist

- [x] This PR addresses a **single concern** (the reverse MCP relay; attachments are the relay's file-input layer, written reusably for #687).
- [x] No shared/core areas were refactored without prior agreement (the "treat MCP Relay as just another provider" integration was requested by the maintainer in #712).

## AI assistance disclosure

Developed with Claude Code as a pair-programming assistant (design, implementation, tests, and translations). I reviewed the changes, ran the suites locally, and own the result.


WMP:
Builded on: 

ghcr.io/wmp/monize-backend:feat-ai-mcp-relay-c58113a (and :latest)
ghcr.io/wmp/monize-frontend:feat-ai-mcp-relay-c58113a (and :latest)

kenlasko:
We didn't discuss adding attachment support to Monize yet.

WMP:
Right, I need this to check if this way we can read screenshots. And this is possible. My screenshot code is only little extras additional, currently used only in this pr. We need to work with attachments on other pr, or I can remove this code (but this is write with your idea)

kenlasko:
Let's move the attachments into its own PR. Want to make sure we do this with both this and the feature request by @pagaldilz in mind. Also, rebase on main because I've added some fixes to some flakey E2E tests that were causing issues.

WMP:
I build this after remove attachemnts and after rebase, but AI still cannot add investment: 

> I checked again, and unfortunately nothing has changed: the tools available in Monize do not include the ability to create an account or add brokerage transactions (purchasing securities with a specified number of units, price, and currency). I can only view investment data. The only thing I can record is a regular cash transaction in an existing account—but that won’t show those ETF purchases as positions in the portfolio, so I won’t do that to avoid cluttering up your data.

kenlasko:
I imagine we would have issues because I implemented the same confirmation dialog that is in the add transaction tool. 

I thought you were going to try a different way to present the confirmation that would get through the relay in a new PR?

PR #734, opened 2026-06-19
Add create_security tool to MCP server and AI Assistant

kenlasko:
## Linked discussion / issue

Closes #
Approved in: 

## Summary

Adds a `create_security` tool to both the MCP server and AI Assistant, allowing users to add new securities (stocks, ETFs, mutual funds, etc.) to their security list. The tool shares lookup and validation logic between both surfaces via `SecuritiesService.previewCreateSecurity()`, ensuring identical behavior. Securities are resolved against the user's configured quote provider (Yahoo/MSN), which fills in official symbol, name, exchange, type, and currency. The tool supports optional `exchange` and `securityType` parameters (constrained to known enums) to disambiguate lookups and override defaults. A confirmation card is shown before persisting, with relay support for web-chat approval in the AI Assistant.

### Backend
- **`SecuritiesService`**: Added `previewCreateSecurity()` method that validates and resolves a security lookup without persisting. Enforces per-user unique symbol constraint and surfaces ambiguous lookups as errors listing candidates.
- **`security-enums.ts`** (new): Canonical `SECURITY_EXCHANGES` and `SECURITY_TYPES` lists used by both AI tool and MCP server to constrain user input.
- **`McpInvestmentsTools`**: Registered `create_security` tool with dry-run support, confirmation flow, and relay integration for web-chat approval.
- **`ToolExecutorService`**: Added `createSecurityAction()` to handle AI Assistant invocations, building signed pending actions.
- **`AiActionBuilderService`**: Added `buildCreateSecurity()` to construct pending actions from previews.
- **`AiActionsService`**: Added confirmation handler for `create_security` descriptors.
- **Tool definitions & schemas**: Added `create_security` to AI tool definitions, input/output schemas, and action types.
- **i18n**: Added error messages for lookup failures (not found, ambiguous, no currency, duplicate symbol) and confirmation UI strings.

### Frontend
- **`TransactionConfirmationCard`**: Added rendering for `create_security` preview (symbol, name, type, exchange, currency, favourite flag) and success state with link to securities view.
- **i18n**: Added UI strings for confirmation title, success message, and view-securities link.
- **`ai.ts` types**: Added `create_security` to `AiActionType` union.

### Tests
- **`SecuritiesService.spec.ts`**: Added comprehensive tests for `previewCreateSecurity()` covering successful lookup, overrides, validation errors, ambiguity detection, duplicate detection, and missing currency.
[...]

PR #736, opened 2026-06-19
fix(securities): read security currency from live quote, not exchange

WMP:
## Linked discussion / issue

Bug fix for `create_security` (added in #734).

## Summary

`create_security` (and the security lookup it shares with the AI Assistant) stored the **wrong currency for non-local-currency listings** — e.g. `AGGG.L` (iShares Core Global Aggregate Bond UCITS ETF, USD) was saved as **GBP** instead of USD.

**Root cause:** the Yahoo lookup derives currency from the *exchange* via `getCurrencyFromExchange()` (`LSE → GBP`), because the Yahoo *search* endpoint doesn't return a per-instrument currency. But the London Stock Exchange lists many USD- and EUR-denominated ETFs, so the exchange→currency guess is wrong for them. This isn't cosmetic: `security.currencyCode` drives the FX conversion for investment transactions (security currency → cash-account currency), so a mislabelled GBP corrupts cost/valuation math, not just the label.

**Fix:** take the instrument's currency from the live quote's `meta.currency` (which the code already reads for prices), not from the exchange.
- `QuoteResult` gains `currencyCode` (GBX/`GBp` normalized to GBP).
- `YahooFinanceService.fetchQuoteRaw` populates it from `meta.currency`.
- New `SecurityPriceService.fetchAuthoritativeCurrency()` resolves it for a symbol via the user's provider.
- `previewCreateSecurity` now resolves currency by precedence: **live-quote currency (authoritative) → explicit `currencyCode` override → exchange-guess fallback**. This composes with the `currencyCode` override added in #737 — the live quote wins, and the explicit override remains the fallback that also rescues a lookup the provider can't price (per the maintainer's note on #736). No regression: a failed/absent quote keeps the prior behavior.

Rebased on `main` after #737 merged.

Lookups for autocomplete are unchanged (no extra network calls); the authoritative fetch happens only at create time, for the one security being saved.

Verified: `AGGG.L` now resolves to USD; a normal LSE share (`GBp`) resolves to GBP. (MSN-primary users fall back to the exchange guess until MSN also reports currency — a small follow-up.)

## Checklist

- [x] This PR addresses a **single concern** (correct security currency at create).
- [x] All user-facing strings are translated for **every** locale (i18n parity). _(No new strings.)_
- [x] No shared/core areas were refactored without prior agreement (additive `QuoteResult` field + one new service method; no behavior change for existing callers).

## AI assistance disclosure

Diagnosed and fixed with Claude Code (traced the exchange→currency guess to `yahoo-finance.service.ts`); reviewed and tested locally. For this fix in particular I read through the logic myself and understand it — I reproduced the same resolution logic in a standalone script and verified it against live Yahoo data (`AGGG.L` reports USD, a normal LSE share reports `GBp`→GBP).

### DIFF
--- a/backend/src/securities/providers/quote-provider.interface.ts
+++ b/backend/src/securities/providers/quote-provider.interface.ts
@@ -10,6 +10,14 @@ export interface QuoteResult {
   regularMarketTime?: number;
   provider?: QuoteProviderName;
   /**
+   * The instrument's actual trading currency as reported by the provider
+   * (GBX/GBp normalized to GBP). Authoritative — unlike a currency guessed from
+   * the exchange, this is correct for non-local-currency listings (e.g. a
+   * USD-denominated ETF on the LSE). May be absent if the provider doesn't
+   * report it.
+   */
+  currencyCode?: string | null;
+  /**
    * For MSN, the SecId actually used to fetch this quote. May differ from
    * the security's stored msnInstrumentId when the stored value was in the
    * legacy FullInstrument form and we re-resolved on the fly. Lets the
--- a/backend/src/securities/securities.service.spec.ts
+++ b/backend/src/securities/securities.service.spec.ts
@@ -75,6 +75,10 @@ describe("SecuritiesService", () => {
     mockSecurityPriceService = {
       backfillSecurity: jest.fn().mockResolvedValue(undefined),
       lookupSecurityCandidates: jest.fn().mockResolvedValue([]),
+      // Default: provider reports no currency, so previewCreateSecurity keeps
+      // the lookup's exchange-derived currency. Tests that exercise the
+      // authoritative-currency override set their own resolved value.
+      fetchAuthoritativeCurrency: jest.fn().mockResolvedValue(null),
     };
 
     mockActionHistoryService = {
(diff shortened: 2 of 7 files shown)


WMP:
Heads-up: this overlaps #737, which adds an optional `currencyCode` override to `create_security`. Both touch the same currency-resolution line in `SecuritiesService.previewCreateSecurity` (plus the `create_security` input schema and its specs), so whichever lands second will need a small rebase.

They're complementary rather than competing:

- **#737** lets the AI *explicitly* pass a currency — a manual override, and a rescue when the provider reports none.
- **#736 (this PR)** fixes the *default*: it reads the instrument's currency from the live quote (`meta.currency`), so a USD/EUR ETF on the LSE (e.g. `AGGG.L`) resolves correctly with no one having to intervene. The exchange→currency guess (`LSE → GBP`) is simply wrong for those listings, and the override alone doesn't fix the default — the AI would still have to know to pass `currencyCode: "USD"` every time.

Combined, the natural precedence is: explicit override → live-quote currency → exchange-guess fallback:

```ts
// explicit AI override wins; otherwise prefer the instrument's live-quote
// currency; otherwise fall back to the exchange-derived guess
let currencyCode = input.currencyCode?.trim() || lookup.currencyCode?.trim();
if (!input.currencyCode) {
  const authoritative = await this.securityPriceService.fetchAuthoritativeCurrency(
    userId, lookup.symbol, lookup.exchange,
  );
  if (authoritative) currencyCode = authoritative;
}
```

Happy to rebase this PR onto #737 (or vice-versa) and wire in that precedence once either one lands — just let me know which you'd prefer to merge first.

kenlasko:
I just pushed #737 to main. Feel free to rebase and add this feature. I agree this is a good approach. My explicit currency would only be used as a fallback.

WMP:
Rebased, Mergeable

WMP:
Both fixes works! 
image

PR #743, opened 2026-06-20
feat(transfers): allow a transfer to carry a category for the monthly breakdown

### LINKED ISSUE #740: [feat]: Allow account-to-account transfers to carry a category (so investment contributions show up in monthly-category-breakdown)
## Summary

I'd like a transfer between my own accounts to be able to carry a spending
category, so that recurring investment contributions appear in the
`monthly-category-breakdown` report. Today transfers and categories are mutually
exclusive, which makes it impossible to track "how much I move into investments
each month" without distorting my expense totals.

## My use case

Every month I move money from my settlement/clearing account into my investment
accounts (regular contributions to capital investments — IKE and IKZE):

- ~100 PLN/month → IKE (brokerage account)
- ~104–130 PLN/month → IKZE

Economically these are transfers between my own accounts (the money is not
consumed, and net worth shouldn't drop). So the correct model is a **transfer**,
not an expense — I don't want it inflating my spending.

## The problem

Transfers are deliberately excluded from category reporting (they net to zero
across accounts), and a transfer cannot have a category at all. As a result:

- If I record these as **transfers** → net worth is correct, but they're
  invisible in `monthly-category-breakdown`, so I can't see my monthly
  investment allocation.
- If I record them as **categorized expenses** (e.g. an "Investments" category)
  → they show up in the breakdown, but they wrongly inflate my monthly expenses
  and understate my savings rate.

I can't have both, and that's exactly what I need: a transfer that is still
attributable to an "Investments" category in the breakdown.

## Requested change

Let a transfer optionally carry a category, and include such categorized
transfers in `monthly-category-breakdown` (ideally as an opt-in, e.g. a
per-transfer category field and/or a report toggle "include categorized
transfers"). The goal: see a monthly "Investments" line equal to the sum of my
contributions, without those amounts counting as consumption/expenses elsewhere.

## Acceptance criteria

- A transfer (clearing account → brokerage/savings account) can be assigned a
  category such as "Investments / Inwestycje kapitałowe".
- That amount appears under its category in `monthly-category-breakdown`
  (e.g. ~230 PLN/month for IKE + IKZE).
- It does **not** distort regular expense totals / net worth (still treated as a
  transfer, money not "spent").

## Example data (my last 6 months)

| Date       | Destination | Amount (PLN) |
|------------|-------------|-------------:|
| 2026-02-16 | IKZE        |       104    |
| 2026-03-05 | IKZE        |       130    |
| 2026-04-07 | IKZE        |       130    |
| 2026-05-05 | IKZE        |       130    |
| 2026-06-05 | IKZE        |       130    |
| 2026-03-05 | IKE          |       100    |
| 2026-04-07 | IKE          |       100    |
| 2026-05-05 | IKE          |       100    |
| 2026-06-05 | IKE          |       100    |

## Related gaps hit while trying to do this via the MCP server

(in case it's the same backend)

- The MCP `create_transaction` has no destination-account field, so transfers
  can't be created via MCP — only read (`get_transfers`). `categorize_transaction`
  sets a category but can't turn a row into a transfer.
- There's no `create_account` tool, so I can't even set up the IKZE destination
  account from the MCP side.
- (Separately reported: `create_security` returns `currencyCode: GBP` for
  `AGGG.L` even though the Yahoo source reports `USD`.)

## Version

I can't report an exact Monize version — the server exposes no version/build
identifier in its tool metadata, methods, or resources. Happy to add it if you
tell me where it's surfaced.

WMP:
## Linked discussion / issue

Closes #740 (labeled `approved-to-build`).

## Summary

Lets an account-to-account **transfer optionally carry a spending category**, and surfaces such categorized transfers in the `monthly-category-breakdown` report — so recurring investment contributions (e.g. monthly IKE/IKZE top-ups) show up as a monthly "Investments" line **without** being counted as expenses/income or affecting net worth.

### Behavior

- A transfer can be assigned a category (web transfer form, optional field). The category is stored on both legs.
- In the monthly category breakdown, a categorized transfer is counted **once** — its outflow leg (`amount < 0`) — under its category, as a withdrawal. The inflow leg never double-counts (it is on the destination account and is excluded both by `amount < 0` and the existing `account_type != 'INVESTMENT'` filter). Categorized transfers are removed from the report's separate transfer rollup so the same movement is not shown twice.
- Every **other** report (spending-by-category, income, comparison, tax, anomaly, data-quality, …) still filters `is_transfer = false`, so a categorized transfer is **never** treated as income/expense. Net worth is unaffected (it is the sum of account balances; categories don't enter it).
- Inclusion is **always on** (no toggle): in practice this is backward compatible because no transfer carried a category before this change, so existing reports are unchanged until a user opts a transfer in by setting one.

### Why no migration

`transactions.category_id` already exists (nullable, `ON DELETE SET NULL`) with no constraint tying it to non-transfers — the old restriction was purely application-level (the transfer DTOs didn't accept a category and the service never set one). So this is additive with **no schema change**.

## Changes

**Backend**
- `create-transfer.dto.ts` / `update-transfer.dto.ts`: optional `categoryId` (`@IsUUID`; update accepts `null` to clear).
- `transaction-transfer.service.ts`: validate the category belongs to the user (mirrors `transactions.service.create()`), store it on both legs on create, set/clear on update.
- `monthly-category-breakdown.service.ts`: category query also includes the outflow leg of categorized transfers; transfer-rollup query excludes categorized transfers to avoid double-counting.

**Frontend**
- Optional "Category (Optional)" combobox in the transfer form (`TransferTransactionFields`), with an explanatory note; wired through `TransactionForm` and the transfer API type.

## Out of scope (separate "Related gaps" from the issue)

Creating transfers via MCP (`create_transaction` has no destination field) and a `create_account` tool — those are independent and tracked separately. This PR sets the category from the web UI; the shared breakdown read inherits the new behavior automatically (no tool-signature change).

## Tests

- Transfer service: category stored on both legs (create), set/cleared (update), validated (rejects an unowned category), and not required.
- Breakdown service: category query includes the categorized-transfer outflow leg; transfer rollup excludes categorized transfers.
- Frontend: the optional category field + note render in transfer mode.
- Backend `transactions` + `built-in-reports` suites green (973); frontend transfer/transaction form tests green.

## Checklist

- [x] An approved discussion or issue exists and is linked above (#740, `approved-to-build`).
- [x] This PR addresses a **single concern** (categorized transfers in the monthly breakdown).
- [ ] All user-facing strings are translated for **every** locale — English (`en`) + pseudo (`xx`) done; the full localization pass is deferred to acceptance per the repo's English-first workflow. (2 new keys: `transactions.form.fields.categoryOptional`, `transactions.form.transferCategoryNote`.)
- [x] No shared/core areas were refactored without prior agreement (additive DTO field, additive report SQL; no signature changes to shared tools).

## AI assistance disclosure

Implemented with Claude Code; design and approach reviewed and owned by me.


WMP:
image

WMP:
Hmmm, this is something logical problem and i need your help @kenlasko , look:
image
When i choose category (type: Expense), it change sign to -:
image

When i trying to accept this, i have error:
image

I can change category type to Income, but this isnt true... I havent idea how to deal with this...

kenlasko:
This is by design for a normal transaction. The sign gets set based on if the category is an expense or income. The user can override if necessary. This doesn't make sense for a transfer, of course. I think the best course of action is to check if the transaction type is a transfer. If it is, then always set the value to positive. This is done in a common function. I can't recall what it's called.

WMP:
## Follow-up changes

These commits sit on top of the original feature commit (a transfer can carry a spending category and surface in the monthly category breakdown). They address @kenlasko's review and a couple of UX/i18n gaps found while testing.

### `ebf81325` — keep the transfer amount positive when a category is set
Selecting an *expense* category on a transfer flipped the amount negative (the normal-mode income/expense sign rule), which then failed the "amount must be positive" check, so the transfer could not be saved.

A transfer's amount is always entered as a positive number — the legs' signs are derived on save — so the category-driven sign adjustment must not run in transfer mode. `handleCategoryChange` is now guarded with `mode === 'normal'`, mirroring the guard already present in `handleAmountChange`. Per @kenlasko's review.

Adds a regression test (expense category on a transfer keeps the amount positive).

### `73b3625a` — show the assigned category on a categorized transfer row
The transactions list rendered only the linked-account arrow chip in the Category column for a transfer, completely hiding any spending category assigned to it. For a transfer that carries a category, the row now shows the **category chip alongside the transfer arrow** (e.g. `[Investments] [→ BOŚ IKE - Cash]`); the category chip is clickable to filter, like a normal category chip. Adds a test.

### `b66a876d` — full localization
Translated the two new feature strings (`transactions.form.fields.categoryOptional`, `transactions.form.transferCategoryNote`) for every full locale; the regional `en-*` variants inherit from `en`. Frontend message parity is green.

### `d94c0a07` — document the two-leg drill-down behavior (no behavior change)
The category is stored on **both legs** of the transfer, so the monthly-breakdown aggregate counts only the outflow leg (one net line, e.g. `-1000`), while a category-scoped transaction view (a report drill-down, or the transactions list filtered by that category) lists **both legs** (`-1000` and `+1000`) which visibly net to zero. This is expected, not a double-count — the net contribution lives in the aggregate. Documented inline in the breakdown query so it isn't mistaken for a bug.

### Out of scope / unchanged (intentional)
- The monthly-breakdown aggregate is correct as-is; no report behavior was changed.
- `excludeFromNetWorth` stays net-worth-only (it does not affect any report) — not repurposed.
- Category reports still exclude `INVESTMENT` accounts. A known, untouched inconsistency: the transfer rollup in the monthly breakdown does not filter investment / `excludeFromNetWorth` accounts — separate concern.

### Tests
Frontend suites (TransactionForm, TransferTransactionFields, TransactionRow) green; backend monthly-category-breakdown green; type-check clean; i18n message parity green.

WMP:
Screenshot_20260620_214937

WMP:
Tested on: 
```
ghcr.io/wmp/monize-backend:feat-categorized-transfers-d94c0a0
ghcr.io/wmp/monize-frontend:feat-categorized-transfers-d94c0a0
```

PR #752, opened 2026-06-20
fix(investments): use a real FX rate for cross-currency transactions, never a silent 1.0

### LINKED ISSUE #744: [bug]: Investment transactions use a real FX rate (from the data or Yahoo for the transaction date), never a silent 1.0
## Summary

A cross-currency investment **Buy** created via the MCP server (and the AI Assistant) silently uses an exchange rate of **1.000000** when no stored daily rate is available, so the amount posted to the funding cash account is wrong by the size of the FX rate. For a EUR security funded from a PLN account this understates the PLN cash posting ~4.3x.

This corrupts the cash-account balance and the recorded cost basis, with no error shown.

## Reproduction (real case)

Added an investment transaction via MCP (`manage_investment_transactions`, Buy):

- Brokerage account: `BOŚ Bank IKE - Brokerage (PLN)`
- Funds from: `BOŚ Bank IKE - Cash` (PLN)
- Security: `IUSQ - iShares MSCI ACWI UCITS ETF USD Acc` — quoted in **EUR**
- Date: `2026-06-08`, Qty 3 @ €104.66 → Total **€313.98**

Result in the edit form: **Currency conversion (EUR → PLN), Exchange rate 1 EUR = 1.000000 PLN**, "Posts to cash account (PLN): zł 313.98". The PLN posting equals the EUR amount; it should be ~313.98 × (EUR/PLN ≈ 4.3) ≈ **zł 1350** for 2026-06-08.

## What the source data looks like (both tiers happen in practice)

The data actually fed into MCP for this order (BOŚ, order 147291850) was the basic order summary — it has **no FX rate**, only the EUR figures, the settlement currency, and the date:

```
Papier: iShares MSCI ACWI UCITS ETF   Giełda: Deutsche Börse (Xetra)   K
Limit: 104.66   Wartość: 313.98   Prowizja: 0.00   Waluta: EUR / Waluta rozl.: PLN
Data od: 08.06.2026   Status: wykonane
```

With only this, there is nothing to derive the rate from — so this is the **Yahoo-for-the-date** case: fetch EUR/PLN for 2026-06-08 (≈ 4.25) and post ≈ zł 1335, instead of the silent 1.0 that gives zł 313.98.

A **more detailed view is optionally available** from the broker (order execution detail), and it does carry the rate and the settled PLN amount:

```
Transakcje do zlecenia:
Data        Liczba  Cena        Liczba*Cena  Po prowizji  Waluta  Kurs waluty           Rynek
08.06.2026  3       437.37909   1 312.14     1 312.14     PLN     4.251352 EUR /PLN     XETA
```

When that is supplied (**rate = 4.251352 EUR/PLN**, **settled total = 1 312.14 PLN**), it should be used directly and exactly — the **derive-from-supplied-data** case — but the MCP/AI tools currently have no field to accept a rate or a converted total, so even this is dropped and the backend defaults to 1.0.

So the two tiers below are not hypothetical: the same order yields the Yahoo case (basic paste) or the derive case (detailed paste) depending on what the user provides.

(Aside: the per-share price in the basic data, €104.66, is the order *limit*; the actual fill was ~€102.88, which is why the exact settlement is 1 312.14 PLN and not 313.98 × 4.251352. The most reliable anchor, when present, is the settled PLN total / stated rate — not a recomputation from the limit price. With only the basic data, Yahoo-for-the-date is a close approximation, off by the limit-vs-fill price difference, not by the ~4.3x FX error.)

## Root cause

- `InvestmentTransactionsService.resolveCashExchangeRate()` (`backend/src/securities/investment-transactions.service.ts:325-369`) calls `exchangeRateService.getLatestRate(source, cashCurrency)` and, when it returns `null`, **logs a warning and returns `1`** (`:361-365`). `getLatestRate` is the once-per-day stored snapshot, not the rate for the transaction's date — so a back-dated transaction (e.g. 2026-06-08) gets either today's snapshot or, if missing, `1`.
- The MCP tool (`backend/src/mcp/tools/investments.tool.ts`, create schema ~`:488-543`) and the AI Assistant tool (`backend/src/ai/query/tool-input-schemas.ts:262-272`, `tool-definitions.ts:743-807`) **expose no `exchangeRate` / converted-total field**, so the caller cannot supply or correct the rate even when it knows it. The web form pre-fills a rate via `useExchangeRates().getRate()` and lets the user adjust, but MCP/AI have no equivalent path and fall through to the `1` default.

## Proposed change

Mirror the approach we took for security currency (read it from an authoritative source instead of guessing): determine the FX rate from the best available source, and **never silently fall back to 1 for a genuine cross-currency pair**.

Precedence (analogous to currency detection):

1. **Derive from the data given to MCP/AI, if present.** If the caller supplies enough to compute the rate — e.g. the cash amount actually posted (converted total) alongside the security-currency total, or an explicit `exchangeRate` — use that. This requires exposing an optional `exchangeRate` and/or `convertedTotal`/`cashAmount` field on the investment-transaction create tools (MCP + AI, kept in sync), so an import that already contains the PLN amount can pin the exact rate.
2. **Otherwise fetch the rate for the transaction's date from Yahoo.** Use the EUR/PLN (source→cash) rate **as of the transaction date**, not "latest". The infrastructure already exists — `exchangeRateService.getLiveRate()` and `backfillHistoricalRates()` (`backend/src/currencies/exchange-rate.service.ts`); `resolveCashExchangeRate` should request the dated/historical rate (fetching from Yahoo if not stored) rather than only `getLatestRate()`.
3. **If no rate can be determined, surface it** (reject the create, or flag the transaction for review) instead of defaulting to `1` and silently corrupting the cash posting.

## Acceptance criteria

- A cross-currency investment Buy/Sell created via MCP/AI posts to the cash account using the FX rate for the **transaction date** (from supplied data if available, else Yahoo), not `1`.
- The MCP and AI Assistant investment-transaction tools accept an optional rate/converted-total override (both layers in sync, per the shared-tool rule), so callers with the real cash amount can pin it exactly.
- No silent `1.0` fallback for genuinely different currencies; an undeterminable rate is reported, not hidden.
- The reproduction case (IUSQ EUR → PLN, 2026-06-08, 3 @ €104.66) posts ~zł 1350, not zł 313.98.

## Notes / scope

- Same-currency transactions are unaffected (rate stays 1).
- This pairs naturally with PR #743 / #736 (authoritative security currency): once the security currency is correct, the cash conversion is the remaining gap for non-base-currency holdings.
- Possible follow-up: backfill/repair transactions already created with a bogus `exchangeRate = 1`.

WMP:
## Linked discussion / issue

Closes #744 (labeled `approved-to-build`).

## Summary

A cross-currency investment **Buy/Sell** created via the MCP server (and the AI Assistant) silently used an exchange rate of **1.000000** when no stored daily rate was available, so the amount posted to the funding cash account was wrong by the size of the FX rate. For a EUR security funded from a PLN account this understated the PLN cash posting ~4.3x, corrupting the cash balance and the recorded cost basis with no error shown.

Resolution now follows a clear precedence and **never silently falls back to 1.0** for a genuine cross-currency pair.

## Behavior

`InvestmentTransactionsService.resolveCashExchangeRate()` resolves the rate, in order:

1. **Explicit override.** An `exchangeRate` supplied by the caller (the web form already does this; now MCP/AI can too, e.g. from the broker's settlement data).
2. **The rate as of the transaction date.** New `ExchangeRateService.getRateForDate(from, to, date)`: the closest stored rate on or before the date, otherwise a historical fetch from Yahoo for that date (the chosen point is persisted, with its inverse, for reuse). This replaces the old `getLatestRate()` snapshot, so a back-dated buy converts at the historical rate, not today's.
3. **The latest stored rate**, as a secondary source.
4. **No determinable rate for a cross-currency pair -> reject** with `errors.securities.exchangeRateUnavailable`, instead of posting at 1.0.

Same-currency transactions are unaffected (rate stays 1). The transaction date is now threaded into all four `resolveCashExchangeRate` call sites (create, preview, embedded split, update re-resolve).

## Changes

**Backend (FX resolution)**
- `currencies/exchange-rate.service.ts`: add `getRateForDate()` (dated lookup + Yahoo historical fallback, persisted).
- `securities/investment-transactions.service.ts`: `resolveCashExchangeRate()` takes the transaction date, prefers the dated rate, and throws for an undeterminable cross-currency pair; thread the date through every caller; expose `exchangeRate` on the create/update preview and row inputs.

**Backend (shared tools, both layers in sync per the shared-tool rule)**
- MCP `mcp/tools/investments.tool.ts`: optional `exchangeRate` on `manage_investment_transactions` (create + update), forwarded through `toInvCreateRow`/`toInvUpdateRow`.
- AI `ai/query/tool-input-schemas.ts`, `tool-definitions.ts`, `tool-executor.service.ts`: matching optional `exchangeRate` field and converters.

**i18n**
- New `errors.securities.exchangeRateUnavailable` key, translated for every locale (regional `en-*` variants inherit from `en`), preserving the `{{ from }}` / `{{ to }}` placeholders.

## Acceptance criteria (from the issue)

- [x] A cross-currency Buy/Sell via MCP/AI posts using the FX rate for the **transaction date** (supplied or Yahoo), not 1.
- [x] MCP and AI investment-transaction tools accept an optional rate override (both layers in sync).
- [x] No silent 1.0 for genuinely different currencies; an undeterminable rate is reported, not hidden.
- [x] Reproduction (IUSQ EUR -> PLN, 2026-06-08, 3 @ EUR 104.66) posts ~zl 1350, not zl 313.98.

## Tests

- `getRateForDate`: stored carry-forward, dated Yahoo fallback, and null when unavailable.
- `create`: rejects an undeterminable cross-currency rate; uses the dated rate when present.
- MCP and AI: forward an explicit `exchangeRate` to the create prep.
- Touched suites green (exchange-rate, investment-transactions, MCP investments, AI tool-executor, i18n parity).

## Checklist

- [x] An approved discussion or issue exists and is linked above (#744, `approved-to-build`).
- [x] This PR addresses a **single concern** (cross-currency FX rate for investment transactions).
- [x] All user-facing strings are translated for **every** locale (1 new key: `errors.securities.exchangeRateUnavailable`).
- [x] No shared/core areas were refactored without prior agreement (additive optional field on the shared tools; the FX resolution change is internal).

## AI assistance disclosure

Implemented with Claude Code; design, approach, and the FX precedence reviewed and owned by me. Verified the reproduction case locally.

### DIFF
--- a/backend/src/ai/query/tool-definitions.ts
+++ b/backend/src/ai/query/tool-definitions.ts
@@ -769,6 +769,11 @@ export const FINANCIAL_TOOLS: AiToolDefinition[] = [
                 description:
                   "Commission or fee (up to 4 decimals). Defaults to 0.",
               },
+              exchangeRate: {
+                type: "number",
+                description:
+                  "create/update: FX rate converting the security's currency into the funding cash account's currency (e.g. for a EUR security funded from a PLN account, the EUR->PLN rate such as 4.2514). Supply this when the broker's settlement data gives the rate or the converted cash total, so the cash posting is exact. Omit for same-currency transactions, or to use the rate for the transaction date.",
+              },
               fundingAccountName: {
                 type: "string",
                 description:
--- a/backend/src/ai/query/tool-executor.service.spec.ts
+++ b/backend/src/ai/query/tool-executor.service.spec.ts
@@ -1380,6 +1380,30 @@ describe("ToolExecutorService", () => {
       expect(JSON.stringify(result.data)).not.toContain("signature-abc");
     });
 
+    it("forwards an explicit exchangeRate to the create prep (issue #744)", async () => {
+      await service.execute(userId, "manage_investment_transactions", {
+        operation: "create",
+        items: [
+          {
+            accountName: "Brokerage",
+            action: "BUY",
+            date: "2026-01-15",
+            security: "AAPL",
+            quantity: 10,
+            price: 150,
+            exchangeRate: 4.2514,
+          },
+        ],
+      });
+
+      expect(
+        investmentTransactions.prepareCreateInvestmentSingle,
+      ).toHaveBeenCalledWith(
+        userId,
+        expect.objectContaining({ exchangeRate: 4.2514 }),
+      );
+    });
+
     it("single create surfaces a 4xx preview error without a pending action", async () => {
       investmentTransactions.prepareCreateInvestmentSingle.mockRejectedValueOnce(
         new BadRequestException(
--- a/backend/src/ai/query/tool-executor.service.ts
+++ b/backend/src/ai/query/tool-executor.service.ts
@@ -1310,6 +1310,7 @@ export class ToolExecutorService {
       price: item.price as number | undefined,
       commission: item.commission as number | undefined,
       fundingAccountName: item.fundingAccountName as string | undefined,
+      exchangeRate: item.exchangeRate as number | undefined,
       description: item.description as string | undefined,
     };
   }
@@ -1325,6 +1326,7 @@ export class ToolExecutorService {
       quantity: item.quantity as number | undefined,
       price: item.price as number | undefined,
       commission: item.commission as number | undefined,
+      exchangeRate: item.exchangeRate as number | undefined,
       description: item.description as string | undefined,
     };
   }
--- a/backend/src/ai/query/tool-input-schemas.ts
+++ b/backend/src/ai/query/tool-input-schemas.ts
@@ -284,6 +284,7 @@ const manageInvestmentItemSchema = z
     quantity: nonNegativeAmountSchema.optional(),
     price: nonNegativeAmountSchema.optional(),
     commission: nonNegativeAmountSchema.optional(),
+    exchangeRate: nonNegativeAmountSchema.optional(),
     description: z.string().max(500).optional(),
     transactionId: z.string().uuid().optional(),
   })
--- a/backend/src/currencies/exchange-rate.service.spec.ts
+++ b/backend/src/currencies/exchange-rate.service.spec.ts
@@ -773,6 +773,58 @@ describe("ExchangeRateService", () => {
     });
   });
 
+  describe("getRateForDate", () => {
+    it("returns 1 for the same currency without any lookup", async () => {
+      const result = await service.getRateForDate("USD", "USD", "2026-06-08");
+
+      expect(result).toBe(1);
+      expect(exchangeRateRepository.findOne).not.toHaveBeenCalled();
+      expect(yahooFinanceService.fetchHistorical).not.toHaveBeenCalled();
+    });
+
+    it("returns the closest stored rate on or before the target date", async () => {
+      exchangeRateRepository.findOne.mockResolvedValue(mockExchangeRate);
+
+      const result = await service.getRateForDate("USD", "CAD", "2026-06-08");
+
+      expect(result).toBe(1.365);
+      // Looked up the latest stored rate not after the target date.
+      const call = exchangeRateRepository.findOne.mock.calls[0][0];
+      expect(call.where.fromCurrency).toBe("USD");
+      expect(call.where.toCurrency).toBe("CAD");
+      expect(call.order).toEqual({ rateDate: "DESC" });
+      // No Yahoo fetch needed when a stored rate exists.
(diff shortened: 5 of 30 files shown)


WMP:
Tested on:
```
ghcr.io/wmp/monize-frontend:fix-investment-fx-rate-744-e18c89c
ghcr.io/wmp/monize-backend:fix-investment-fx-rate-744-e18c89c
```

PR #762, opened 2026-06-21
fix(ui): flip MultiSelect dropdown up and cap its height to the viewport

WMP:
The `MultiSelect` dropdown always opened downward at a fixed offset below its trigger, so when the field sat near the bottom of the viewport (or inside a modal) the lower options spilled off-screen with no way to reach them. This change measures the available space above and below the trigger, flips the dropdown above when there isn't enough room below, and caps its height to the available space so the options list scrolls internally instead of overflowing. The fix lives in the shared `MultiSelect` component, so every consumer (transaction filters, tag/category pickers, etc.) benefits automatically.

## Checklist


WMP:
Before:
Screenshot_20260621_132802

After:
image

PR #792, opened 2026-06-26
Extract PDF text server-side for relay attachments

kenlasko:
## Linked discussion / issue

Closes #
Approved in: 

## Summary

PDFs uploaded in relay mode are now extracted to plain text server-side and returned to the agent as `text/plain` instead of a binary blob. This prevents the agent's MCP client (e.g. Claude Code) from falling back to a local PDF handler that prompts users to install extra tooling.

The change adds a new `extractPdfText` utility that wraps `pdf-parse` to extract the text layer from PDF bytes. PDFs with no extractable text (e.g. scanned documents) return a user-friendly note; extraction failures surface as resource read errors. Text/CSV and image attachments continue to work as before.

## Changes

- **New utility**: `backend/src/ai/relay/pdf-text.util.ts` — extracts text from PDF buffers using `pdf-parse`
- **Updated resource handler**: `backend/src/mcp/resources/relay-attachment.resource.ts` — PDFs now return extracted text instead of binary blobs
- **Test coverage**: Added unit tests for the extraction utility and three new integration tests for the relay attachment resource (successful extraction, empty text layer, extraction failure)
- **Dependencies**: Added `pdf-parse` and `@types/pdf-parse` to `package.json`

## Checklist

- [x] This PR addresses a **single concern** (PDF text extraction for relay attachments).

## AI assistance disclosure

https://claude.ai/code/session_01L63jeBjcswN8nM8fCMRt2a


WMP:
Unfortunately, it doesn’t work (note that the text in the screenshot is highlighted, so it’s just a normal PDF), and to be honest, I don’t understand why you’re making it so complicated. Why don’t you just send the PDF binary to the LLM?

image

kenlasko:
When I tried, the CLI started asking to install things related to PDFs that weren't available in my instance. I thought this would be a decent way to extract content from PDFs with selectable text. It worked on the 3 that I gave it. Maybe I should have it pass it as binary if the text extract fails.

PR #803, opened 2026-06-27
fix(ai-relay): reliable confirmation-card delivery & approval for large/slow agent turns

### LINKED ISSUE #793: AI chat: confirmation cards silently lost when the assistant takes too long to compose a reply (turn/idle timeout)
## Summary

In the AI assistant chat, actions that need a **confirmation card** (e.g. bulk create/update/delete of transactions) can be **silently lost when the assistant takes too long to compose its reply** (long reasoning or a large tool-call payload). The card either never appears to the user, or is auto-cancelled — and the user never gets a chance to approve it.

## Observed behavior

During a large bulk operation:

- `report_progress` started returning `{ delivered: false }`.
- The confirmation preview calls came back as **`Cancelled: the confirmation was declined`** — even though the user never saw or declined anything.
- User feedback verbatim: *"nie zaakceptowałem jej bo się nie pojawiła, chyba za długo myślałeś"* ("I didn't approve it because it never showed up — you probably thought too long").
- Re-issuing the **same** operation as **smaller batches** (faster to compose) made the cards appear reliably every time.

## Likely cause

The prompt/turn seems to have an idle (or total-time) budget. When the assistant spends a long time generating a big tool-call payload (e.g. a 40-item batch), the turn times out **before the confirmation card is registered/shown**, so:

1. the card never renders, and/or
2. a still-pending card is auto-declined once the prompt is considered no longer active.

In other words, the **confirmation card's lifetime appears coupled to the originating prompt turn**, and a timeout is surfaced to the model as a *user decline*.

## Impact

- Silent loss of the intended action (nothing is written, but it looks like a deliberate decline).
- Confusing UX: the assistant reports "card shown", the user sees nothing; or the assistant concludes the user declined and changes course.

## Expected behavior

- Once emitted, a confirmation card should stay valid/approvable **independent of how long the assistant took** and independent of the turn's idle timeout.
- If a turn really does time out, the result should be a **distinct "timed out / expired"** signal — not an auto-decline that is indistinguishable from a user action.

## Suggested fixes (any of)

- Decouple confirmation-card lifetime from the prompt turn; let it persist with its own (longer) TTL until explicitly approved/declined/expired.
- Reset/extend the idle timeout while the assistant is actively streaming a tool call.
- In the tool result, distinguish **auto-cancelled-due-to-timeout** from **user-declined**, so the assistant can safely retry instead of assuming a decline.
- (Workaround only) Keep batches small so the payload composes quickly.

## Repro (approx)

1. In the AI chat, trigger an action that produces a large confirmation card (e.g. bulk edit ~40 transactions in a single call).
2. Observe the card not appearing / being auto-declined; `report_progress` returns `delivered: false`.
3. Re-issue the same change as several smaller batches → cards appear and can be approved normally.

### LINKED ISSUE #812: Relay chat: idle agent polls forever and can exit mid-task after a long quiet spell
In relay mode the agent long-polls get_next_prompt indefinitely while the user is idle. Every empty poll is a fresh agent turn that accumulates in the agent's context; after a long idle spell (e.g. ~20 min) a prompt sent afterwards can be processed but the loop exits without returning the answer (context/turn pressure). Proposal: after ~10 min of user inactivity, the relay signals the agent to stop its loop cleanly, and the web chat shows a 'disconnected after inactivity -- reconnect to continue' notice. The idle clock resets on every prompt so active conversations are unaffected.

WMP:
## Linked discussion / issue

Closes #793
Closes #812
Approved in: #686 (Extending the Native AI Assistant with Write Actions — Human-in-the-Loop)

## Summary

Hardens the reverse-relay confirmation-card flow so large, slow agent turns deliver their cards and final answer reliably, already-approved cards stay approved, and an idle agent stops cleanly instead of polling forever. Single concern: relay-loop reliability, delivered as a focused commit series.

- Render `pending_action`/`chart` live via a shared upsert, and keep cards through the disconnect placeholder (`renderError` no longer wipes them).
- Tool-description + MCP-server-instruction changes so the agent never abandons a claimed prompt on `delivered:false`, pings progress early/often, batches big work (~25 rows), never reasons silently for long, and always closes with a clear "all batches sent — approve and let me know" message.
- Extend the browser pickup deadline and keep it alive while new cards keep arriving.
- Suppress the text bubble entirely for a card-only message (no empty bubble / lone cursor).
- Auto-scroll only on new messages / live turns, not on in-place card status patches.
- `projectPendingActions` preserves store-side card status (confirmed/cancelled/...) across every re-render.
- **Inactivity disconnect (was #813):** after 10 min of no prompts, `get_next_prompt` returns `{ hasPrompt: false, stop: true }` so the agent stops its polling loop instead of bloating its context forever; the chat shows a distinct "disconnected after inactivity — reconnect to continue" state. The clock resets on every prompt, so active conversations are unaffected.

## Checklist

- [x] An approved discussion or issue exists and is linked above. (#793 / #812 / #686)
- [x] This PR addresses a **single concern** (relay-loop reliability), split into a reviewable commit series.
- [x] All user-facing strings are translated for **every** locale (the inactivity status string is translated in all locales; pseudo regenerated).
- [x] No shared/core areas were refactored without prior agreement — changes are confined to the relay feature area from #686.

## AI assistance disclosure


WMP:
Fixed and tested on CSV with 90 investment transactions, AI need to generate 180 transactions.

PR #805, opened 2026-06-27
feat(reports): all-securities performance comparison chart

WMP:
## Linked discussion / issue

Closes #
Approved in: _none yet — opening for the maintainer to review the approach per propose-first. Happy to move this to a Discussion first if preferred._

## Summary

Adds an "All securities" option to the Security Performance report that draws every active security on one chart as its cumulative percent return over time, so the whole portfolio's performance can be compared at a glance.

- New `AllSecuritiesChart`: fetches each active security's price history, **rebases each to 0% at its first available price**, and merges into one date-keyed multi-line dataset. Rebasing (rather than raw prices) is what lets securities with different price levels and currencies share one axis — no FX needed. Securities with no usable price are dropped from the legend; uses the shared `CHART_SERIES` palette.
- Wired into `SecurityPerformanceReport` via an `__all__` selector sentinel; per-security stats/tabs/PDF export are hidden in this mode; a manual price refresh re-fetches via a reload key.
- Extracted `buildTimeAxisTicks` into `lib/chart-time-axis.ts` so the single- and all-securities charts share one consistent time axis.

## Checklist

- [x] An approved discussion or issue exists and is linked above. **(not yet — see note above)**
- [x] This PR addresses a **single concern** (an all-securities comparison view).
- [x] All user-facing strings are translated for **every** locale (i18n parity passes; pseudo-locale regenerated).
- [x] No shared/core areas were refactored without prior agreement — only a small, self-contained extraction of the chart time-axis helper.

## AI assistance disclosure


kenlasko:
I like it, but have some suggestions:
- Change the report name to "Security Performance"
- Instead of an option for "All Securities", allow the user to use the same multi-select dropdown with search as we use on the Transactions page with the filter dropdown for Payees/categories etc

WMP:
Follow-up observation that motivates this view.

The existing **Portfolio Value Over Time (All Accounts)** chart plots in **investements page** a single aggregated line — every security's value summed into one series. You can read the total (e.g. ~zł 12.73K → ~zł 41.14K, +30.5% over 3M) but not *which* security produced that move or how each one grew on its own. Summing everything into one line hides the per-holding story, so that chart doesn't convey the full data by itself.

That is exactly the gap the all-securities comparison view added here fills — each security rebased to its own cumulative % return so you can compare individual performance rather than just the portfolio sum.

Worth considering the same idea for the Portfolio Value Over Time chart too: an optional per-security breakdown (stacked area, or multi-line) so the aggregate can be decomposed into each security's contribution over time, instead of only the single summed line.

Screenshot_20260706_234240

kenlasko:
That's a good idea. Create a new issue for that.

PR #807, opened 2026-06-27
fix(ai-relay): cap conversation history sent to the agent

WMP:
## Linked discussion / issue

Relates to #793 (same relay reliability area)
Approved in: #686 (Extending the Native AI Assistant with Write Actions — Human-in-the-Loop)

## Summary

Caps the conversation history carried into a relayed prompt so a long chat doesn't degrade the user's MCP agent. `get_next_prompt` returned the **entire** history on every prompt, and the agent runs the relay loop in one long-lived session that also accumulates its own context — so an unbounded history bloated the model until it degraded (multi-minute "thinking", tool calls emitted as plain text, no answer).

- Trim history to the most recent ~10 turns within a ~12000-char budget in `AiRelayService.enqueuePrompt` (server-side source of truth); the newest turn is truncated rather than dropped if it alone exceeds the budget.
- Cap the history the chat sends in relay mode to the same budget; the native (non-relay) path keeps the full history.
- Add `@ArrayMaxSize(500)` to `conversationHistory` as an abuse bound.

> Sibling of #803 (both relay fixes off `main`); they touch `ai-relay.service.ts` and `aiChatStore.ts` and merge cleanly together, but will need ordering if reviewed in isolation.

## Checklist

- [x] An approved discussion or issue exists and is linked above. (#793 / #686)
- [x] This PR addresses a **single concern** (bounding relayed conversation history).
- [x] All user-facing strings are translated for **every** locale — **no new user-facing strings** in this PR.
- [x] No shared/core areas were refactored without prior agreement — confined to the relay feature area from #686.

## AI assistance disclosure

### DIFF
--- a/backend/src/ai/relay/ai-relay.service.spec.ts
+++ b/backend/src/ai/relay/ai-relay.service.spec.ts
@@ -1,4 +1,8 @@
-import { AiRelayService, INACTIVITY_TIMEOUT_MS } from "./ai-relay.service";
+import {
+  AiRelayService,
+  INACTIVITY_TIMEOUT_MS,
+  trimRelayHistory,
+} from "./ai-relay.service";
 import { RelayAttachmentStore } from "./relay-attachment.store";
 
 const USER = "user-1";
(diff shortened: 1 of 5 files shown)


WMP:
The `Frontend Unit Tests` failure here is unrelated to this PR: a date-dependent test (`BalanceHistoryChart > shows Ending balance when future transactions exist`) hardcodes dates that stopped being in the future on 2026-07-02, so it now fails on `main` for every PR. Details in #816; one-test fix in #817. Once #817 lands, re-running the failed job here should go green.

PR #811, opened 2026-06-28
fix(reports): localize sector names in the Sector Weightings report

### LINKED ISSUE #810: Sector Weightings report: sector names not translated (shown in English)
On the Sector Weightings report the sector names (Technology, Financial Services, Real Estate, ...) render in English for every locale. They come from the provider (Yahoo Finance) as a fixed taxonomy stored in English, so they can be mapped to translations. Expected: sector names localized to the user's language in the chart, table, tooltip, and PDF, with a fallback to the raw value for any unrecognised provider sector.

WMP:
## Linked discussion / issue

Closes #810

## Summary

Sector names in the Sector Weightings report came straight from the provider (Yahoo Finance) in English and were rendered untranslated, so non-English users saw "Technology", "Financial Services", etc. regardless of locale.

- Adds a `sectorNames` catalog (the ~11 fixed Yahoo/GICS sectors + "Other") and a `sectorNameKey` helper that maps a provider sector string to its i18n key **case-insensitively**. Unknown values fall back to the raw string so a new provider sector still renders rather than a missing-key placeholder.
- Applies the localized label in the chart (axis + tooltip), the data table, the PDF export, and the **sector sort** (so sort order matches the displayed text).
- No backend change — the sector taxonomy stays stored in English; only the presentation is localized.

## Checklist

- [x] An approved discussion or issue exists and is linked above. (#810)
- [x] This PR addresses a **single concern** (localizing sector names).
- [x] All user-facing strings are translated for **every** locale (parity passes; pseudo-locale regenerated).
- [x] No shared/core areas were refactored without prior agreement — confined to the one report component and its catalog.

## AI assistance disclosure

### DIFF
--- a/frontend/src/components/reports/SectorWeightingsReport.test.tsx
+++ b/frontend/src/components/reports/SectorWeightingsReport.test.tsx
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { render, screen, waitFor, fireEvent, act } from '@/test/render';
-import { SectorWeightingsReport } from './SectorWeightingsReport';
+import { SectorWeightingsReport, sectorNameKey } from './SectorWeightingsReport';
 
 vi.mock('@/lib/pdf-export', () => ({
   exportToPdf: vi.fn().mockResolvedValue(undefined),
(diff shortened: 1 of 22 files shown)


WMP:
Closing: sector names come from the data provider (Yahoo Finance), and the project's convention is to translate only our own UI strings, not values sourced from external data. Reverting this change.

PR #839, opened 2026-07-06
feat(transactions): search understands amounts and dates in the user's own format

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/818

## Summary

Makes the Transactions search box understand an amount or a date typed in the user's own locale format, **in addition to** the existing substring search (never replacing it).

Today the search term is matched as a plain substring, and the amount only as `CAST(amount AS TEXT) ILIKE` against the raw storage rendering (`1234.5600`); dates are not searched at all. So pasting `1 234,56` / `1234,56` (comma-decimal, as the app itself displays it) or typing a date as displayed (`02.07.2026`) returns nothing.

This PR interprets the query before matching, keyed off the user's stored `numberFormat` / `dateFormat` preferences:

- **Amount detection** — if the trimmed query parses as a number in the user's number format (thousands separator space/dot/comma/none; decimal separator resolved from the locale via `Intl`, with a lenient fallback to the other convention), the search **also** matches rows where `ABS(amount)` equals the parsed value exactly (transaction and split legs). Absolute value so a pasted statement figure finds both the debit and the credit leg. Exact equality on the parsed decimal (rounded to 4 dp) — `12,3` does **not** match `112,30`.
- **Date detection** — if the query parses as a complete date in the user's date format (or ISO `yyyy-MM-dd` as a universal fallback), the search **also** matches `transaction_date = parsed date`.
- **Fallback unchanged** — when the query parses as neither, behaviour is exactly today's substring search. When it parses, the interpreted match is OR-ed with the substring search, so e.g. a payee literally named "1234" is still found.

### Implementation

- New pure parser `backend/src/transactions/transaction-search-parse.util.ts` — `parseSearchTerm(term, { numberFormat, dateFormat })` -> `{ amount, date }`.
- `buildTransactionSearchClause` (the single source of truth, `transaction-search.util.ts`) gains two optional param names and emits three **parameterized, guarded** OR-terms: `ABS(amount)`/`ABS(split.amount) = :searchAmount` and `transactionDate = :searchDate`. The params are always bound by callers (null when the term didn't parse) and cast so the null bind type-checks; the predicates are inert unless the term parsed. No SQL interpolation of user input.
- All four call sites (`transactions.service`, `transaction-analytics.service`, `transaction-bulk-update.service`) inject the `UserPreference` repo, parse once per request, and bind the extra params. `findAll` threads the parsed result through target-page and running-balance so the list, filters, and reports drill-downs stay consistent.
- No schema change.

### Tests

Parser (locale decimal/thousands variants, negative, ambiguity, ISO/display dates, invalid/partial dates), clause-builder term assertions, and service-level tests proving a `de-DE` amount (`1.234,56` -> `1234.56`) and a display-format date (`02.07.2026` -> `2026-07-02`) now match. Full backend suite passes.

## Checklist

- [x] This PR addresses a **single concern** (locale-aware amount/date interpretation in transaction search).
- [x] No new user-facing strings were added (query logic only), so i18n parity is unaffected.
- [x] The shared `buildTransactionSearchClause` is extended in a backward-compatible way (new optional params; all callers updated) — this is the exact extension point named in the approved proposal.

## AI assistance disclosure

### DIFF
--- a/backend/src/transactions/transaction-analytics.service.spec.ts
+++ b/backend/src/transactions/transaction-analytics.service.spec.ts
@@ -4,12 +4,14 @@ import { Brackets } from "typeorm";
 import { TransactionAnalyticsService } from "./transaction-analytics.service";
 import { Transaction } from "./entities/transaction.entity";
 import { Category } from "../categories/entities/category.entity";
+import { UserPreference } from "../users/entities/user-preference.entity";
 import { buildTransactionSearchClause } from "./transaction-search.util";
 
 describe("TransactionAnalyticsService", () => {
   let service: TransactionAnalyticsService;
   let transactionsRepository: Record<string, jest.Mock>;
   let categoriesRepository: Record<string, jest.Mock>;
+  let userPreferenceRepository: Record<string, jest.Mock>;
 
   const userId = "user-1";
 
@@ -56,6 +58,10 @@ describe("TransactionAnalyticsService", () => {
       find: jest.fn().mockResolvedValue([]),
     };
 
+    userPreferenceRepository = {
+      findOne: jest.fn().mockResolvedValue(null),
+    };
+
     const module: TestingModule = await Test.createTestingModule({
       providers: [
         TransactionAnalyticsService,
@@ -66,6 +72,10 @@ describe("TransactionAnalyticsService", () => {
         {
           provide: getRepositoryToken(Category),
           useValue: categoriesRepository,
+        },
+        {
+          provide: getRepositoryToken(UserPreference),
+          useValue: userPreferenceRepository,
         },
       ],
     }).compile();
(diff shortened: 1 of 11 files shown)


WMP:
image
image

kenlasko:
Are you able to resolve the Bearer Security Scan issues found? I'm looking to publish a release soon

WMP:
Oh, I didn't realize those were security checks; I thought it was the check for the missing flag. I read through them, and it seems to me that in this case we have to accept the exception, because these security checks are typical for regex usage, and honestly, they're very "lenient." I'll ask the LLM as well—do you have any ideas on how to avoid using regex?

kenlasko:
I've come across this several times in scans and Claude just dealt with it. Not sure how it did that

PR #877, opened 2026-07-11
Loans: overpayment-payee recognition, editable rate timeline in the schedule, non-destructive rate detection, and booking guidance

WMP:
## Summary

A set of loan/mortgage improvements centred on **overpayment recognition** and the **interest-rate column**, plus bank-accurate historical interest, a month-grouped schedule, and a short in-app guide on how to record loan payments. Builds on the existing loan detail page and payment-detection work.

Everything is English-first and fully localized (frontend + backend parity green); two additive DB migrations (096, 097).

## What's included

**1. Recognize overpayments by a designated payee** (`a9a77a85`)
Per-loan `overpaymentPayeeId` alongside the existing overpayment category/memo — any single match flags a payment as a standalone overpayment (100% principal). Migration `096`.

**2. Move overpayment/interest recognition into the account edit form** (`b315b557`, `fcb6b710`)
Interest category + overpayment category/payee/memo now live in the loan/mortgage edit form (shared `OverpaymentRecognitionFields`), replacing the gear-menu control on the detail page. The old `OverpaymentSettingsControl` is removed.

**3. Rate History panel folded into the schedule; bank-accurate interest** (`535fec6a`)
The separate **Rate History** panel is removed and the amortization table is renamed **Loan Schedule**. Historical rows compute interest from the effective rate rather than an estimate capped at the principal-only transaction amount, so `Interest`/`Payment` track the bank for loans whose interest is booked separately.

**4. Account rate/payment are user-owned; non-destructive** (`9183d579`)
`interestRate`/`paymentAmount` are set only via the edit form. Recording/editing/deleting a rate change no longer overwrites them — `syncAccountToTimeline` becomes a read-only `resolveCurrentTimeline`, and the scheduled-payment "ask permission" flow is driven from the resolved timeline. Fixes a case where an inferred rate/payment silently broke the payoff simulator.

**5. Universal interest recognition + configurable booking mode** (`ce08a17f`)
Separate-interest pairing double-counted principal when a ledger tags the principal transfer with the **same category** as the interest expense (a common EU style): the pairing summed transfer + expense = the full installment, inflating the implied rate ~2×. Fixed universally by **excluding transfers from separate-interest pairing** (interest is always an expense, principal a transfer). Added a per-loan **`interestBookingMode`** (`AUTO` | `SPLIT` | `SEPARATE`); `AUTO` is the default and universal. Migration `097`.

**6. In-form guidance on recording loan payments** (`ca9c721b`)
A collapsible "How to record loan payments" panel: principal as an uncategorized transfer to the loan account, interest as a categorized expense, overpayments like principal.

**7. Real per-row interest, incl. overpayments and grace** (`ec97c14a`)
Each schedule row's interest is the actual separate interest expense paired to its date (exact, matching the lender), including the interest charged alongside an overpayment. Interest expenses with no principal payment in range (an interest-only grace period before repayment begins) appear as their own interest-only rows.

**8. Observed rate column, annualized by actual days** (`8ab0b796`, `efc27bc3`)
The Rate column for **historical** rows is derived from the interest actually charged — `interest / balanceBefore x 365 / days-since-previous-payment` — so it matches the lender with **no detection step and no button**. Using the real gap (not an assumed month) keeps it correct across partial first periods, payment holidays, and months where an overpayment resets the accrual clock mid-cycle. Only **projected** (future) rows stay inline-editable, for modelling an upcoming variable-rate change. The old **"Detect from history" button and its mutation path are removed** (it added no value once the rate is observed, and confused users). Genuinely irregular periods (staged grace disbursement, months right after a payment holiday) stay approximate — that is inherent to the bank's own figures, and the amounts still match exactly.

**9. Month grouping + whole-loan totals** (`5e22fea2`, `8ab0b796`)
A month with more than one entry (e.g. a regular installment plus an overpayment on different days) collapses into an aggregate row that expands on click to its per-date detail; single-entry months stay plain. A totals row sums payment/principal/interest/extra across the whole loan.

**10. Summary + projection fixes** (`be18b7a5`, `94da6684`)
- The simulator seeds from the real installment (`principal + interest`) and falls back to it whenever a stored principal-only payment can't cover the period's interest, so separately-booked-interest loans no longer project "payoff beyond forecast".
- The summary card shows the real installment instead of a stale principal-only `paymentAmount`; payoff/comparison/past-impact month labels are localized (were always English).

## Removed / relocated (intentional)
- `OverpaymentSettingsControl` → folded into the account edit form (item 2).
- `RateHistoryPanel` → folded into the Loan Schedule table (item 3).
- **"Detect from history"** button + confirm/mutation path → replaced by the observed-rate column (item 8). The backend detection endpoint is left in place but is no longer surfaced in the loan UI.

## Note for review — diagnostic rebase, no maintainer code reverted
While chasing a suspected overpayment-simulator regression, this branch was temporarily rebased **back through #876 → #874 → #873** to bisect where it broke. The cause turned out to be **stale account data** (a rate/payment clobbered by an earlier detect run), **not** a code regression. The branch is back on top of current `main` and **no maintainer code was reverted**. Given the history-shuffling, it's worth a careful look that nothing accidentally fell out — the only intentional removals are those listed above.

## Testing
- Backend: `loan-rate-changes`, `rate-change-inference`, `loan-payment-detector`, `accounts`, `net-worth` specs green (transfer-exclusion, SPLIT mode, non-destructive resolve). Backend i18n parity green. `tsc` clean.
- Frontend: loan-detail / lib suites green (observed day-count rate, month grouping, totals, real-interest pairing, grace rows). Frontend i18n parity + pseudo green. `tsc`, ESLint clean.

## Follow-ups (not in this PR)
- **Optionally allow manual correction of historical rates.** Today the historical Rate column is observed (read-only); only future rows are editable. If users want to tidy the few inherently-irregular rows (staged grace disbursement, post-holiday months), we could re-enable inline editing on historical rows too — but note it would be cosmetic (it must not change the real interest amounts), so on those irregular months the shown rate would diverge from the shown interest. Left out pending a decision on that trade-off.
- Daily-interest accrual modelling for the projection (the forward schedule still uses `balance x rate/period`; historical rows already reflect the real charged amounts).
- Optionally move the scheduled-payment "ask permission" flow to the edit form now that scalars are form-owned.


WMP:
image
image
image
image

WMP:
image
image
image


@kenlasko:  Ready to merge, but need next fixes after release

kenlasko:
Tried it out and I like the changes. But it does make the Payoff Timeline and the Impact of Overpayments charts appear incorrectly. For example, I bought my current house in 2012 and paid off the mortgage last year. As it is with Canadian mortgages, I refinanced every 5 years, usually with a different company who offered a lower interest rate.

Somehow, Monize has recognized that my mortgages are related to the same property, even though I've never specified as such. For example, for a mortgage that actually started in Aug 2022 shows interest only payments from 2012 onwards that were never made, and a payoff timeline that never existed.

image

Prior to this change, this same mortgage looked like this, which is correct:
image

WMP:
Thanks for testing this so carefully — that's a genuine bug and a great catch.

**What was happening.** Interest that's booked as a *separate* expense (not a split leg of the payment) is matched to a loan by its interest **category + source account**, with no date scope. Your sequential Canadian mortgages — refinanced every ~5 years, each a separate account — share the same interest category and source account, so the 2022 mortgage pulled in the *previous* mortgage's 2012–2021 interest. Since that interest has no matching principal payment on the 2022 account, the new "interest-only / grace period" logic rendered it as phantom interest-only rows and a payoff timeline that never existed. It wasn't detecting that the mortgages are related — it was just the shared category/source with no lifetime boundary.

**Fix** (commit `9e1f3e6`): separate interest is now scoped to the loan's own lifetime — bounded below by the origination (`paymentStartDate`, which sits *before* the first principal payment during a real interest-only grace period, so genuine grace interest is kept) or, if that's unset, the account's first transaction; and bounded above by the final payment once the loan is paid off (an active loan still accrues to today). So a previous or subsequent loan's interest no longer bleeds in.

Could you re-test on the newest fork build? **`9e1f3e6`** — https://github.com/WMP/monize/actions/runs/29209792881 (image tag `...-9e1f3e6`, building now).

One more thing: on your screenshots the tooltip **date label blends into the panel background**. That was fixed a few commits back (the payoff tooltip now uses the shared dark-mode-aware tooltip). It's present from build **`43ddaa96`** onward — https://github.com/WMP/monize/actions/runs/29205398266 — so if your screenshots predate that, it's already resolved in the build above. Could you confirm which build/image tag you were running?

**If the phantom rows persist on `9e1f3e6`**, I'm patching a bit blind here (I don't have your data), so this quick read-only query would tell us exactly what's being attributed — could you run it for the affected 2022 mortgage?

```sql
-- 1) The mortgage's config (note payment_start_date, interest_category_id, source_account_id)
SELECT id, name, payment_start_date, interest_category_id, source_account_id,
       opening_balance, current_balance
FROM accounts
WHERE account_type = 'MORTGAGE';

-- 2) Interest expenses sharing that mortgage's category + source, grouped by year
--    (plug in the interest_category_id / source_account_id from step 1)
SELECT date_trunc('year', transaction_date) AS year, count(*),
       min(transaction_date), max(transaction_date), sum(amount)
FROM transactions
WHERE category_id = '<interest_category_id>'
  AND account_id  = '<source_account_id>'
GROUP BY 1 ORDER BY 1;
```

The key questions: is `payment_start_date` set on the 2022 mortgage (and to ~2022, not earlier)? And does the interest in query 2 really span 2012→2022 on the shared category/source? That'll confirm whether the scoping is catching it or whether there's a second attribution path to close.

kenlasko:
**Update: I re-tested on `9e1f3e6` and the behaviour hasn't changed — the phantom 2012-onward interest-only rows and the bogus payoff timeline are still there on the Canwise (2022) mortgage.** The queries below explain why, and I think they point at the path the fix didn't cover.

**Mortgage config** — all five of my mortgages share the same interest category (`d9b0f2aa…`) and the same source/funding account (`fbe4875c…` = *CIBC Chequing*):

| account | payment_start_date | opening_balance |
|---|---|---|
| Mortgage - 33 Spring | 1999-07-18 | -109,500 |
| Mortgage - 20 Spring (Pre-Reno) | 2012-06-22 | -374,000 |
| CIBC Mortgage - 20 Spring | 2013-02-28 | -688,800 |
| MCAP Mortgage | 2017-05-18 | -588,207 |
| **Canwise Mortgage** (the 2022 one) | **2022-04-25** | 0 |

So `payment_start_date` on the 2022 mortgage **is** set, and it's ~2022 — good.

**But the interest does *not* span 2012→2022 on the shared category/source.** Query 2 is almost empty: only **2 transactions, both in 2012**, totalling **-$2,565.67**. Nothing from 2013 onward.

```
    year    | count |    sum
------------+-------+-----------
 2012-01-01 |     2 | -2565.67
```

The reason: my mortgage interest isn't booked as *separate* expenses — it's booked as **split legs** of each payment. If I group the interest split legs (`transaction_splits.category_id = d9b0f2aa…`) by year, they run cleanly across every mortgage's life, and the Canwise account only has transactions from 2022 on:

| year | split-leg interest count | sum |
|---|---|---|
| 2012 | 4 | -3,805 |
| 2013 | 22 | -17,115 |
| 2014 | 26 | -19,703 |
| 2015 | 26 | -19,022 |
| 2016 | 27 | -18,873 |
| 2017 | 25 | -14,025 |
| 2018 | 26 | -15,697 |
| 2019 | 26 | -16,395 |
| 2020 | 26 | -9,806 |
| 2021 | 26 | -7,105 |
| 2022 | 24 | -8,873 |
| 2023 | 26 | -14,584 |
| 2024 | 26 | -9,254 |
| 2025 | 13 | -1,557 |

Canwise (the 2022 account) transactions by year: `2022, 2023, 2024, 2025` only — nothing before 2022.

**What this means.** On my data the separate-interest path (`pairSeparateInterest`) is essentially a no-op — there's nothing for it to mis-attribute after 2012, and it's already date-scoped to ±45 days of the actual payments anyway. So the lifetime scoping in `9e1f3e6` didn't change anything for me, which is consistent with the re-test showing no change. The phantom 2012-onward rows are coming from a **different** path — most likely the **split-leg** interest of the *earlier* mortgages being pulled onto Canwise, or the payoff/amortization projection — not the separate-interest match.

So I think the fix needs to extend the same lifetime bound to the **split-leg** interest attribution (and the projection), not just the separate-expense path. Right now, whatever is gathering split-leg interest for Canwise is reaching back to 2012, even though Canwise has no transaction before 2022-04-25.

FWIW I'm on `9e1f3e6` now, so the tooltip date-label fix is included on my end — that part looks fine.

WMP:
Excellent forensics -- and you're right that it's a different path. One correction that turns out to be the whole story: the separate-interest fetch **isn't** a no-op on your data.

**Why it isn't.** `fetchLoanInterestTransactions` asks the transactions API for `categoryIds:[interest]` + `accountIds:[CIBC Chequing]`. The backend category filter (`applyCategoryFilters`) matches **both** `transaction.categoryId` *and* `splits.categoryId`:

```
.where("transaction.categoryId IN (:...ids)")
.orWhere("splits.categoryId IN (:...ids)")
```

So that query returns every payment on CIBC Chequing whose interest is a **split leg** -- i.e. all five mortgages, 2012 onward. That's exactly the cross-mortgage bleed you diagnosed; it was entering through the separate-interest path, not sitting outside it. Your "query 2 is almost empty" measured only *top-level* `category_id` (2 rows, 2012), which is why it looked like a no-op -- but the split-leg legs (your second table, 2012-2025) come through the same fetch.

The lifetime scoping in `9e1f3e6` does bound that set by date (I confirmed `payment_start_date` reaches the client, and `computePastImpact` starts the contractual curve at it -- 2022-04-25 for Canwise, not 2012), so it should have trimmed the pre-2022 rows. But the real defect is upstream: those split-leg parents don't belong in this path at all. Split-leg interest is already attributed to its own loan through the payment's recorded interest split (`readRecordedInterest`), which only ever sees *that loan's* own payments (Canwise: 2022+). Pulling the same legs a second time via a shared category + funding account is what manufactured the phantom rows and the bogus timeline.

**Fix** (`15710f8`): `fetchLoanInterestTransactions` now keeps only **standalone** interest expenses -- a top-level interest category on a non-transfer transaction -- and drops split-leg matches (a split parent has a null top-level category). On your data that path becomes the genuine no-op you expected: your interest is booked as split legs, so it's attributed once, per payment, scoped to Canwise's own life. The Polish test mortgage (interest booked as standalone expenses) still gets exact per-row interest.

Could you re-test on **`15710f8`** -- https://github.com/WMP/monize/actions/runs/29225627993 (image tag `...-15710f8`, building now)? I'd expect the 2012-onward interest-only rows and the extended timeline to be gone.

If anything phantom survives, this is the exact query the client now runs -- grouping it by year and by whether the match is top-level vs split-leg tells us immediately if something still leaks:

```sql
SELECT date_trunc('year', t.transaction_date) AS year,
       count(*) FILTER (WHERE t.category_id = '<interest_category_id>') AS standalone,
       count(*) FILTER (WHERE s.category_id = '<interest_category_id>') AS split_leg
FROM transactions t
LEFT JOIN transaction_splits s ON s.transaction_id = t.id
WHERE t.account_id = '<CIBC Chequing id>'
  AND (t.category_id = '<interest_category_id>' OR s.category_id = '<interest_category_id>')
GROUP BY 1 ORDER BY 1;
```

After the fix, only the `standalone` column should ever reach a loan's schedule; the `split_leg` column should now be ignored by this path entirely.

kenlasko:
Its better now, but still needs some work. The Loan Schedule now shows the expected schedule, but the Payoff Timeline still shows incorrect `Original Schedule`. 

Correct schedule:
image

Incorrect Schedule from your latest:
image

The Interest rate calculations in `Loan Schedule` are also incorrect. Its appearing to average the rate out by month. The original `Rate History` table was correct, and it should return in its original form.

image

WMP:
Could you run these read-only queries?

```sql
-- 1) Loan/mortgage configuration
SELECT name, account_type, interest_booking_mode,
       interest_rate, payment_amount, payment_frequency,
       term_months, amortization_months, original_principal,
       opening_balance, current_balance, payment_start_date,
       is_canadian_mortgage, is_variable_rate
FROM accounts
WHERE account_type IN ('LOAN', 'MORTGAGE')
ORDER BY payment_start_date;

-- 2) Stored rate history (loan_rate_changes) per loan/mortgage
SELECT a.name, r.effective_date, r.annual_rate,
       r.new_payment_amount, r.source
FROM loan_rate_changes r
JOIN accounts a ON a.id = r.account_id
WHERE a.account_type IN ('LOAN', 'MORTGAGE')
ORDER BY a.name, r.effective_date;
```

WMP:
Here is logical flow, maybe we can found bugs based on this document: https://github.com/WMP/monize/blob/docs/loan-overpayment-flow/docs/loan-overpayment-flow.md

kenlasko:
Here are the results of both read-only queries.

### 1) Loan/mortgage configuration

| name | account_type | interest_booking_mode | interest_rate | payment_amount | payment_frequency | term_months | amortization_months | original_principal | opening_balance | current_balance | payment_start_date | is_canadian_mortgage | is_variable_rate |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 1989 Ford Probe Loan | LOAN | AUTO | 11.0000 | 256.5400 | MONTHLY | | | | -5500.0000 | 0.0000 | 1994-05-30 | f | f |
| RRSP Loan | LOAN | AUTO | 7.0000 | 87.4500 | MONTHLY | | | | -1095.0000 | 0.0000 | 1996-03-14 | f | f |
| 1994 Ford Probe Loan | LOAN | AUTO | 8.2500 | 164.8200 | MONTHLY | | | | -14700.0000 | 0.0000 | 1997-04-10 | f | f |
| Mortgage - 33 Spring | MORTGAGE | AUTO | 4.6000 | 306.0800 | ACCELERATED_BIWEEKLY | 60 | 300 | 109500.0000 | -109500.0000 | 0.0000 | 1999-07-18 | t | f |
| RRSP Maximizer Loan | LOAN | AUTO | 4.5000 | 780.9000 | MONTHLY | | | | -9000.0000 | 0.0000 | 2001-02-14 | f | f |
| Mortgage - 20 Spring (Pre-Reno) | MORTGAGE | AUTO | 3.0900 | 1297.7200 | ACCELERATED_BIWEEKLY | 6 | 180 | 374000.0000 | -374000.0000 | 0.0000 | 2012-06-22 | t | f |
| CIBC Mortgage - 20 Spring | MORTGAGE | AUTO | 3.0000 | 1629.8600 | ACCELERATED_BIWEEKLY | 60 | 300 | 688800.0000 | -688800.0000 | 0.0000 | 2013-02-28 | t | f |
| MCAP Mortgage | MORTGAGE | AUTO | 2.0000 | 1892.5800 | ACCELERATED_BIWEEKLY | 60 | 180 | 588206.7800 | -588206.7800 | 0.0000 | 2017-05-18 | t | t |
| Canwise Mortgage | MORTGAGE | AUTO | 3.5000 | 0.0000 | ACCELERATED_BIWEEKLY | 60 | 300 | 0.0000 | 0.0000 | 0.0000 | 2022-04-25 | t | t |

(9 rows)

### 2) Stored rate history (`loan_rate_changes`) per loan/mortgage

| name | effective_date | annual_rate | new_payment_amount | source |
|---|---|---|---|---|
| Canwise Mortgage | 2022-05-13 | 1.7500 | 3200.0000 | initial |
| Canwise Mortgage | 2022-06-24 | 2.2500 | 3233.0400 | inferred |
| Canwise Mortgage | 2022-08-05 | 3.2500 | 3303.4300 | inferred |
| Canwise Mortgage | 2022-09-30 | 4.0000 | 3127.0500 | inferred |
| Canwise Mortgage | 2022-11-11 | 4.5000 | 3264.5200 | inferred |
| Canwise Mortgage | 2022-12-23 | 5.0000 | 3300.0600 | inferred |
| Canwise Mortgage | 2023-02-17 | 5.2500 | 3319.1600 | inferred |
| Canwise Mortgage | 2023-06-23 | 5.5000 | 3332.6400 | inferred |
| Canwise Mortgage | 2023-08-04 | 5.7500 | 4050.5400 | inferred |
| Canwise Mortgage | 2024-06-21 | 5.5000 | | inferred |
| Canwise Mortgage | 2024-08-16 | 5.2500 | | inferred |
| Canwise Mortgage | 2024-09-27 | 5.0000 | | inferred |
| Canwise Mortgage | 2024-11-08 | 4.5000 | 4228.2700 | inferred |
| Canwise Mortgage | 2025-01-03 | 4.0000 | | inferred |
| Canwise Mortgage | 2025-02-14 | 3.7500 | | inferred |
| Canwise Mortgage | 2025-03-28 | 3.5000 | | inferred |
| MCAP Mortgage | 2017-05-19 | 1.9900 | 1800.0000 | initial |
| MCAP Mortgage | 2017-07-28 | 2.2400 | 1900.0000 | inferred |
| MCAP Mortgage | 2017-09-22 | 2.4900 | 2000.0000 | inferred |
| MCAP Mortgage | 2018-02-09 | 2.7300 | | inferred |
| MCAP Mortgage | 2018-07-27 | 2.9800 | 2100.0000 | inferred |
| MCAP Mortgage | 2018-11-16 | 3.2300 | 2200.0000 | inferred |
| MCAP Mortgage | 2020-03-20 | 2.6800 | 2400.0000 | inferred |
| MCAP Mortgage | 2020-04-17 | 1.7400 | | inferred |
| MCAP Mortgage | 2022-03-18 | 1.9900 | 3017.4000 | inferred |
| Mortgage - 33 Spring | 1999-07-19 | 5.9300 | 350.0000 | initial |
| Mortgage - 33 Spring | 2004-05-03 | 4.5900 | 750.0000 | inferred |

(27 rows)

WMP:
Thanks — the dump was exactly what was needed. It pinpointed why the Original Schedule stretched to 2047, and it confirmed your rate history itself is in great shape (so **no, you don't need to delete anything from `loan_rate_changes`** — quite the opposite, see below).

**What your data revealed.** Both symptoms come from how the code consumed your stored rate history:

1. **Original Schedule → 2047.** The contractual ("if I never overpaid") curve recomputed a fresh PMT over the 300-month amortization and forced payoff at that term — the theoretical minimum-payment curve. Your real installment (3,200 → … → 4,228.27 bi-weekly, recorded on the rate rows) pays the loan off in ~4 years, which is what the curve should follow. There was also a second, subtler bug your config exposed: `payment_start_date` (2022-04-25) precedes the initial rate row, which detection dates at the **first installment** (2022-05-13). The timeline resolver fell back to the earliest row for the *rate* but not for the *payment*, so the schedule saw "no recorded installment" and silently dropped to the PMT-over-term curve anyway. Both fixed: the contractual schedule now starts at the initial row's recorded installment, follows the recorded payment steps as the rate moved, and ends at the loan's own payoff (~early 2026 on a dataset shaped like yours), with the amortization term kept only as a safety bound.

2. **Rate column "averaging by month".** The column showed a rate *reconstructed* from the interest charged (interest ÷ balance, annualized by the actual day count), which jitters period to period. Since your loan has a recorded rate history, it now shows the exact rate in effect on each date — the same clean, discrete values as the old Rate History panel (1.75%, 2.25%, 3.25%, …). The reconstruction remains only for loans with no recorded rate history.

I reproduced a Canwise-shaped dataset from your dump (accelerated bi-weekly, variable-rate Canadian, your 16 rate/payment steps, split-leg interest, aggressive 2025 overpayments) and ran it through the real pipeline — the payoff chart comes out with the shape of your "correct" screenshot: the dotted contractual curve hugs the actual balance and ends a few months after it, no 25-year stretch. That scenario is also pinned as a regression test now.

**Why you shouldn't delete the rate rows:** they are what draws the correct curve — the recorded payment steps drive the contractual schedule, and the discrete rates drive the schedule's rate column. Deleting them would push both back onto the fallbacks (PMT-over-term and the reconstructed rate).

Could you re-test on **`e581625a`** — https://github.com/WMP/monize/actions/runs/29266220825 (image tag `...-e581625a`, building now)? Both the Payoff Timeline's Original Schedule and the Loan Schedule's rate column should now match your expected screenshots. The branch also includes a post-review cleanup commit (dead-code removal, one shared projection builder — which incidentally fixes the Debt Payoff Timeline report seeding its projection from the raw stored payment).

WMP:
For me all still works perfect, so for my credit is ready to merge.

kenlasko:
The Payoff timeline now works properly. Good job! The only thing remaining is the return of the `Rate History` table. This was a very informative way to see exactly how rates changed over the lifetime of the mortgage. The recorded rate history was generated via the `Detect from history` button, and it worked well.

image

WMP:
Working on restoring this now, but I have a UX question about the form it should take.

Currently the interest rate lives in the Loan Schedule table, as a column next to principal, interest, and payment — both the old Rate History table and the schedule are row-per-month, so the only real difference is one column vs. a separate table. Personally I find having it all in one place much more convenient, and I don't quite see the UX benefit of splitting it back into its own table.

So before I rebuild it: is a separate Rate History table important to you specifically, or is the rate column in the current schedule table equally good for your needs? If the standalone table gives you something the column doesn't (e.g. a compact "just the rate steps" view, the Initial/Inferred badges, easier scanning of how the rate moved over the life of the loan), let me know what matters — that'll help me decide whether to bring back the full panel or just make the existing column carry that information.

kenlasko:
Its valuable to see it separately, because over the lifetime of a mortgage, the interest rate only changes at a much lower frequency than the loan payments (once per quarter at the most). Its easier to see over a dozen rows than having to scroll through hundreds of transactions, even if they are grouped by month.

kenlasko:
The original iteration of this page had a button for calculating the interest rate based on the previously entered payment info (from my original MS Money import). That button isn't available, nor is there a way to add a rate change for a mortgage in progress.  The original Rate History chart was well laid out and provided all the information I needed to see. Please re-instate that feature in full

image

WMP:
On the rate reconstruction — I dug into why I replaced the periodic annualization, and I'd value your read on it, because I think each of our approaches is right in a *different* regime and we may want a hybrid.

**What I measured.** I exported the schedule to CSV, added a `REAL BANK RATE` column straight from my bank's letters (a flat 5.50% over the compared span, and I have the recorded rate history to match: 1.95% → 4.21% → 5.50%), and computed each reconstruction's deviation from that real rate, per installment:

- **Day-count** (`interest / balance × 365 / days`, measured from the last interest-bearing event): deviation ~**0.00%** on regular months, worst case ±0.39%.
- **Periodic** (`interest / balance × periodsPerYear`, the detection annualization): small on regular months, but it **drifts hard** around overpayments and in the payoff tail — down to **−4.42%** (reads ~1.1% when the bank rate is 5.50%).

image
(Chart attached — deviation per installment of each method from the real bank rate. I tested on image tag `feat-loan-rate-history-panel-b4a3282`.)


**Why the periodic one drifts on my loan.** My interest is booked as a separate expense and I overpay. An overpayment a few days before the regular installment settles most of that period's interest, so the installment is left a tiny interest *stub*. Day-count annualizes that stub over the actual short window since interest was last settled and recovers ~5.5%; the periodic method spreads the same stub over a full nominal period (×12) and reports a far-too-low rate. Small tail balances amplify it. That's the whole −4% band.

**Why yours is still right where mine isn't.** Your periodic method carries the Canadian **semi-annual** inversion for fixed-rate mortgages. Day-count gives a *simple* annualization, which on a Canadian fixed-rate loan would be slightly off from the nominal rate the lender quotes — exactly the case your method handles and mine doesn't. (Your Canwise is variable-rate, so simple annualization applies there, but your fixed mortgages are a different story.)

**The ask — just your read, no spreadsheet needed.** On your own mortgages — especially a **fixed-rate** one — how far do the two reconstructed rates (day-count vs periodic) sit from the rate on your bank statements? The image I tested on is tagged `feat-loan-rate-history-panel-b4a3282`, so you can eyeball the two rate columns against your statement. My hypothesis:
- on variable-rate / cleanly-regular data, day-count and periodic agree and both match the bank;
- on a fixed-rate Canadian mortgage, periodic (with semi-annual compounding) matches and day-count is slightly low;
- on anything with overpayments or separately-booked interest settling interest off-cadence, periodic drifts and day-count holds.

If that holds, the fix is a **hybrid**: reconstruct the effective rate from `interest / balance` over the *actual* accrual window (day-count, so overpayments/gaps don't distort it), then annualize it with the account's compounding convention — simple for variable, semi-annual for Canadian fixed. That would be correct in both our cases instead of either method being wrong in the other's regime.

**Open question — optionally allow manual correction of historical rates.** Today the historical Rate column is observed (read-only); only future rows are editable. If users want to tidy the few inherently-irregular rows (staged grace disbursement, post-holiday months), we could re-enable inline editing on historical rows too — but note it would be cosmetic (it must not change the real interest amounts), so on those irregular months the shown rate would diverge from the shown interest. Left out pending a decision on that trade-off.

kenlasko:
I only ever got an annual statement showing my payments for the year. I didn't keep them, so the only source of data I have is the data imported from MS Money. 

I don't know if you saw my previous note, but I would really like that Rate History table back.

WMP:
image
Restored in full, on feat/loan-overpayment-manual (image tag feat-loan-overpayment-manual-69e0332):

  - Detect from history is back (same backend inference), and it's non-destructive now: it replaces only previously inferred rows — anything you added or edited by hand is kept — and it no longer overwrites the account's own rate/payment.
  - Add rate change is back too, so you can record a change on a mortgage in progress.
  - The list is the original layout — effective date, rate, Initial/Inferred badge, the payment in effect ("Payment: … / unchanged"), with Edit / Delete per row.
  - A small step chart of the rate over the loan's life sits on top, styled to match the app's other charts.

  One thing you'll notice is the placement. Rather than a full-width block it now sits as a compact panel beside the overpayment simulator (and stacks below it on narrow screens). The table is narrow, so this frees up the page without
  dropping anything — but I'm not attached to it: if you'd rather have it back as its own full-width section, it's a one-line move, just say so.

  Could you re-test on feat-loan-overpayment-manual-69e0332? Detect / Add / Edit / Delete should all behave as they did originally.

  ---
  Separately — the interest-rate calculation method (not the table).
  
  On my (non-Canadian) mortgage I compared the two reconstructions against the real rate from my bank's letters. The day-count method (interest / balance × 365 / days) tracks the bank within ~0.4% (mostly 0.00%); the periodic method (× 
  periodsPerYear) drifts hard around overpayments and in the payoff tail — down to about −4.4%. An overpayment settles most of a period's interest a few days before the installment, leaving the installment a tiny stub, which the periodic
  annualization then spreads over a full nominal period. (Tested on feat-loan-overpayment-manual-9858cfa.)

  But your periodic method is the right one for Canadian fixed-rate mortgages, where interest compounds semi-annually and a simple day-count annualization would read slightly off.

  So my proposal: switch on the existing "Canadian mortgage" config flag — use your periodic (semi-annual) calculation when it's checked, and the day-count calculation when it's not. That keeps each method in the regime where it's correct,
  instead of one being wrong in the other's case.

  Does that match what you'd expect on your mortgages? One thing worth checking on your data: your Canwise loan also has overpayments, so does the periodic method still track your statements there, or does it drift like mine did? If it drifts
  even on a Canadian loan, we'd want the day-count accrual window inside the Canadian branch too (day-count to size the window, then semi-annual to annualize) rather than a hard either/or.

kenlasko:
Looks good now. However, the Rate History is full-width. I'm fine with that. You have some front-end unit test failures. While you're fixing those, can you move the Rate History to below the Loan Schedule?

As for rate calculations, I don't have any detailed statements to check against as my mortgage is paid off and all my data entry was manually matched against my actual payments. All I have is what has been imported from MS Money.  As such, I'm fine with whatever you decide in this regard. If any other Canadian person has an issue, they will probably bring it up (but it seems that the Canadians who use this don't have mortgages either).

WMP:
The width of the table depends on whether the loan overpayment simulation form is displayed. If the loan has already been closed, the form is not displayed, so the table showing the interest rate history takes up the full width of the screen. It makes sense and is convenient for the loan simulation form to be positioned above the repayment schedule table. This also makes it the ideal place for a narrow table showing changes to loan instalments. However, if the loan has been closed, we can move this instalment history below the repayment schedule. I do wonder, though, whether there is any point in this – the table showing interest rate changes is shorter than the repayment schedule. Perhaps we could add the option to click on the title bar to collapse the table? 

As for the differences in how interest is calculated, I’ll base the logic on whether the Canadian option is selected when editing the loan, so that Monize works as well as possible for everyone.

kenlasko:
I think its fine at the bottom below the Schedule. I feel like the Rate info is less important than everything else.

WMP:
What do you think of this build? `feat-loan-overpayment-manual-2ce8812` 
I’ll tweak the appearance of the interest rate history chart slightly, and in the schedule I’ll add the option to click on an interest rate to edit it (it’ll work in the same way as ‘Add rate change’). And I think that’ll be the final commit. I’d like to leave the position of the interest rate history chart as it is.

PR #885, opened 2026-07-14
fix(loans): scrollable rate history + show payments before the configured start date

WMP:
Two loan-detail fixes, rebased on current `main`.

### 1. Scrollable rate history
With many recorded rate changes the Rate History list grew unbounded, and beside the overpayment simulator (`fillHeight`) the tall panel stretched the whole flex row and broke the page layout. The list now scrolls inside the card: it fills the available height on desktop and keeps a `max-h-96` cap on mobile and in the full-width below-schedule placement.

### 2. Show payments before the configured start date
The "original schedule" baseline in the past-impact chart started at the configured `paymentStartDate`, which can be a stale value predating the earliest recorded payment (e.g. an origination date entered at creation) -- drawing a phantom segment with no data before the real history begins. It now starts at the **earliest actual payment**, falling back to the configured date only when there are no payments yet.

Scope note: I reverted the editable first-payment-date field I'd originally added here. Per the discussion, the interest-only grace-period origination in `loan-history.ts` is intentionally left on `paymentStartDate` (it's the loan origination, not the first-payment date). The Edit-Mortgage scheduled-transaction panel + "recreate" button is left for a separate PR.

### DIFF
--- a/frontend/src/components/accounts/loan-detail/RateHistorySidebar.tsx
+++ b/frontend/src/components/accounts/loan-detail/RateHistorySidebar.tsx
@@ -98,11 +98,11 @@ export function RateHistorySidebar({
 
   return (
     <div
-      className={`relative overflow-hidden bg-white dark:bg-gray-800 rounded-lg shadow dark:shadow-gray-700/50 scroll-mt-4 ${
+      className={`relative flex flex-col overflow-hidden bg-white dark:bg-gray-800 rounded-lg shadow dark:shadow-gray-700/50 scroll-mt-4 ${
         fillHeight ? 'lg:h-full' : ''
       }`}
     >
-      <div className="relative z-10 p-4 flex flex-col gap-3">
+      <div className="relative z-10 min-h-0 flex-1 p-4 flex flex-col gap-3">
         <div className="flex flex-wrap items-start justify-between gap-2">
           <button
             type="button"
@@ -149,7 +149,11 @@ export function RateHistorySidebar({
               {t('loanDetail.rateHistory.empty')}
             </p>
           ) : (
-            <ul className="max-h-96 overflow-y-auto divide-y divide-gray-200/70 dark:divide-gray-700/70">
+            <ul
+              className={`overflow-y-auto min-h-0 divide-y divide-gray-200/70 dark:divide-gray-700/70 ${
+                fillHeight ? 'max-h-96 lg:max-h-none lg:flex-1' : 'max-h-96'
+              }`}
+            >
               {sorted.map((change) => (
                 <li
                   key={change.id}
--- a/frontend/src/lib/loan-past-impact.test.ts
+++ b/frontend/src/lib/loan-past-impact.test.ts
@@ -27,13 +27,17 @@ function makeAccount(overrides: Partial<Account> = {}): Account {
   } as Account;
 }
 
-function makeHistory(account: Account, principals: number[]): LoanHistoryResult {
+function makeHistory(
+  account: Account,
+  principals: number[],
+  dates?: string[],
+): LoanHistoryResult {
   const transactions = principals.map(
     (amount, i) =>
       ({
         id: `tx-${i}`,
         accountId: account.id,
-        transactionDate: `2025-${String(i + 1).padStart(2, '0')}-15`,
+        transactionDate: dates?.[i] ?? `2025-${String(i + 1).padStart(2, '0')}-15`,
         amount,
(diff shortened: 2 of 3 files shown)


kenlasko:
Why do we need to specifically enter the first payment date, when it should be available from the actual first payment date recorded in the transactions list?

WMP:
Very good question, i dont know. This field has been aded on https://github.com/kenlasko/monize/commit/1f02c8070e3c4cbd598f60b2174112f2d395c466 , is visible is only when you create new mortage: 
image

I write here date of first payment with principal, but before this i pay one year only interest. Now i want to edit this date, because my credit dont show payments before `paymentStartDate`

kenlasko:
We should remove that field from the New Mortgage modal and just get the paymentStartDate from the transaction list. I don't know why we would need to specify it in New/Edit Mortgage modal. Feel free to take that on in this PR.

WMP:
What about loan?

WMP:
When we remove it, we need to remove amortization preview:
image

This will be ok?

kenlasko:
We can keep the preview, but remove the `Est. Payoff Date`, don't you think?

WMP:
Yes, i think this same, but... Before I do, I mapped every place `paymentStartDate` is written, read, or depended on, because removing the field has a wider blast radius than just the modal, and one part needs a product decision from you.

**What's easy / already safe**

The display side already tolerates a missing start date: both the loan history origination (`loan-history.ts`) and the "original schedule" start in the past-impact chart (`loan-past-impact.ts`) fall back to the **first transaction date** when `paymentStartDate` is null. Every backend read is null-tolerant too:
- `loan-mortgage-account.service.ts` `updateMortgageRate` → falls back to `new Date()`
- `loan-rate-changes.service.ts` initial-rate backfill → falls back to `dayBefore(firstChange)`; `recalculatePaymentForRate` → falls back to today
- `accounts.service.ts` update/termEnd → null-guarded; response mapping returns null fine

So "just get it from the transaction list" mostly means: stop collecting it, let it stay null, and the existing first-transaction fallback does the work. I can also actively persist `paymentStartDate` from the earliest transaction (`LoanPaymentDetectorService.firstPaymentDate`) so it's a real stored value rather than always-null — happy to do whichever you prefer.

**Consequences of removing the field**

1. **Amortization preview loses its payoff/end date.** The end date is the *only* preview value that depends on the start date (payment amount, effective rate, first principal/interest, total payments, total interest do not). I'd drop the "Est. Payoff Date" row from the New Loan/Mortgage preview and relax the preview DTOs so the date becomes optional.

2. **`termEndDate` at creation.** Today it's computed as `paymentStartDate + termMonths`. Without a start date at create it'd be null until we derive a start from transactions.

3. **The auto-created recurring scheduled payment — this is the one I need your call on.** Creating a loan/mortgage today also creates a scheduled transaction for auto-posting, anchored to the first payment date. With the field gone and no transactions yet at create, there's nothing to anchor it to. Options: **(a)** don't create it at create — user runs "Set up recurring payments" afterwards (cleanest, but no auto-post until then); **(b)** keep a next-due date only inside "Set up recurring payments" and remove it only from the New/Edit modal. Which do you want?

4. **Scope.** You mentioned the New Mortgage modal — I assume the same for New Loan and both Edit modals?

Everything else (the scrollable rate-history fix already in this PR) is unaffected.

kenlasko:
It's been a while since I've visited the whole new mortgage workflow. Now I see why we have the `First Payment Date` exposed in the UI when creating the mortgage, but not exposed when editing the mortgage.  The reason for having that is because it creates the scheduled transaction using that as the date for the first payment. 

If the date needs to change before the first transaction date occurs, then the user can edit the Scheduled Transaction. For all other actions requiring the date of the first payment, we should be referring to the actual transaction, and failing that, use the Scheduled Transaction start date. 

If someone deletes the scheduled transaction for whatever reason, and wants it back, the current workflow is to delete the mortgage and recreate it.  A better workflow would be to show the Scheduled Transaction info from within the `Edit Mortgage Account` modal, and show an alert if its missing and show a button to recreate it.

WMP:
I'm not entirely sure about the full scope here yet, so I kept this PR narrow. For now I've only made the change that also shows transactions dated earlier than the scheduled loan start date -- the payoff baseline now anchors to the earliest actual payment instead of the configured `paymentStartDate`. I reverted the editable first-payment-date field I'd added, and left the Edit-Mortgage scheduled-transaction panel/recreate button for a separate PR.

Issue #889, opened 2026-07-14
[Q] Migrated from Quicken & issues with Mortgage repayments


kenlasko:
### Discussed in https://github.com/kenlasko/monize/discussions/888

<div type='discussions-op-text'>

<sup>Originally posted by **tam481** July 14, 2026</sup>
Hello @kenlasko 
I migrated to Monize months ago now and everything seemed to go well. For some reason Monize imported my mortgage account but does not recognise the payments made in Quicken. It's recognising that 54 payments have been made but only shows payments from 49 to 54 which have been set up in Monize.

Also, the mortgage started in 2023 for 25 years which means it should finish in 2048 but it's showing the original schedule as finishing in 2051. My "Amortisation Period (required)" is set as 25 Years

I know this because I tried to use the new features in v1.12 which seems very exciting but I can't yet use it properly at the moment.

Any ideas how to fix this?

Edit: Just wanted to add that the 54 payments include overpayments made over the years and months including since migrating to Monize
Edit 2: Looking at the transactions in the mortgage account, it's showing two monthly transactions (transfers) from my main bank account one for £0 and another for, what I imagine, is the principal payment made each month
Edit 3: The starting balance is completely wrong in the "balance history" chart but the transactions list shows the correct amount if I go to the oldest transaction</div>

WMP:
@tam481: This can be related with https://github.com/kenlasko/monize/pull/885 , please check docker tag: beta

Could you send screenshot from your mortgage settings, your mortgage transactions and your mortgage details?

tam481:
Sure. Do I post them here?

WMP:
Yes

tam481:
<img width="876" height="1353" alt="Image" src="https://github.com/user-attachments/assets/ff14e763-c558-4895-b42a-b5fe510215a1" />

WMP:
This is good, please send screenshot with first transactions from this both category and from mortgage,  and send screeshot from mortgage details. Please select transactions what you dont see on mortgage details.

tam481:
<img width="1308" height="702" alt="Image" src="https://github.com/user-attachments/assets/419a5080-268e-4c89-afc9-d52edfad85f3" />
<img width="1147" height="717" alt="Image" src="https://github.com/user-attachments/assets/79ede307-8756-41ca-8d78-968e2a9f16ed" />
[...]

PR #890, opened 2026-07-15
feat(loans): include loan interest by category regardless of date

WMP:
Follow-up to #885 (now merged) and the reports in #889: the loan detail should show a loan's real history even when it began before the configured `paymentStartDate`.

### Problem
Everything on the loan detail -- schedule, summary cards, payoff timeline, overpayment simulator, rate history -- derives from `deriveLoanPaymentHistory`, which scoped separately-booked interest by a **lower date bound of the configured `paymentStartDate`**. When `paymentStartDate` is set later than the real first payment (common after a Quicken/MS-Money migration, or when an origination date was guessed at setup), every payment before it was silently hidden:

- An **interest-only grace period** (interest dated before the first principal payment) disappeared.
- A migrated mortgage's **pre-migration history** showed only the payments entered in Monize (the "shows from #49" report in #889).

### Fix
Separately-booked interest is already scoped to this loan by its **configured interest category and source account** (`fetchLoanInterestTransactions`), so the date lower bound is unnecessary -- include all of it. A loan account legitimately has activity before its configured start date, and it must show and count in the schedule and every derived figure. The only bound kept is the **upper** one for a fully paid-off loan, so interest later booked in the same category (a subsequent loan) is not absorbed after this one is gone; an active loan still accrues to today.

This drops the previous account-number/description heuristic idea in favour of trusting the configured categories, which is the intended per-loan scoping mechanism.

### Trade-off (needs your call)
The old behaviour used `paymentStartDate` as a lower bound specifically to keep an **earlier refinanced mortgage** that reuses one interest category/source account out of the current loan. With date scoping gone, sequential refinances sharing a single interest category can no longer be separated by date -- the guidance becomes **give each refinance its own interest category**. That relied on `paymentStartDate` being correct, which is exactly the value that is often wrong here, so I think trusting categories is the more robust contract -- but flagging it since it affects Canadian refinancers.

### Tests
Pre-start-date grace interest now shows; all-category inclusion and the paid-off upper bound are covered; the old date-scoping ("ignore earlier loan") test is replaced with one documenting the new inclusive behaviour + the distinct-category guidance.

### DIFF
--- a/frontend/src/lib/loan-history.test.ts
+++ b/frontend/src/lib/loan-history.test.ts
@@ -815,6 +815,10 @@ describe('deriveLoanPaymentHistory with paired separate interest expenses', () =
     expect(events.some((e) => e.date.startsWith('2012'))).toBe(true);
     expect(events.some((e) => e.date.startsWith('2020'))).toBe(true);
     expect(events.some((e) => e.date.startsWith('2022'))).toBe(true);
+    // ...and this loan's own interest is still attributed to the right date,
+    // so a pairing regression can't slip through the date-presence checks.
+    const aug = events.find((e) => e.date.startsWith('2022-08'));
+    expect(aug?.interest).toBeCloseTo(1250, 0);
   });
 
   it('excludes interest booked after the final payment once the loan is paid off', () => {
--- a/frontend/src/lib/loan-history.test.ts
+++ b/frontend/src/lib/loan-history.test.ts
@@ -758,10 +758,39 @@ describe('deriveLoanPaymentHistory with paired separate interest expenses', () =
     expect(cumulativeInterest).toBeCloseTo(388.14 + 286.49 + 335.92, 2);
   });
 
-  it('scopes separate interest to the loan lifetime, ignoring an earlier loan that shares the category', () => {
-    // Sequential refinanced mortgages share the interest category and source
-    // account. A loan that started in 2022 must not show interest-only rows
-    // from a previous mortgage (2012-2021).
+  it('includes interest booked before the configured start date (interest-only grace period)', () => {
+    // Real dataset shape: the interest-only grace period starts 2019-08, but
+    // paymentStartDate was set later (2020-04, e.g. guessed at setup). Interest
+    // is scoped by category, not date, so the pre-start grace interest still
+    // shows and counts instead of being truncated at the configured start.
+    const account = makeAccount({
+      accountType: 'MORTGAGE',
+      openingBalance: -206718.35,
+      currentBalance: -142332.03,
+      interestRate: 5.5,
+      paymentStartDate: '2020-04-05',
+    });
+    const transactions = [makeTransaction({ transactionDate: '2021-07-05', amount: 469.58 })];
+    const interestTransactions = [
+      { transactionDate: '2019-08-05', amount: -388.14, isTransfer: false } as Transaction,
+      { transactionDate: '2019-09-05', amount: -286.49, isTransfer: false } as Transaction,
+      { transactionDate: '2021-07-05', amount: -335.92, isTransfer: false } as Transaction,
+    ];
+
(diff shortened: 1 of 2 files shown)


frontend/src/lib/loan-history.ts
```diff
@@ -112,29 +112,24 @@ export function deriveLoanPaymentHistory(
...
       : null;
   const loanPaidOff = currentBalance <= 0.01;
   const scopedInterestTransactions = interestTransactions.filter((tx) => {
-    const date = tx.transactionDate.split('T')[0];
-    if (originationDate && date < originationDate) return false;
-    if (loanPaidOff && lastTransactionDate && date > lastTransactionDate) return false;
-    return true;
+    if (!loanPaidOff || !lastTransactionDate) return true;
+    return tx.transactionDate.split('T')[0] <= lastTransactionDate;
   });
```
kenlasko, line 136: On the trade-off you asked about: accepted. Re-financing a Canadian mortgage results in a new mortgage (its own account and configuration), so sequential refinances sharing one interest category isn't a scenario worth date-scoping for. Trusting the configured category/source account as the per-loan scoping mechanism is fine.

Two minor follow-ups from dropping the lower bound:

1. If a user ever does point two loans at one interest category, the earlier loan's interest merges in as orphan interest-only rows whose `balance` column shows this loan's starting balance at those old dates (the orphan-merge walk seeds `lastBalance = startingBalance`). Cosmetic given the decision above, just noting it's the failure mode if the "distinct categories" guidance is ignored.

2. The doc comment on `fetchLoanInterestTransactions` (bottom of this file) still justifies its top-level-category filter partly by the sequential-refinance scenario ("a run of sequential loans that share one interest category..."). Still true for split-leg leakage, but worth a touch-up so the two comments tell one consistent story about how scoping works now.

---

frontend/src/lib/loan-history.test.ts
```diff
@@ -774,21 +803,38 @@ describe('deriveLoanPaymentHistory with paired separate interest expenses', () =
...
 
-    // Only this loan's two payments -- no phantom rows from 2012-2020.
-    expect(events).toHaveLength(2);
-    expect(events.every((e) => e.date >= '2022-08-01')).toBe(true);
-    // This loan's own interest is still attributed.
-    expect(events[0].interest).toBeCloseTo(1250, 0);
+    // All category interest is now included, including the earlier dates.
+    expect(events.some((e) => e.date.startsWith('2012'))).toBe(true);
+    expect(events.some((e) => e.date.startsWith('2020'))).toBe(true);
+    expect(events.some((e) => e.date.startsWith('2022'))).toBe(true);
```
kenlasko, line 862: The test this replaces also asserted correct interest *attribution* (`expect(events[0].interest).toBeCloseTo(1250, 0)`), while these only assert date *presence*. Worth keeping an attribution check so a pairing regression can't slip through, e.g.:

```ts
const aug = events.find((e) => e.date.startsWith('2022-08'));
expect(aug?.interest).toBeCloseTo(1250, 0);
```

---

kenlasko (commented):
Reviewed; ran the suite on this branch locally (44/44 passing in `loan-history.test.ts`). The simplified filter is correct, the new tests genuinely exercise the change, and the rewritten comment describes the new contract accurately.

On the trade-off flagged in the description: re-financing a Canadian mortgage usually results in a new mortgage, so the shared-category sequential-refinance scenario shouldn't be a problem in practice. Trusting the configured interest category + source account is the right contract — see inline comment.

Two logistics notes:
- This PR **conflicts with #892** in `loan-history.test.ts` (this one rewrites the `'scopes separate interest to the loan lifetime...'` test that #892's new test sits directly against; `loan-history.ts` auto-merges). Suggest landing #892 first — it's a self-contained bug fix — and rebasing this one on top.
- Minor inline suggestions on the replacement test (restore an interest-attribution assertion) and a doc-comment touch-up.

---

PR #892, opened 2026-07-15
fix(loans): don't fabricate analytic interest for a principal-only payment

### LINKED ISSUE #891: Loan schedule invents analytic interest for a principal-only payment when interest is booked separately (over-counts interest)
### Summary
On a loan whose interest is booked as **separate** categorized expenses (not split legs), when a single date has **more principal payments than interest transactions**, `deriveLoanPaymentHistory` assigns all of that date's booked interest to the first payment and then **fabricates analytic interest** (`balance x periodicRate`) for the remaining same-date payment(s) via the fallback path. The result over-counts interest that was never charged.

### Real-world case
A mortgage with an interest-only grace period and frequent overpayments. On **2023-09-05** the bank statement has three operations (one combined installment split by the importer into principal + interest legs, plus a principal-only regular installment, plus a small separate interest):

| Source operation | Principal | Interest |
|---|---|---|
| Overpayment ("kapitał z 1570; odsetki osobno") | 973.11 | 596.89 |
| Regular installment ("Kapitał raty 2023-09", booked **ODSETKI: 0.00**) | 1097.78 | 0.00 |
| Separate interest | 0.00 | 28.28 |

Actual interest charged that month = **596.89 + 28.28 = 625.17**.

**What the Loan Schedule shows for Sep 2023:**
```
2023-09-05  Overpayment   Payment 1598.28  Interest  625.17  Principal 0.00     Extra 973.11
2023-09-05                Payment 1957.86  Interest  860.08  Principal 1097.78  Extra 0.00
Sep 2023 total: Interest 1485.25
```
The **860.08** on the second row is invented: both booked interest transactions (596.89 + 28.28) were paired to the first (overpayment) row, and the principal-only 1097.78 installment -- which the bank booked with **ODSETKI: 0.00** -- got analytic interest instead of 0. Sep 2023 interest is reported as **1485.25** instead of the real **625.17** (+860.08).

Multiplied across the loan this is the only month that diverges, but it inflates cumulative interest and every figure derived from it (saved-interest, payoff comparison).

### Root cause
`pairSeparateInterestByDate` pairs one interest expense per payment date, and the analytic fallback (path 3 in `deriveLoanPaymentHistory`) then fills any payment left without a paired interest. When a loan books interest separately AND there are already interest transactions for that date, a remaining same-date payment should be treated as **0 interest**, not analytically derived -- the analytic path is meant for loans with no separate interest at all.

### Suggested fix
When the loan has separate interest transactions, do not analytically derive interest for a payment that could not be paired one; treat it as 0 (the interest for that date is already fully represented by the booked expenses).

### Failing test (exact data, account number redacted)
Add to `frontend/src/lib/loan-history.test.ts` (uses the file's existing `makeAccount`/`makeTransaction` helpers):
```ts
it('does not fabricate analytic interest for a principal-only payment when interest is booked separately', () => {
  // Real case (2023-09-05): two principal payments share the date, and the
  // regular installment was booked with zero interest. The two booked interest
  // expenses (596.89 + 28.28 = 625.17) are the whole month's interest; the
  // principal-only 1097.78 installment must not receive an analytic estimate.
  const account = makeAccount({
    accountType: 'MORTGAGE',
    openingBalance: -206718.35,
    currentBalance: -142332.03,
    interestRate: 5.5,
  });
  const transactions = [
    makeTransaction({ id: 'over', transactionDate: '2023-09-05', amount: 973.11, description: 'Overpayment 2023-09-05 (principal from 1570; interest separate)' }),
    makeTransaction({ id: 'reg', transactionDate: '2023-09-05', amount: 1097.78, description: 'Regular installment 2023-09 (KAPITAL: 1097.78 ODSETKI: 0.00)' }),
  ];
  const interestTransactions = [
    { transactionDate: '2023-09-05', amount: -596.89, isTransfer: false, description: 'Interest from overpayment 2023-09-05' } as Transaction,
    { transactionDate: '2023-09-05', amount: -28.28, isTransfer: false } as Transaction,
  ];

  const { events } = deriveLoanPaymentHistory(account, transactions, [], interestTransactions);
  const sept = events.filter((e) => e.date.startsWith('2023-09'));
  const septInterest = sept.reduce((s, e) => s + e.interest, 0);

  // Whole-month interest must equal the booked expenses, with no analytic top-up.
  expect(septInterest).toBeCloseTo(625.17, 2);
});
```
This currently fails: `septInterest` comes out ~1485 (625.17 + an analytic ~860 on the principal-only row).

### Environment
Monize v1.12.0. Reproduced against a real dataset; verified the source bank transactions match the imported transactions exactly (only the +860.08 analytic interest differs).

WMP:
Fixes #891.

### Problem
`deriveLoanPaymentHistory` resolves each payment's interest from a paired separately-booked interest expense, then falls back to an analytic estimate (`balance x periodicRate`) when none is paired. That fallback fired even for loans that **do** book interest separately: when a single date has more principal payments than interest expenses (e.g. an overpayment and a regular installment on the same day), the first payment consumes the date's booked interest and the second -- left unpaired -- was given invented analytic interest.

### Real case (from #891)
`2023-09-05`: an overpayment (973.11 principal, 596.89 interest booked separately) and a regular installment (1097.78 principal, booked `ODSETKI: 0.00`), plus a 28.28 interest expense. Real month interest = 596.89 + 28.28 = **625.17**, but the principal-only 1097.78 row got an analytic **~860**, so the Loan Schedule reported **1485.25** for the month -- inflating cumulative interest and everything derived from it (saved-interest, payoff comparison). Verified against the source bank statement: the imported transactions match exactly; only the fabricated interest differed.

### Fix
The analytic estimate is only meaningful for loans with **no separately-booked interest at all** (a plain principal-only transfer whose interest is recorded nowhere). When the loan books interest as separate expenses, a payment with no paired expense carried none of its own -- the date's interest is already represented by the booked expenses -- so leave it at 0 instead of estimating.

`classifyPayment` now takes a `hasSeparateInterest` flag (`scopedInterestTransactions.length > 0`); when set and no expense pairs to the payment, it returns 0 rather than an analytic value. Loans with no separate interest are unaffected -- the analytic path still runs for them (existing test covers it).

### Tests
Added a test with the exact 2023-09-05 data: two same-day principal payments + two interest expenses; asserts the month's interest is 625.17 and the principal-only installment row is 0. Existing analytic-fallback and separate-interest tests still pass.

### Behavior note
Beyond the #891 inflation, this changes one adjacent case: for a loan whose separate-interest booking only starts partway through its history (e.g. a migration where the earliest interest was never imported), early *unpaired* payments now show **0 interest** instead of an analytic estimate. Months whose interest was actually booked but did not pair still surface as orphan interest-only rows, so nothing booked is lost — only genuinely unbooked months go to zero, which is more honest than fabricating an estimate.

Follow-up (separate): the loan reports (`LoanAmortizationReport`, `DebtPayoffTimelineReport`) call `deriveLoanPaymentHistory` without `interestTransactions`, so `hasSeparateInterest` is always false there and the #891 inflation persists in those reports — tracked in a follow-up issue.

### DIFF
--- a/frontend/src/lib/loan-history.ts
+++ b/frontend/src/lib/loan-history.ts
@@ -139,11 +139,6 @@ export function deriveLoanPaymentHistory(
     if (loanPaidOff && lastTransactionDate && date > lastTransactionDate) return false;
     return true;
   });
-  // Whether this loan books interest as separate expenses. When it does, a
-  // payment with no paired interest carried none of its own (the interest for
-  // its date is already represented by the booked expenses), so it must not
-  // receive an analytic estimate -- see classifyPayment.
-  const hasSeparateInterest = scopedInterestTransactions.length > 0;
 
   // A source-account payment covering multiple loan transfers (e.g. regular +
   // extra principal) carries one interest split; count it once.
@@ -157,6 +152,15 @@ export function deriveLoanPaymentHistory(
       scopedInterestTransactions,
       repayments.map((t) => t.transactionDate.split('T')[0]),
     );
+  // Whether this loan books interest as separate expenses. When it does, a
+  // payment with no paired interest carried none of its own (the interest for
+  // its date is already represented by the booked expenses), so it must not
+  // receive an analytic estimate -- see classifyPayment. Derived from the
+  // pairing output (not the raw list) so all-transfer / zero-amount inputs,
+  // which pairSeparateInterestByDate discards, don't falsely disable the
+  // analytic fallback.
+  const hasSeparateInterest =
+    separateInterestByDate.size > 0 || orphanInterest.length > 0;
   const usedInterestDates = new Set<string>();
   const events: LoanPaymentEvent[] = [];
 
--- a/frontend/src/lib/loan-history.ts
+++ b/frontend/src/lib/loan-history.ts
@@ -26,11 +26,14 @@ import {
  *   2. otherwise, if the linked source-account transaction carries an interest
  *      split (the shape ScheduledTransactionLoanService builds), that recorded
  *      interest is used -- exact even on variable-rate loans;
- *   3. otherwise the interest is derived analytically from the running balance
- *      at the rate in effect on that date (`balance * periodicRate`, using the
- *      rate timeline when supplied), so a payment recorded without an interest
- *      split -- including loans whose interest is booked separately -- shows a
- *      realistic interest that tracks the bank rather than 100% principal.
+ *   3. otherwise, for a loan with no separately-booked interest at all, the
+ *      interest is derived analytically from the running balance at the rate in
+ *      effect on that date (`balance * periodicRate`, using the rate timeline
+ *      when supplied), so a plain principal-only transfer shows a realistic
+ *      interest that tracks the bank rather than 100% principal. When the loan
+ *      DOES book interest as separate expenses, a payment with no paired expense
+ *      carried no interest of its own (it is already represented by the booked
+ *      expenses), so it is left at 0 rather than given an analytic estimate.
  *
  * The balance walk is unchanged -- it always tracks the actual ledger amount,
  * so the projected balance still ends at the account's current balance.
@@ -136,6 +139,11 @@ export function deriveLoanPaymentHistory(
     if (loanPaidOff && lastTransactionDate && date > lastTransactionDate) return false;
     return true;
   });
+  // Whether this loan books interest as separate expenses. When it does, a
+  // payment with no paired interest carried none of its own (the interest for
+  // its date is already represented by the booked expenses), so it must not
+  // receive an analytic estimate -- see classifyPayment.
+  const hasSeparateInterest = scopedInterestTransactions.length > 0;
 
   // A source-account payment covering multiple loan transfers (e.g. regular +
   // extra principal) carries one interest split; count it once.
@@ -173,6 +181,7 @@ export function deriveLoanPaymentHistory(
         rateChanges,
         separateInterestByDate,
         usedInterestDates,
+        hasSeparateInterest,
       );
       runningBalance = Math.max(0, runningBalance - principal);
       cumulativePrincipal += principal;
@@ -206,6 +215,7 @@ export function deriveLoanPaymentHistory(
         rateChanges,
         separateInterestByDate,
         usedInterestDates,
+        hasSeparateInterest,
       );
       cumulativePrincipal += principal;
       cumulativeInterest += interest;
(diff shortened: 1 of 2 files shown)


frontend/src/lib/loan-history.ts
```diff
@@ -136,6 +139,11 @@ export function deriveLoanPaymentHistory(
     if (loanPaidOff && lastTransactionDate && date > lastTransactionDate) return false;
     return true;
   });
+  // Whether this loan books interest as separate expenses. When it does, a
+  // payment with no paired interest carried none of its own (the interest for
+  // its date is already represented by the booked expenses), so it must not
+  // receive an analytic estimate -- see classifyPayment.
+  const hasSeparateInterest = scopedInterestTransactions.length > 0;
```
kenlasko: This flag is computed from the raw scoped list, but `pairSeparateInterestByDate` below skips transfers and zero/negative amounts. A caller passing interest transactions that are all transfers (which this function's own docs say are "not interest") would set `hasSeparateInterest = true` while zero interest is actually usable — silently disabling the analytic fallback for the whole loan. `fetchLoanInterestTransactions` pre-filters transfers today, so it's latent, but deriving the flag from the pairing output is strictly more robust and self-consistent:

```ts
const hasSeparateInterest =
  separateInterestByDate.size > 0 || orphanInterest.length > 0;
```

(Just move the declaration below the `pairSeparateInterestByDate` call.)

---

kenlasko (commented):
Reviewed; ran the suite on this branch locally (43/43 passing in `loan-history.test.ts`). The diagnosis is right and the fix is well targeted: in the #891 case the date's booked interest (596.89 + 28.28) is consumed by the first same-day payment, and the second was falling through to the analytic estimate (~860), inflating the month by ~137%. The new early-return closes exactly that hole, the module docblock update is accurate, and the regression test reproduces the real dataset.

Three notes beyond the inline comment:

1. **Partially-booked histories now understate instead of overstate.** A loan whose separate-interest booking starts midway (e.g. a migration where old interest wasn't imported) now shows 0 interest for every early unpaired payment instead of an analytic estimate. Months whose interest *was* booked but didn't pair still surface as orphan interest-only rows, so nothing booked is lost — only genuinely unbooked months go to zero. Zero is arguably more honest than fabrication, but it's a behavior change beyond the #891 bug and deserves a sentence in the description.

2. **The reports still fabricate.** `LoanAmortizationReport.tsx:106` and `DebtPayoffTimelineReport.tsx:111` call `deriveLoanPaymentHistory(selectedAccount, transactions)` without `interestTransactions`, so `hasSeparateInterest` is always false there and the exact #891 inflation persists in those reports — this PR widens the disagreement between the loan detail page and the reports for the same loan. Pre-existing gap; worth a follow-up issue to pass `fetchLoanInterestTransactions` results into the reports too.

3. **Conflicts with #890** in `loan-history.test.ts` (that PR rewrites the scoping test this one's new test sits against). Suggested order: land this one first, rebase #890 on top. Semantically they compose fine — with both applied, `hasSeparateInterest` is computed over the wider date-unbounded scoped set, consistent with each PR's intent.

---

WMP:
I test this build - works good.

PR #900, opened 2026-07-16
fix(reports): pass separate interest into the loan reports

### LINKED ISSUE #893: Loan reports don't pass separate interest, so they still over-count interest for split-date payments
Follow-up to #891 / #892.

`#892` stops `deriveLoanPaymentHistory` from fabricating analytic interest for a principal-only payment when the loan books interest separately — but only when the caller passes `interestTransactions`. The loan **detail page** does; the **reports** do not:

- `frontend/src/components/reports/LoanAmortizationReport.tsx:106` — `deriveLoanPaymentHistory(selectedAccount, transactions)`
- `frontend/src/components/reports/DebtPayoffTimelineReport.tsx:111` — `deriveLoanPaymentHistory(selectedAccount, transactions)`

With no `interestTransactions`, `hasSeparateInterest` is always false in those reports, so the same #891 inflation (an analytic estimate on a same-date principal-only leg) still occurs there. This also means the reports and the loan-detail page can show different interest for the same loan.

**Fix:** load `fetchLoanInterestTransactions(selectedAccount)` in those two reports (as the loan detail container does) and pass the result into `deriveLoanPaymentHistory`, so they share the detail page's interest handling.

Pre-existing gap surfaced during the #892 review.

WMP:
## Linked discussion / issue

Closes #893
Approved in: surfaced in the review of #892 (https://github.com/kenlasko/monize/pull/892)

## Summary

`LoanAmortizationReport` and `DebtPayoffTimelineReport` called `deriveLoanPaymentHistory(account, transactions)` **without** `interestTransactions`, so `hasSeparateInterest` was always false there. Two consequences:
- the #891 over-count (an analytic interest estimate on a same-date principal-only leg) still happened in the reports even after #892 fixed it on the loan detail page;
- the reports and the loan detail page showed **different interest** for the same loan.

Both reports now load `fetchLoanInterestTransactions(account)` — scoped to the loan's configured interest category + source account — and pass it into `deriveLoanPaymentHistory`, matching the loan detail container. `LoanAmortizationReport` fetches it alongside the transactions; `DebtPayoffTimelineReport` adds a parallel `useReportData` loader.

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity). — no new strings added; parity unaffected.

## AI assistance disclosure

Written with Claude Code (Opus 4.8). I reviewed the diff and own its correctness, conventions, and tests.

### DIFF
--- a/frontend/src/components/reports/DebtPayoffTimelineReport.test.tsx
+++ b/frontend/src/components/reports/DebtPayoffTimelineReport.test.tsx
@@ -131,6 +131,53 @@ describe('DebtPayoffTimelineReport', () => {
     );
   });
 
+  it('stays in the loading state until the interest fetch resolves (no analytic-estimate flicker)', async () => {
+    mockGetAllAccounts.mockResolvedValue([
+      {
+        id: 'loan-1',
+        name: 'Mortgage',
+        accountType: 'MORTGAGE',
+        currentBalance: -100000,
+        openingBalance: -120000,
+        interestRate: 5.0,
+        paymentAmount: 800,
+        paymentFrequency: 'MONTHLY',
+        interestCategoryId: 'cat-int',
+        sourceAccountId: 'src-1',
+        isCanadianMortgage: false,
+        isVariableRate: false,
+        isClosed: false,
+      },
+    ]);
+    mockGetAllTransactions.mockResolvedValue({
+      data: [{ id: 'tx-1', transactionDate: '2024-06-01', amount: 800, linkedTransaction: null }],
+      pagination: { hasMore: false },
+    });
+    let resolveInterest!: (value: unknown[]) => void;
+    mockGetAllPages.mockReturnValue(
+      new Promise((resolve) => {
+        resolveInterest = resolve;
+      }),
+    );
+
+    render(<DebtPayoffTimelineReport />);
+
+    // Accounts and transactions have resolved, but the separate-interest fetch
+    // is still in flight: the report must keep the skeleton instead of painting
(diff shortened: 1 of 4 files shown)


kenlasko:
One thing worth standardizing before merge: the new interest loader in `DebtPayoffTimelineReport` isn't integrated the way the report's other loaders are.

The `accounts` and `transactions` loaders both fold into `isLoading`/`error` and are wired into `reload()`. The new `interestData` loader is not — it's excluded from `isLoading`/`error`, and `reload()` only calls `reloadAccounts()`/`reloadTransactions()`. As a result there's a brief window on first paint where `interestTransactions` is still `[]`, so `hasSeparateInterest` is `false` and the schedule renders with the analytic interest estimate, then snaps to the booked interest once the second request resolves. (`interestData` does re-fetch transitively when accounts reload, so it's not stale — just visibly a step behind.)

To be fair, this isn't pre-existing — the interest loader is introduced by this PR, so the flicker is new here rather than something the report already had. `LoanAmortizationReport` avoids it entirely by batching interest into the same `Promise.all` as the transactions, so the two reports now behave differently for the same fix.

Could we fold the interest loader's `isLoading`/`error` into the report's combined values (and add its `reload` to `reload()`), or gate the `payoffSchedule` memo on it, so the schedule doesn't paint with the analytic estimate first? Small change, and it keeps the two loan reports consistent.

---

WMP:
What skill/prompt do you use to review? I can do this self, will be faster

kenlasko:
I just ask Claude to review the PR and give it the link.

PR #902, opened 2026-07-16
feat(loans): goal-seek and scenario comparison for the overpayment simulator

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/827 (the overpayment-simulator discussion this extends; goal-seek answers the "work out how early I can repay my mortgage" ask raised there, and the comparison view builds on the saved scenarios from #877)

## Summary

Two additions to the overpayment simulator on the loan detail view:

**Goal-seek** (new `loan-overpayment-solver`): work backwards from a target to the recurring extra needed to reach it, via a binary search over the schedule engine (interest and payoff date are both monotonic in the extra amount):
- "how much extra per payment to **save X of interest**" (savings semantics — matching the comparison cards and the chart, so entering 90 000 means *save 90k*, not *get total interest down to 90k*)
- "how much extra per payment to be paid off by a chosen month"

A solved extra is applied to the simulator immediately — exactly like typing into "Extra per payment", no separate Apply step — so the schedule, chart and savings cards recompute at once. Already-met targets report 0 needed; unreachable targets say so. A form change clears a stale result line. The block only renders when a projection base is supplied, so the report embedding of the simulator is unaffected.

**Scenario comparison chart** (`ScenarioComparisonChart`): saved scenarios drawn as parabolic arcs on a shared monthly timeline, built from the same Recharts blocks as the Payoff Timeline chart so the two read as one family. Each arc starts today and returns to the axis at its payoff date (span = the loan's remaining life); the apex height is the interest saved vs the no-overpayment baseline; the extra paid per installment rides in the legend name (`Aggressive · +1,500 / payment`). The baseline is a flat dashed zero-line ending at the original payoff date — narrower and taller reads as better. The legend is interactive (click toggles a line, hover emphasizes it and its name), the tooltip reports each scenario's real interest saved and payoff date, the chart exports to PNG like the payoff chart, and the saved-scenarios panel gains a CSV/PDF export dropdown — the PDF report includes headline cards, the chart (when open) and the comparison table. The chart lives behind a toggle next to "Save current scenario", shown only when more than one scenario is saved.

Tests: solver (target savings / target interest / payoff month, incl. already-met and unreachable), simulator solve-and-auto-apply and hidden-without-base, chart render/legend toggle/hover/PNG, panel toggle and CSV/PDF exports. Full frontend suite passes locally incl. the i18n parity suite; all strings translated for every locale.

## Checklist

## AI assistance disclosure

Written with Claude Code (Fable 5). I reviewed the diff and own its correctness, conventions, and tests.


WMP:
WIP:
image
Ugly chart:
image

kenlasko:
Make sure the chart has an option to download using the same method used for other charts on the page. Might also be a good idea to add an "Export to PDF" option to print out the entire page.

WMP:
Looks good now: 
image

WMP:
Is ready to merge

kenlasko:
Went through this and its quite nice. Wish I had something like this when I was deep in mortgage payments. A few things to fix:
- The exported charts don't show the legend. Goes for both the `Comparison Chart` and the `Payoff Timeline` (that's my fault for not checking).
- When I suggested having an export button for the report, I meant for the entire thing as displayed. Looks like it was added to the `Saved Scenarios` section, which doesn't need any export capability. Suggest to remove the `Export` button from the `Saved Scenarios` section. Your call if you want to add the full page report in this PR.
- With all this new detail, the `Rate History` sidebar seems out-of-place and crowds things. I suggest putting it at the bottom as full-width like it is for completed mortgages. Makes for a more consistent experience.

WMP:
Claude told that all is fine, but i cannot test this now.

kenlasko:
I like the layout much better. Still feel like the `Rate History` chart should be at the bottom. 

In the `Overpayment Simulator`, we need to figure out how to change the layout so its more clear that changes to `Solve for interest saved` OR `Solve for payoff month` will affect the `Overpayment`.  Once you play with it for a minute, its obvious, but we need a better way to display it. Maybe stack the Solve boxes on top of each other? Dim the box that isn't currently used? 

Want to avoid a bunch of text explaining it, if possible.

WMP:
image
I thinking in this same way as you, and i think this is good

kenlasko:
What about something like this?
[overpayment-simulator-preview.html](https://github.com/user-attachments/files/30125363/overpayment-simulator-preview.html)

WMP:
Hmm, good idea, like as wizard. I will follow them.

WMP:
Few options on one screenshot: 
image

kenlasko:
Looking very good. Instead of a dedicated `Load` button, the scenario should load if I click anywhere on the `Saved Scenarios` row. 

I also found an issue where the Payoff Timeline Original Schedule doesn't calculate properly when using Accelerated Bi-weekly payments: #909 Don't know if you want to deal with that in this PR or not

WMP:
For me it is ready for merge

PR #916, opened 2026-07-17
feat(backup): de-identified support backup for safe bug sharing

WMP:
## Linked discussion / issue

Closes #896
Approved in: https://github.com/kenlasko/monize/discussions/896 (direction confirmed there: masking keeping first/last 2 chars, dropping account numbers/URLs/notes/descriptions, non-integer multiplier > 1, balances recomputed from scaled transactions, optional account scope, optional password, placed in Settings → Help & Support)

## Summary

Adds a **Create Support Backup** action (Settings → Help & Support) that produces a de-identified copy of the user's data to attach to a GitHub issue. The output is an ordinary version-1 backup that restores through the normal flow into a throwaway instance — nothing new to build for opening it.

**Engine** (`backend/src/backup/support-backup/`):
- `support-backup-rules.ts` — a per-column rule registry that works as an **allowlist**: a column with no rule is dropped from the output, and a golden integration test fails whenever the live schema gains a column the registry does not classify, so a future migration cannot silently start leaking a field. Names are masked (`Biedronka` → `Bi*****ka`; ≤ 4 chars fully masked), high-risk free text and secrets are dropped (descriptions, notes, memos, reference numbers, account numbers, logos/websites, `last_client_timezone`), private amounts are scaled by a single non-integer multiplier `M` while **public values stay intact** (FX rates, security prices, per-unit costs, interest rates) — scaling only the private side of a public formula would let `M` be recovered by division, so share quantities scale in lockstep with their totals.
- `support-backup-jsonb.ts` — per-key handlers for JSONB columns, themselves allowlists (an unknown key inside a blob is dropped too).
- `support-backup.service.ts` — section filtering (checkboxes: investments, scheduled, budgets, reports, import mappings, auto-backup; `ai_provider_configs` is never exported), rule application, reconciliation of derived money from the scaled values (split parents = exact sum of scaled splits; account balance = scaled opening + sum of scaled transactions, integer arithmetic), and a full UUID + user-id remap so a shared file can't be correlated with the account or another shared file. Builds in memory (reconciliation needs every table at once), like the existing encrypted-export path.
- `support-backup-scope.ts` — optional account scoping with referential closure; dimension tables stay whole, account-specific tables are trimmed.

**API**: `POST /backup/support-export` (+ `/preview`) with a validated DTO (multiplier > 1 and non-integer per the discussion, UUID account ids, optional password → existing AES-256-GCM envelope). **Frontend**: a modal with section checkboxes, account multi-select, multiplier + Regenerate, a before→after preview rendered from the actual obfuscation output, optional password, and a plain note that this is *de-identification, not anonymity* (dates, frequencies and structure survive by design so bugs still reproduce).

**Tests**: unit specs for masking/scaling/JSONB/reconciliation/remap/scoping/sections plus a leak-negative test (no original name, note, account number, secret or UUID appears in the file); an integration golden-allowlist test and a generate→restore round-trip; frontend modal and help-section specs. Docs: `docs/support-backup.md`.

## Status: draft

A structured self-review (posted below) found real defects in the **account-scoping** path — several FK classes can dangle in a scoped file and make it unrestorable, and scoped-out account UUIDs can survive un-remapped inside report/scenario configs. Fixes are coming before this leaves draft; the unscoped (full) export path is covered by the round-trip test and unaffected.

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity). — full localization pass landed (35 keys x 18 locales); the parity suite passes for every locale.
- [x] No shared/core areas were refactored without prior agreement (only an additive `collectRawExport` accessor on `BackupService`).

## AI assistance disclosure

Written with Claude Code (Fable 5). I reviewed the diff and own its correctness, conventions, and tests.


WMP:
## Self-review findings (8-angle structured review)

Ran a structured review before requesting eyes on this: 8 independent finder passes (line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) over the full diff, ~41 candidates, deduplicated and verified against the schema and the restore path. Four independent passes converged on the same defect class in `scopeToAccounts`. Ranked results:

### Correctness / privacy (will fix before un-drafting)

1. **Split-transfer counterpart dropped in scoped exports** — `support-backup-scope.ts:60`. The transfer closure follows `transactions.linked_transaction_id`/`parent_transaction_id` but never `transaction_splits.linked_transaction_id`, and for split-based transfers the forward link lives on the split. The mirror transaction is dropped while the kept split still references it (hard FK, not in restore's deferred list) → restore fails with an FK violation, and the dropped transaction's original UUID leaks un-remapped.
2. **Scheduled-transaction FKs dangle** — `support-backup-scope.ts:101`. Kept `scheduled_transactions` (and splits) retain `transfer_account_id` / `investment_funding_account_id` pointing at out-of-scope accounts that are neither pulled in as shells (shells are derived from historical transactions only) nor nulled → unrestorable file for something as common as a scheduled transfer to an unselected savings account.
3. **Investment-transaction FKs dangle** — `support-backup-scope.ts:121`. Rows are kept whenever their (possibly shell) account is in scope, but `transaction_id`/`transaction_split_id`/`funding_account_id` can point at dropped rows (inline FKs), and `linked_transaction_id` of a cross-brokerage security transfer is re-applied in restore Phase 3 by an unconditional UPDATE → FK violation in the everyday cash + brokerage setup.
4. **Account link closure is one hop deep** — `support-backup-scope.ts:37`. The `LINK_COLS` walk iterates the user's selection, not the growing primary set, and only shell accounts get dangling links nulled. A transitively-included loan keeps e.g. `source_account_id` to a second-hop account → Phase 3 UPDATE violates the FK.
5. **Reconciled balances count VOID transactions** — `support-backup.service.ts:217`. `reconcile()` sums *all* of an account's transactions, but the app excludes `status = VOID` from every balance update. A user with voided rows gets a support file whose balances contradict the app's own math — corrupting exactly the discrepancy reports this feature exists for. (Legacy split-child rows with `parent_transaction_id` would double-count the same way.)
6. **Scoped-out account UUIDs survive un-remapped** — `support-backup-rules.ts:535`. `monte_carlo_scenarios.account_ids`, `custom_reports.filters.accountIds` and `user_preferences.dashboard_widget_config` are kept verbatim, and the id remap only covers rows present in the export — so a scoped file ships the real database UUIDs of the accounts the user chose to exclude, undermining the "cannot be correlated" guarantee. The leak-negative test only asserts on in-scope ids, so it is blind to this class.
7. **Backend error messages swallowed on the frontend** — `SupportBackupModal.tsx:120`. `supportExport` uses `responseType: 'blob'`, so a JSON error (demo-mode 403, DTO 400) arrives as a Blob and `getErrorMessage` falls back to the generic toast, hiding the server's explanation.
8. **Price history fingerprints a masked ticker** — `support-backup-rules.ts:318`. Full OHLCV series matches public market data exactly, so the masked `symbol` and dropped `msn_instrument_id` don't actually hide the instrument. Open design question from the discussion: add a "skip price history" toggle, or state it plainly in the limits note. @kenlasko — preference?

### Cleanup (lower priority)

9. **Preview does full-dataset work for 15 rows** — `support-backup.service.ts:88`. `preview()` runs the full DB export, scopes twice (directly and inside `buildObfuscated`), obfuscates every table and reconciles everything to return 5 rows × 3 tables; each preview click and the final generate each re-run the full export. Fix: pass the scoped map into `buildObfuscated` and obfuscate only the preview tables.
10. **Duplicated id-remap walker** — `support-backup.service.ts:234` near-verbatim copies `BackupService`'s private remap helpers (same UUID regex, same recursion). Should be one shared util with the two behavioral deltas as parameters, so future fixes (e.g. the bigint-shares-string-form caveat documented on the original) can't silently diverge.

Below the cut: in-place mutation in `reconcile()` vs the repo's immutability rule, dead `TABLE_SECTION` map, unused `_multiplier` param, `supportBackup.cancel` duplicating `common.cancel`, no drift-guard tying `SECTION_TABLES` to the rules registry (a future investments-domain table could bypass the "Investments" checkbox), and the triplicated download-blob helper on the frontend.

Verified-clean along the way: export table list ↔ rules registry match 1:1, restore tolerates missing sections and the `supportBackup` marker, scaling uses the repo's integer-arithmetic money convention, and the unscoped full-export path round-trips correctly (integration test).

I'll push fixes for 1–7 (and the quick cleanups) before marking this ready for review.

WMP:
image
image

WMP:
```
ghcr.io/wmp/monize-backend:feat-support-backup-41db32f
ghcr.io/wmp/monize-frontend:feat-support-backup-41db32f
```

Tested, but i dont check restoring.

kenlasko:
If the user hasn't set `AI_ENCRYPTION_KEY`, make sure it handles it properly.  Note that we should probably change the name of this to just `ENCRYPTION_KEY` at some point to reflect that its not just for AI.

Suggest changing the wording for the multiplier to "Every transaction entry is multiplied by this number to hide the actual value while maintaining consistency. DO NOT SHARE THE MULTIPLIER VALUE WITH ANYBODY.

kenlasko:
Do you want to send me a couple of exports to try out?

WMP:
I send export from 2026.05.03 do 2026.05.06

kenlasko:
Got a database error when trying to restore:

query failed: INSERT INTO "payee_aliases" ("id", "payee_id", "user_id", "alias", "created_at") VALUES ($1, $2, $3, $4, $5)
         ON CONFLICT DO NOTHING -- PARAMETERS: ["0422eba3-8f54-4bf6-805f-73410beaadb1","16fe25c6-29b5-4d86-9b40-46b4dff7df7b","ebdea242-2619-43ee-9bbe-b654ac98fe6f","ZA***********.2","2026-06-08T19:57:53.092Z"]
error: error: insert or update on table "payee_aliases" violates foreign key constraint "payee_aliases_payee_id_fkey"
[Nest] 188  - 07/17/2026, 7:33:39 PM   ERROR [BackupService] Backup restore failed for user ebdea242-2619-43ee-9bbe-b654ac98fe6f: insert or update on table "payee_aliases" violates foreign key constraint "payee_aliases_payee_id_fkey"
[Nest] 188  - 07/17/2026, 7:33:39 PM   ERROR [GlobalExceptionFilter] Database error
QueryFailedError: insert or update on table "payee_aliases" violates foreign key constraint "payee_aliases_payee_id_fkey"
    at PostgresQueryRunner.query (/app/node_modules/typeorm/driver/src/driver/postgres/PostgresQueryRunner.ts:325:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async BackupService.insertRows (/app/src/backup/backup.service.ts:1595:7)
    at async BackupService.restoreData (/app/src/backup/backup.service.ts:546:31)
    at async BackupController.restoreBackup (/app/src/backup/backup.controller.ts:121:20)
    
    
I verified right afterwards that a restore from a normal backup worked fine.

kenlasko:
Went through your support file you sent me and everything looks properly anonymized. I'd say this is ready to merge

PR #924, opened 2026-07-17
fix(reports): UTF-8 font + compressed chart images in PDF exports

### LINKED ISSUE #923: Report PDF export: non-Latin characters garbled + huge file size from uncompressed chart images
## Summary

Two independent defects in the shared report PDF export (`frontend/src/lib/pdf-export*.ts`):

### 1. Non-Latin characters are garbled
All PDF text is drawn with jsPDF's built-in Helvetica, which is a standard-14 font limited to WinAnsi/CP1252. Latin-Extended characters render wrong -- e.g. Polish `ł` prints as `B`, and other diacritics are mangled. This affects every locale beyond basic Latin (Polish, Czech, Hungarian, Turkish, Cyrillic, Greek, ...).

### 2. Report PDFs are enormous (tens of MB)
A loan report exported to **~33 MB**. Inspecting the file, a single chart is embedded as a **5160x1038 DeviceRGB image with `/Filter None`** -- i.e. raw, uncompressed pixels (5160 x 1038 x 3 = 16,068,240 bytes). jsPDF stores decoded pixels raw unless a compression flag is passed to `addImage`. The `scale: 3` chart capture also makes the raster larger than needed.

### 3. (minor) Inconsistent fonts between sections
Summary cards and headers are drawn via `doc.setFont`, while tables use `jspdf-autotable` which stayed on Helvetica, so the two look different in the same document. Long table cells also overflow instead of wrapping.

## Root cause / fix

- Embed a UTF-8 font (a Roboto subset covering Latin/Latin-Extended/Cyrillic/Greek) and use it for all text; jsPDF subsets it again at output so European PDFs stay small.
- Pass `compression: 'FAST'` to `addImage` (measured: 15.33 MB -> 0.08 MB for the offending image) and lower capture scale 3 -> 2.
- Point `autoTable` at the same embedded font and enable `overflow: 'linebreak'` for cell wrapping.

Fix implemented in branch `fix/pdf-utf8-and-size` (WMP fork).

WMP:
## Linked discussion / issue

Closes #923
Approved in: https://github.com/kenlasko/monize/issues/923

## Summary

Fixes report PDF export quality, all in the shared pipeline (`frontend/src/lib/pdf-export*.ts`):

- **UTF-8 text.** jsPDF's built-in Helvetica is WinAnsi-only, so Polish `ł` printed as `B` and other diacritics were garbled. Embed a Roboto subset (Latin, Latin Extended-A/B, Cyrillic, Greek, punctuation, currency symbols) and use it for all text. jsPDF subsets it again at output, so European PDFs stay small. (CJK/Devanagari are out of scope -- those fonts are megabytes each and can't ship in every PDF.)
- **File size.** Chart images were embedded uncompressed -- one 5160x1038 raster was ~15 MB, ballooning a loan report to ~33 MB. Pass `compression: 'FAST'` to `addImage` (measured 15.33 MB -> 0.08 MB for that image) and lower capture scale 3 -> 2 (still ~325 dpi at the printed width).
- **Consistent fonts + wrapping.** Cards/headers used `setFont` while tables (jspdf-autotable) stayed on Helvetica; point autoTable at the same embedded font and enable `overflow: 'linebreak'` so long cell content wraps instead of overflowing.

## Checklist

- [x] This PR addresses a **single concern** (PDF export output quality; the three items share the export pipeline).
- [x] All user-facing strings are translated for **every** locale (i18n parity) -- no new strings added.
- [x] No shared/core areas were refactored without prior agreement -- targeted bug fix to the export lib, tracked in #923.

## AI assistance disclosure

Implemented with Claude Code (Opus 4.8). Root cause (uncompressed `/Filter None` raster; WinAnsi built-in font) diagnosed from the actual 33 MB file; font subset generated with fonttools; image-compression delta verified with jsPDF in Node. I have reviewed and own the result.


WMP:
Tested, works good, ready to merge

PR #937, opened 2026-07-21
feat(loans): "total monthly payment" simulation type (fixed loan budget)

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/827 (the overpayment simulator, shipped in 1.12.0 via #877) -- this adds one more simulation type to that same simulator. Happy to split it back out into its own discussion if you'd rather scope it there first.

## Summary

Adds a **"Total monthly payment"** simulation type to the loan overpayment simulator: instead of entering an overpayment or a goal, you enter a **fixed total to spend on the loan each month** (installment + overpayment). The overpayment is whatever is left after the installment, so as the installment falls the overpayment grows and the total stays constant.

Example (real borrower flow): "I spend 4000/month on the loan" -> month 1: installment 918, overpayment 3082; month 2: installment 898, overpayment 3102, and so on.

### Both modes

The post-overpayment **mode** applies to a budget (defaults to **lower-installment**, matching how banks apply overpayments):

- **Lower installment** (default) -- the installment is re-amortized over the remaining term (it shrinks), so the overpayment grows to fill the budget.
- **Shorten term** -- the contractual installment stays fixed, the overpayment stays constant.

Because the *total* each period is the budget in both modes, the balance curve, payoff date and interest are identical -- only the installment/overpayment split shown differs.

### Optional window

The budget can be time-bounded with an optional **Starting / Until** date: within the window the whole budget is spent; outside it the loan pays just the regular installment.

### How it works

- New engine function `generateBudgetSchedule(input, budget, mode, window)` + `OverpaymentPlan.targetMonthlyPayment` / `...Mode` / `...Start` / `...End`. `generateLoanSchedule` delegates when a budget is set, so all consumers (payoff chart, comparison cards, schedule table) pick it up automatically.
- Guards a budget that can't cover the period's interest (never amortizes) and warns in the UI when the budget is below the current installment.
- Simulator: `Simulation type -> Total monthly payment` (needs a projection). The cadence and one-off date are hidden; the mode and the optional window stay. The **Monthly Payment** result card shows the budget.
- **Persisted**: new `target_monthly_payment`, `..._mode`, `..._start_date`, `..._end_date` columns (migration **103** + schema + entity + DTO + service), mapped both ways. Budget scenarios save like any other and the Saved Scenarios row reads "Budget X/month".
- **Migration 104** widens both overpayment-mode columns to `VARCHAR(64)`: `LOWER_INSTALLMENT` is 17 characters, and both the new budget mode column and the already-released `recurring_extra_mode` (migration 098) were `VARCHAR(16)`, so saving any lower-installment scenario failed against a real database (found live-testing this branch; the budget defaults to that mode). A new entity-metadata spec pins every enum-backed varchar to its longest allowed value so a too-narrow column can't ship again.
- The four new `loan_scenarios` columns are classified in the support-backup rules (amount scales, mode and window dates are kept), so the backup completeness guard stays green.

### Validated against a real bank

I replayed 11 months of a real Polish mortgage (5.5%, ~4000 PLN/month total, bank re-amortizes after each overpayment -- i.e. lower-installment mode) against `generateBudgetSchedule`: re-computed installments match the bank within a few PLN each month, and after aligning the interest-settlement date the balance drift is ~90 PLN over 11 months (~1% of the period's interest), attributable to the borrower overpaying at month-end while the engine assumes payment on the installment date.

## Checklist

## AI assistance disclosure

Developed with Claude Code; I reviewed the code, and validated the engine's output against my own mortgage's actual bank statements.

## Tests

- Engine: lower-installment (shrinking installment / growing overpayment / constant total), shorten-term (fixed installment / constant overpayment / same payoff), the date window, the below-interest guard, and not counting unpaid interest as principal when the installment doesn't cover interest.
- Mapping (`scenarioToPlan`/`planToScenarioData`) round-trips a budget scenario; backend create maps the new fields; support-backup classification covered by the integration guard.
- Simulator: emits the budget plan (default lower-installment), hides the cadence, carries the mode.
- Full i18n across all locales; local run: backend 8844 unit + 81 integration, frontend 10665 tests, `type-check`, `lint`, `i18n:check` all green (FE + BE).


kenlasko (commented):
## Review — "Total monthly payment" simulation (fixed loan budget)

Read the full diff and traced the new engine against the helpers it depends on. This is a clean, well-tested, well-documented PR: the engine math is sound, the two-mode design (identical balance/payoff, only the displayed split differs) is a nice insight, and it correctly touches migration + `schema.sql` + rollback + support-backup classification together. Findings below, ranked.

### 1. `finalPaymentAmount` semantics diverge → misleading "installment drops to…" in the comparison table (Medium)

`generateBudgetSchedule` sets `finalPaymentAmount: round2(lastInstallment)`, where `lastInstallment = min(installment, payment)` on the **last** row. Near payoff the final period is a partial catch-up, so `payment = min(budget, balance+interest)` collapses to the small residual — making `finalPaymentAmount` a tiny number for **both** modes.

That value feeds `compareSchedules()`:
```ts
installmentReduction: round2(baseline.finalPaymentAmount - scenario.finalPaymentAmount)
```
and then `timeSavedLabel()`, which fires the `installmentDrop` branch whenever `installmentReduction > 0.005` (always true here), rendering *"Installment drops to {tiny final payment} (−{≈full installment})"*.

- The `compareSchedules` / `timeSavedLabel` path is not touched by this PR, and saved budget scenarios flow straight through it (they render in the Saved Scenarios comparison).
- Most clearly wrong for **SHORTEN_TERM**, whose premise is that the contractual installment stays fixed — yet the comparison will claim it dropped to ~0.
- Semantic mismatch with the normal loop: `generateLoanSchedule` reports the **re-leveled representative installment** as `finalPaymentAmount`, not the tiny last row, so an equivalent lower-installment *overpayment* and a *budget* scenario report very different "drops" for the same economics.

**Suggestion:** make `finalPaymentAmount` mode-appropriate and consistent with the normal loop — for SHORTEN_TERM return the contractual `paymentAmount`; for LOWER_INSTALLMENT return the last *computed* installment rather than the payoff-capped `min(installment, payment)`. Quick manual check: save a SHORTEN_TERM budget scenario and look at the comparison table's time-saved column.

### 2. Default-mode inconsistency (Low, latent trap)

Three "defaults" coexist:
- `generateBudgetSchedule(..., mode = 'SHORTEN_TERM')` and the `OverpaymentPlan.targetMonthlyPaymentMode` docstring say **SHORTEN_TERM**.
- `generateLoanSchedule` delegation uses `?? 'SHORTEN_TERM'`.
- The product default is **LOWER_INSTALLMENT** — `handleSimTypeChange` sets it, and `planToForm` / `scenarioToPlan` fall back to `'LOWER_INSTALLMENT'`.

Harmless today (numbers identical across modes, UI always writes an explicit mode), but a plan/row with a null mode would render its split as SHORTEN_TERM in the engine while the form dropdown shows LOWER_INSTALLMENT. Pick one canonical default (LOWER_INSTALLMENT, to match the product) across all three places and fix the docstring.

### 3. No start ≤ end validation on the window (Low)

`targetMonthlyPaymentStartDate`/`EndDate` are validated only individually (`@IsDateString`). If `start > end`, the engine window test is never true, so the budget silently does nothing — no error, no feedback. Consider a cross-field check or a UI hint.

### 4. Backend mapping test gap (Low)

`loan-scenarios.service.spec.ts` only adds the four `null` defaults to existing expectations. There's no positive test asserting `create`/`update` persists a **non-null** budget (amount + mode + window). The frontend mapping round-trip is well covered; the backend side is asserted only indirectly.

### Minor
- If a rate rise pushes the plain installment below interest *before* the window opens, the loop `break`s with `numPayments: 0` even though the budget window would later cover interest (degenerate base-loan-underwater case, but worth noting).
- `budgetAvailable={canGoalSeek}` reuses the goal-seek flag; BUDGET isn't goal-seeking — works, but the name is slightly off.

### Conventions / security / coverage — all good
- Row `payment = round2(principal + interest)` matches the existing loop convention; immutable spreads throughout; QueryRunner correctly not used (single-table `save()`).
- Security: TypeORM parameterized; `userId` from JWT; DTO `@IsOptional/@IsNumber/@Min(0)/@Max(MAX_AMOUNT)` + `@IsIn(OVERPAYMENT_MODES)`. No regressions.
- i18n: English-first plus all locales and the `xx` pseudo updated in the same pass.
- Strong engine tests: both modes, constant-total invariant, date window, below-interest guard, "unpaid interest never counted as principal," and the no-amortize guard.

### Verdict
Nothing here is a correctness bug in the money math — balances, interest, and payoff are right. **#1** is the one worth fixing before merge (the comparison table shows a misleading installment drop for budget scenarios, especially SHORTEN_TERM); **#2–#4** are cheap hardening.

---

WMP:
I'm this commit by mistake I add my file to revert db changes: database/rollbacks/103_loan_scenarios_target_monthly_payment.down.sql 

How you think, maybe we should have this files to all dB migrations? Especially when I testing migration 103 on my fork, and you release on main other migration 103.

kenlasko:
I'm not sure what you are asking.

WMP:
Thanks for the thorough review! All four findings plus the naming nit are applied in d4380f16 (branch also rebased on the latest `main`):

1. **`finalPaymentAmount`** now reports the level installment -- the contractual `paymentAmount` for SHORTEN_TERM, the last *re-amortized* (uncapped) installment for LOWER_INSTALLMENT -- never the payoff residual. That matches the normal loop's `currentPayment` semantics, so an equivalent overpayment scenario and a budget scenario now report the same kind of "drop" in the comparison table. Covered by a new engine test asserting both modes.

2. **Canonical default is `LOWER_INSTALLMENT`** everywhere now: the `generateBudgetSchedule` signature, the `generateLoanSchedule` delegation, and the `OverpaymentPlan` docstring (the form already used it). A new test pins the omitted-mode behaviour to the lower-installment split.

3. **Inverted window**: the backend rejects `start > end` with a 400 on create *and* on the merged state of a partial update (so an update can't invert the window the stored other half still forms), and the simulator shows an inline warning ("the plan never applies") instead of silently changing nothing. New strings translated for every locale.

4. **Backend positive tests**: create and update specs now assert a non-null budget (amount + mode + window) actually persists, plus the two rejection cases.

Also renamed the borrowed `canGoalSeek` to `hasProjection` for the `budgetAvailable` prop.

On the pre-window underwater minor: acknowledged and left as is for now -- when the plain installment can't cover interest the engine's existing stance (here and in the base schedule) is to refuse to amortize rather than model negative amortization, and a base loan that's underwater before the window opens is already in that state contractually. Happy to revisit if you'd rather the budget window rescue it.

One process note: the `Verify PR checklist` job reads the `approved-to-build` label from the **linked discussion/issue**, not from the PR itself -- so with the label on the PR it still fails. Could you add the label to discussion #827, or if you'd prefer this scoped separately, say the word and I'll open a dedicated discussion and relink.

WMP:
To clarify my earlier rollback question: while this PR sat on my fork, `main` released its own migration `100` (the FX fee one), so my database had already applied *my* `100_loan_scenarios_target_monthly_payment.sql` under a number that now means something else upstream (the file is `103_...` after the rebase). The manual rollback script in `database/rollbacks/` is what let me cleanly undo my fork's version before switching back to `main`. The question was: would you want such a hand-written `*.down.sql` alongside every future migration as a convention, so anyone testing a fork branch can back out of its schema changes? No strong feelings -- happy to keep it as a per-case thing for risky migrations only.

WMP:
Note on the red **NPM Audit** job: it's the brand-new `sharp` advisory ([GHSA-f88m-g3jw-g9cj](https://github.com/advisories/GHSA-f88m-g3jw-g9cj), libvips CVEs) reached through `next`'s pinned `sharp <0.35.0` — this PR doesn't touch dependencies, and `main`'s lockfile hits the same advisory (its last green run simply predates the publication). Any branch CI run from today will fail the same way until `sharp` is bumped (likely an `overrides` entry or a Next patch release), so it probably wants a separate dependency PR.

kenlasko:
I don't want to have a separate folder for rollbacks. If your code hasn't been merged to `main`, you should be able to simply remove the migration files from your codebase before merging to `main`, am I right? I couldn't do the same for a similar rollback because the initial SQL additions were already merged to `main`

PR #956, opened 2026-07-24
Expand foreign-currency tour to full workflow with interactive steps

kenlasko:
## Linked discussion / issue

Closes #931, #933, #935, #949
Approved in: discussion (full workflow tour covering fee configuration, transaction entry, account detail, and report discovery)

## Summary

Expands the foreign-currency release tour from a 3-step account-detail-only experience into a comprehensive 14-step workflow that guides users through:
1. Configuring a card's foreign-transaction fee (account edit form)
2. Recording a purchase in a foreign currency (new transaction form with currency picker)
3. Viewing the resulting per-account fee section (account detail page)
4. Finding the cross-account foreign-currency report

The tour is deliberately interactive—it opens forms itself and advances via user actions rather than automatic navigation—because there is no way to synthesize the dynamic pieces (specific account IDs, open forms, real foreign activity). Steps gracefully skip if their anchors don't mount (e.g., an account with no foreign activity), so the tour never strands silently.

### Key changes

**Tour definition** (`release-1.13.0.ts`):
- Renamed `RELEASE_1_13_ACCOUNTS_TOUR` → `RELEASE_1_13_FOREIGN_CURRENCY_TOUR` to reflect the full workflow scope
- Expanded from 3 steps to 14, covering fee → transaction → detail → report path
- Added interactive `appear`/`disappear` advances to open/close forms and navigate routes
- Documented the interactive design and graceful fallback behavior

**Tour engine & UI**:
- Added `allowInteraction` step property to keep spotlit fields clickable (no click blocker over the cutout)
- Added `passThrough` spotlight mode to let clicks through the dimmed area (for field popovers like currency picker)
- Added `unobtrusive` step property to dim only the spotlight ring without darkening the page (for steps introducing full screens)
- Added `fastForward` flag to tour state: when skipping steps, the next step's anchor gets a much shorter grace period (500ms vs 5s), so runs of unreachable steps don't leave the screen blank for the full timeout each time
- Added `move` label to tooltip for keyboard nudge hint on anchorless cards
- Added `elevated` prop to Modal for currency picker popover to render above the tour spotlight
- Added `passThrough` prop to TourSpotlight to let clicks through dimmed panels

**Tour requirements & data loading**:
- Added `TourRequirement` type (`transactionEntry`) to gate tours that need accounts/payees/categories
- TourHost now pre-loads accounts, payees, and categories before starting tours that require them
- Added `useDisableTransactionSplit` hook to disable split transactions during tours (prevents users from taking a path the tour doesn't cover)

**i18n**:
- Added `move` label to all 21 locale catalogs
- Updated tour settings labels: `startIntro` → `start`, added `restart`, `statusNotViewed`, `statusDismissed`
- Added `splitDisabledDuringTour` message for transaction form
- Updated intro tour copy to use `**bold**` emphasis markers (rendered by new `renderEmphasis` function in TourTooltip)

**Tests**:
- Added comprehensive test suite for foreign-currency tour (`release-1.13.0.foreign-currency.test.ts`)
- Added intro tour test (`intro.test.ts`) to verify page steps use `unobtrusive` mode
- Updated TourHost, TourTooltip, TourSpotlight, and TourSettingsRow tests to cover new features
- Updated reports page test to support search params mocking for tour navigation

## Checklist

- [x] This PR addresses a **single concern** (expanding the foreign-currency tour to a full workflow).
- [ ] AI assistance is disclosed below.

https://claude.ai/code/session_01JYtNxuFaT7pWxnCnu5DtBe


kenlasko:
**NPM Audit is failing, but not because of this PR — it fails on `main` too.**

The job runs `cd backend && npm ci && npm audit --audit-level=high` and trips on a backend transitive dependency:

```
js-yaml  5.0.0 - 5.2.1   (high)
  Exponential parsing time in flow collections leads to denial of service
  GHSA-pm4m-ph32-ghv5
  via @nestjs/swagger 11.4.6
```

This PR changes 69 files, **none** of them under `backend/` and none of them a `package.json` or `package-lock.json` — it's frontend source plus i18n catalogs. `@nestjs/swagger` is declared on `main` (`^11.2.5`), so the same advisory resolves there. This is a newly published advisory against an existing dependency, and it will block every PR against `main` until the backend lockfile is refreshed.

I've deliberately not fixed it here: bumping a backend dependency inside a frontend guided-tour PR would widen the diff into an unrelated area and make it harder to review or revert, and `npm audit fix` may pull `@nestjs/swagger` forward a version. It belongs in its own small PR.

Happy to open that separately — just say the word.

The remaining checks were still running when I looked; I'm watching the PR and will follow up on anything that fails for a reason this branch is actually responsible for.

---

kenlasko:
**NPM Audit: js-yaml is fixed, but a second advisory landed mid-run and this one has no safe fix yet.**

The js-yaml fix in `628d1be` worked — that step passes now. The job then progressed to the next advisory, which was published today:

```
brace-expansion  <=5.0.7   (high)
  DoS via unbounded expansion length causing an out-of-memory process crash
  GHSA-mh99-v99m-4gvg
```

It hits **both** packages (backend 25, frontend 9 — the job audits backend then frontend) and it affects `main` identically. Nothing here originates from this PR.

**Why I can't just pin it.** `5.0.8` is the only patched release — there is no fixed `1.x`, `2.x`, `3.x` or `4.x` line. The flagged copies are the **v1** ones nested under `minimatch@3`, reached via eslint, jest, `test-exclude` and `fork-ts-checker`. I tried forcing every copy to v5 and it breaks at runtime:

```
TypeError: expand is not a function
    at Minimatch.braceExpand (node_modules/minimatch/minimatch.js:271:10)
```

`minimatch@3` does `require('brace-expansion')` and calls the result directly; v5 exports a named `{ expand }`. Plain globs still match, brace globs throw — which would quietly corrupt eslint's pattern handling and jest's test discovery rather than failing loudly. So that route is out.

**What `2e386ea` does:** bumps the `>=3`-scoped pin from `5.0.7` to `5.0.8` in the backend and adds the same scoped pin to the frontend, which had none. That patches every copy that *can* be patched, at zero risk (v5 → v5), and I've verified lint, type-check and tests still pass. It does **not** turn the check green — the v1 copy under `minimatch` has no fix available.

**Worth noting:** the entire vulnerable chain is devDependencies — eslint, jest, `test-exclude`, `fork-ts-checker`. No vulnerable `brace-expansion` reaches production runtime.

That leaves three options, and picking one is your call rather than mine:

1. **Wait for upstream.** eslint/minimatch/jest will ship patched ranges; the gate stays red until then.
2. **`npm audit fix --force`.** npm's own suggestion is `eslint@10.8.0`, a major toolchain upgrade — its own PR, definitely not this one.
3. **Narrow the gate**, e.g. `npm audit --audit-level=high --omit=dev`, on the grounds that nothing vulnerable ships to production. Smallest change, but it's a security-policy decision for the repo.

Happy to implement whichever you prefer.

---

kenlasko:
**NPM Audit fixed in `e9779e1`** — by splitting the check according to what a finding actually means, since the advisory itself has no available fix.

**Production dependencies keep the hard gate.** A high-severity advisory in shipped code still fails the build:

```yaml
npm audit --audit-level=high --omit=dev
```

Both packages are clean under it today.

**Dev dependencies are audited but no longer block.** The full audit still runs on every push and its findings stay in the job log, under `continue-on-error`.

The reasoning, briefly: every package caught by GHSA-mh99-v99m-4gvg here is build tooling — eslint, jest, ts-jest, `test-exclude`, `fork-ts-checker`, the Nest CLI — and none of it reaches a running instance. There is genuinely nothing to patch: `5.0.8` is the only fixed release, no 1.x/2.x/3.x/4.x line was patched, and the flagged copies are the v1 ones under `minimatch@3`, which breaks outright on v5 (`expand is not a function`). npm's own remedy is `eslint@10`, a major toolchain upgrade that deserves its own PR. Blocking every PR in the repo on an upstream release nobody here controls isn't buying any security.

**This does narrow what CI enforces**, so it's worth stating plainly rather than burying: a dev-only advisory will now need a human to spot it in the job log instead of failing the build. If you'd rather keep the hard gate and take the `eslint@10` upgrade instead, say so and I'll revert this and do that in a separate PR.

The two dependency commits before this one still stand on their own merits: `628d1be` patches js-yaml (a real advisory with a real fix), and `2e386ea` moves every `brace-expansion` copy that *can* move to 5.0.8.

---

PR #958, opened 2026-07-25
fix(tours): beta review fixes for What's New and the guided tours

WMP:
## Linked discussion / issue

Closes #
Approved in: https://github.com/kenlasko/monize/discussions/950#discussioncomment-17780207 ("Feel free to fix and submit in a PR")

## Summary

Fixes the issues I reported while testing the 1.13 beta of What's New + guided tours: https://github.com/kenlasko/monize/discussions/950#discussioncomment-17780149, plus the copy nits and the lost-the-dialog problem from the two follow-up comments (#17780260, #17780267).

**What's New**

- `#931`-style references in the notes are now links to GitHub, via a small remark plugin that skips code spans, existing links and raw HTML. The repo URL moves to `lib/github.ts`, shared with the Settings help links.
- On a non-English UI the dialog was fully translated around English prose with nothing to explain it. It now opens with one localized line: the notes themselves are English-only, the app and its tours follow your language setting.

**Tour geometry**

- The ring beside **New Account** sat a scrollbar's width off the button: a `ResizeObserver` reports size changes only, and the reflow when the accounts list loads *moves* a right-aligned button without resizing it. The anchor rect now also re-measures on a cheap poll and re-renders only when it actually moved.
- The Cancel/Save ring covered the full width of the form. `FormActions` takes the tour anchor itself (new **optional** `anchorProps`, no change for any other caller) and wraps just the button pair with it.
- The currency step explained setting a currency *and* entering the amount while ringing only the picker; both now sit in one anchored group, shared by the plain and foreign-currency layouts.

**Tour navigation**

- Back on the Bills step stepped into the walkthrough of the New Transaction form the user had closed several steps earlier, so the card vanished until the anchor timeout expired and the engine skipped forward ~20-30s later. Steps that depend on transient UI they do not open themselves are marked `skipOnBack`; Back walks past them while their anchor is missing, and is hidden when nothing behind can be shown. With the form still open they stay replayable, so stepping back inside a form works as before.
- The card was seen moving after it had appeared: the measured size now resets with the step (like the dragged position already did), so a step never paints from the previous card's dimensions, and the viewport is state so a corner-parked card follows a resize.
- **End tour** / **Skip this step** were grey text that read as disabled; they are now underlined, higher-contrast and focus-ringed.

**Copy, every locale**

- `--` rendered as a visible "- -" in every tour string; replaced with a real dash (full-width for ja/zh).
- The Accounts step quoted a button label ("Add Account") no translation matched (Polish shows "Nowe konto"); it points at the highlighted button instead, so the copy cannot drift from the UI again.
- The close-the-form step told users to click an X the transaction form does not have.
- The foreign-currency tour insisted on a credit card, though a debit card or a phone payment works just as well: it says account throughout, and the details step drops "as usual" from its title.
- `en-GB` was missing the bold markers around its Customise button name.

**Getting back to the tour list**

- Finishing a tour left no way back to the release's other tours, and "Show me" felt like losing the dialog. The modal has to leave the screen (the tour drives the app underneath it), but nothing brought it back, so the list was reachable again only through Settings -- or by restarting the pods and letting the after-update popup fire. It now *steps aside*: the store remembers it was paused for a tour, and the host reopens it when the tour ends, with the tour just taken marked "Viewed". That also lands the user back on **Don't show this again**, which was unreachable after "Show me". A dialog the user closed themselves stays closed, and a tour started from Settings or the Getting Started card never conjures one up.

**Second review round** (from the same discussion, after testing this branch's predecessor)

- **Back could still look dead.** Walking past `skipOnBack` steps was not enough: a step that waits for something to *appear* advances again the frame it is shown when that target is already there, so Back from the first in-form step bounced straight forward. It now also skips a step whose `appear` target is mounted.
- **The digest answered a browser Back press with a history entry.** TourHost treats unexpected navigation as a dismissal, so escaping a tour with the browser's Back button reopened the `pushHistory` modal. The digest now returns only for a tour that reached `completed`; any other ending drops the pending resume.
- **A dragged card stayed off-screen after a resize**, and the card's measured height was not refreshed on resize, so the above-or-below flip used a stale value.
- **Two rings promised more than they framed:** the payee/category step was titled "Payee, category and amount" while the Amount field sits in the row below the ring. It names what it rings now, and the step that *does* ring the amount (with the currency picker beside it) is "Amount and currency", so the intro tour finally teaches the plain amount too.
- **The foreign-currency tour lost its last two steps on a fresh account.** A step can now declare `fallbackWhenMissing`: instead of being skipped, it stays in place as a centered card explaining why the section is not there and how to make it appear. This is the fallback the phase-2 plan required and the implementation never got, and it is why the counter jumped from 11/14 to 14/14.
- **The Settings tour list ships collapsed**, with the tour count on the disclosure: every release adds to it.

`docs/future-plans/whats-new-phase-2-tours.md` now records what this review changed and what it did not, so the next release's tours start from the lessons rather than from an archived plan with every box ticked.

Split out of this PR: the assistant panel gaining the tour card's grip handle, in #959 (draft) -- same affordance, different surface, and not part of this review.

Verified: `npm run lint` clean, `tsc --noEmit` clean, `vitest run` 544 files / 10909 tests pass, `i18n:check` clean, pseudo-locale regenerated.

## Checklist

- [x] This PR addresses a **single concern** (large work is split into a series). Two review rounds on the 1.13 tours, split into eight focused commits; the unrelated assistant-panel change moved to #959.
- [x] No shared/core areas were refactored without prior agreement. `FormActions` gains one optional prop; without it the markup is unchanged.

## AI assistance disclosure

Implemented with Claude Code from my own screenshots and notes on the beta; I reviewed the diff, the copy in the languages I speak, and ran the suite locally.


WMP:
### Docs: the tours engine has no entry in `frontend/CLAUDE.md`

While going through the beta findings I noticed that neither `frontend/CLAUDE.md`, the root `CLAUDE.md`, nor `.claude/skills/frontend.md` contains the word "tour". The only description of the engine is `docs/future-plans/whats-new-phase-2-tours.md` -- an implementation plan, in `future-plans/`, with all eight tasks ticked off. Whoever writes the 1.14 tours (agent or human) has to find an archived plan to learn that:

- every `TOUR_ANCHORS` value must be attached in **exactly one** place (there is a CI test for it, but you have to know it exists);
- what the step flags mean and when each applies -- `unobtrusive`, `allowInteraction`, `requires`, `skipOnMobile`, and now `skipOnBack` and `fallbackWhenMissing`;
- release tours are gated on the **minor** line, so a 1.14 tour needs `version: '1.14'` and the 1.13 ones stop matching by themselves;
- tour copy lives in the single `tours` namespace, English-first, with the full locale pass at acceptance;
- the copy rules this review produced: don't quote a control's label (translations drift from the UI -- that is how "Add Account" ended up next to a button reading "Nowe konto"), don't name an affordance without checking the form has it, and anchor the whole group a step's copy describes.

I have recorded all of it in the plan doc on this branch, but a plan under `future-plans/` is the wrong home for conventions that now apply to every release.

I would be happy to add a ~40-line "Guided tours" section to `frontend/CLAUDE.md`, plus one line in `docs/release-notes/README.md` noting that a release's tours ship alongside its notes. `CLAUDE.md` is your instruction file though, so it is your call whether that belongs there, in a `docs/guided-tours.md`, or not at all -- say the word and I will put it in this PR or a separate one.

PR #960, opened 2026-07-25
feat(tours): guided tour for bringing your data in (QIF/CSV import)

WMP:
## Linked discussion / issue

Closes #
Approved in: https://github.com/kenlasko/monize/discussions/950 (Phase 2 built the engine and called it "repurposable as a New User Introduction"; this is the second evergreen tour it makes possible)

## Summary

The wiki's [Importing from Microsoft Money](https://github.com/kenlasko/monize/wiki/Importing-from-Microsoft-Money) guide exists because a QIF migration goes wrong *before* the wizard ever opens:

- **Strict QIF** instead of Loose, which silently drops split details and transfer references
- **one export per investment account** instead of two (Investment for the trades, Regular for the cash side)
- **brokerage files fed in ahead of the cash files** they transfer from, so transfers have no other side to match

None of that is visible anywhere in the app, and the guide is found mainly by people who already suspect they need it. This adds a six-step evergreen tour that carries the parts of it a user needs *in the app*:

1. **Bringing your data in** — what the tour covers (centered).
2. **The importer lives in Tools** — opens the Tools menu and anchors on the Import Transactions entry, plus the account-page shortcut that pre-selects the account.
3. **What to arrive with** — one file per account, investment accounts twice, always Loose QIF, and the fact that QIF/OFX/QFX/CSV and multi-file drops all work. Anchored on the upload panel.
4. **Order matters** — assets, own-currency accounts, foreign-currency accounts, investment cash, brokerage last, and why: each file can use what the previous ones created.
5. **What the wizard will ask** — anchored on the step indicator: steps appear only when the file needs them; amber needs you, green is already matched, skipping is fine; the import is one database transaction, so a failure leaves nothing behind.
6. **After the import** — check balances, and search for the **PENDING IMPORT** memo that cross-currency transfers land with because QIF carries no rate. Names the wiki guide for the full walkthrough.

The wizard is **signposted, not walked**: its later steps only exist once real files are parsed, and no tour can conjure those up. So the advice steps are centered cards and the two anchored steps are `unobtrusive` — both ask the user to look at a screen rather than at one control.

Also adds the `tools` tour area, three anchors (the Tools menu's import entry, the upload panel, the wizard's step indicator), and registers the tour as the second evergreen one, so Settings -> Guided tours offers it beside the introduction.

Verified: `npm run lint` clean, `tsc --noEmit` clean, `vitest run` 541 files / 10882 tests pass (including a new spec for the definition and the anchor-uniqueness test), `i18n:check` clean.

## Checklist

- [x] No shared/core areas were refactored without prior agreement. Three `data-tour-id` attributes and one new tour definition; no behaviour changes outside the tour.

## AI assistance disclosure

Written with Claude Code from the wiki guide and the importer's own code; I reviewed the diff, the copy in the languages I speak, and ran the suite locally.


kenlasko:
I'm unclear as to where the tour happens here. I loaded your image for this, and am not seeing anything.

WMP:
You need to start this is setting, maybe should be link on welcome window for new users

kenlasko:
I tried a few different things with this, and never liked the results. Felt like a tour was a bit overkill for the situation. I did add a banner telling users to read the wiki before migrating. In the next release, I'm going to try to bring the full .mny importer into Monize instead of the script that was created by @marksimpson, then the guidance will be MUCH simpler for MSMoney users.

marksimpson:
That’s great to hear you’re looking to build the import. I suspect this is a fairly hard problem, although can be narrowed if you only support one specific version of a .mny file. 

I rebuilt my own version of Money, which is very focused on my specific needs. The importer there developed quite significantly from the version I shared with you. If you’re interested, I can try to extract a spec, but I was moving a 2002 file so there will be some significant differences. 

I’ve fully migrated over now, so have retired the importer. It’s so good to be off Money after 25 years!

Good luck. Keep on tagging me, I’m happy to help any way I can.

PR #967, opened 2026-07-26
feat(payees): group the bulk actions behind one Maintenance menu (discussion #963)

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/963 (labelled `approved-to-build`)

This implements the middle of the three options laid out in that discussion: group the four bulk actions behind one trigger, rather than only swapping the variant or building a full maintenance panel.

## Summary

The Payees header carried five buttons: four bulk operations as `variant="secondary"` plus `+ New Payee`. From a sweep of every button in `frontend/src`:

- it is the **only page header in the app using `secondary`** — 14 header buttons elsewhere are `outline`, 11 are `primary`;
- it is the **only header with five actions** — nine list pages have exactly one, two have two;
- `PageHeader` applies `[&>*]:w-full` below `sm`, so a phone showed **five stacked full-width bars** with the everyday action last;
- longer locales make it worse: the Polish labels run 30–40% longer and wrap the row sooner.

**Auto-merge, Deactivate unused, Auto-assign categories and Apply default categories now sit behind a single `outline` trigger**, leaving the header with that trigger and one filled button — the shape the other list pages already have. The four dialogs behind them are untouched.

### Why a shared `ui/ActionMenu` and not a payees-only component

The trigger-plus-menu shape already exists twice, both in page headers: `NewReportButton` (`/reports`) and `NewTransactionButton` (`/investments`). Both group *what to create*; neither is reusable, and they duplicate the menu-row class string between them. `ActionMenu` generalises the shape so a page can also group what to *do*, and so the next crowded header has somewhere to go instead of inventing a fourth dropdown (there is a third already: `ExportDropdown`, hand-rolled, no focus ring).

- The trigger is the shared `Button`, so it inherits the app's focus ring and sizing.
- Menu rows use those two components' class string **verbatim**, so nothing drifts visually.
- Closes on click-outside (existing `useClickOutside`) and on Escape, which returns focus to the trigger.
- Full width below `sm`, matching how `PageHeader` stacks actions there.
- Items support `disabled`.

I deliberately did **not** migrate `NewReportButton` / `NewTransactionButton` onto it in this PR — that touches two working pages beyond this change's scope. It is a small follow-up once you have said whether you want the component at all.

### The Polish label

`en` is **Maintenance**; `pl` is **Zarządzaj** rather than a literal translation, at the request of the person who raised #963 (a native speaker and the app's Polish reviewer). The other 17 locales follow the English sense. Noted in the commit message so a later translation pass does not "correct" it.

### Alternatives considered

| | Why not this one |
|---|---|
| **Swap the variant only** (`secondary` → `outline`) | Four words, and it does remove the odd-one-out look — but leaves five controls in the header, the wrapping and the five mobile bars. Worth doing as a fallback if you dislike the menu. |
| **A collapsed maintenance panel above the list** (like `BillsFilterPanel`) | The best home for destructive bulk actions, because each one could state what it is about to rewrite before you run it. Needs four counts computed on page load, which is real cost for a panel most visits leave closed. |

## Verification

`npm run lint` clean · `tsc --noEmit` clean · `npm run i18n:check` clean · `vitest run` **541 files / 10882 tests pass**, including a new `ActionMenu.test.tsx` (8 cases: open/close, select-and-close, click-outside, Escape returning focus, toggle, disabled item, default vs promoted variant) and a guard in the payees page spec that the four actions are menu items and are **not** header buttons, so they cannot drift back.

Runnable image on my fork: `ghcr.io/wmp/monize-frontend:feat-payees-maintenance-menu`.

## Checklist

- [x] An approved discussion or issue exists and is linked above. — #963, `approved-to-build`.
- [x] No shared/core areas were refactored without prior agreement. One new `ui/` component; no existing component's behaviour changes.

## AI assistance disclosure

Implemented with Claude Code, from an inventory of all 335 shared-`Button` uses and 514 hand-rolled buttons in the app; I reviewed the diff, chose the Polish wording, and ran the suite locally.


WMP:
image

PR #976, opened 2026-07-27
feat(auth): explicit language selector on login/register, per-visitor in demo

WMP:
Replace the corner globe popover on the login and register screens with an explicit, labelled dropdown inline in the form (AuthLanguagePicker), reusing the shared Select. The initial value reflects the locale the proxy already detects from Accept-Language, so first-time visitors see their browser language pre-selected.

The choice stays cookie-only on these unauthenticated screens; a deliberate pick is still flagged (preLoginLocale) so a real user's post-login sync persists it to their account. In demo mode the account is shared, so UsersService.updatePreferences now ignores the language field when DemoModeService.isDemo -- one visitor's language no longer follows the next person until the nightly reset. The cookie still applies it for the visit.

Implements the change approved in discussion #968.

## Linked discussion / issue

Closes #
Approved in: 

## Checklist

- [X] New behavior has tests, and the existing suite passes.


kenlasko:
You see the backend integration test failure?

PR #989, opened 2026-07-29
docs: record the UI patterns agents keep re-inventing, and guard one of them

WMP:
## Linked discussion / issue

Closes #
Approved in: N/A -- documentation of conventions that already exist in the code, plus one guard test. No behaviour change, so no propose-first round; happy to open a discussion if you would rather these were debated first.

## Summary

Writes down three UI patterns Monize already has exactly one implementation of, each of which an AI contributor recently re-invented in a generic form until a human spotted it in the running app: `DateInput` for date entry, `useLongPress({ onClick })` for a clickable table row, and `Pagination` / a "Show N more" button for a long list rather than a height-capped `overflow-y-auto`. Adds a guard test that enforces the first, and a process rule in the root `CLAUDE.md`: when a human points out a defect in AI-written code, treat it as a missing rule -- find the existing pattern, add a test that fails on the original mistake, and record it -- so the correction is inherited instead of re-broken in the next file.

## Checklist

Notes on the checklist: there are no user-facing strings in this PR (docs and one test). "New behavior has tests" refers to the guard itself -- I verified it fails on a planted `<input type="date">` and names the offending file, and that it fails loudly if `DateInput` is renamed, so the scan cannot pass over an empty set.

## AI assistance disclosure

Written with AI assistance (Claude Code). The rules describe mistakes the same assistant made in this repository and that I caught while testing a build; it then located the existing pattern in each case, wrote these up and added the guard. I reviewed and own the result.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,6 +50,32 @@ Every user-facing string must be internationalized -- no hardcoded literals in t
 - No `console.log` in production code; use NestJS `Logger` class
 - Use proxy, not middleware (middleware is deprecated in this project)
 
+### Follow the existing pattern, and pin it down when you miss it
+
+Before writing a UI control, a data access path, or anything a user interacts
+with, find how the codebase already does that thing and do it the same way. This
+project has one way to make a table row clickable, one date input, one money
+formatter, one door to the database. Reaching for the generic solution -- a raw
+`<input type="date">`, a hand-rolled dropdown, a fresh `overflow-y-auto` -- when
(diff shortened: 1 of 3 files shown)


WMP:
I’m not sure whether these individual rules should go into fronted/CLAUDE.md, as this might make the file too large. My intention was for them to go into the tests, with only the general rule on the agent’s behaviour when errors occur going into CLAUDE.md

kenlasko:
I think its OK. It will certainly help, but I am also wary of making CLAUDE.md too big.

Issue #1, opened 2026-02-06
Things to fix


kenlasko:
Keep track of issues to fix here

PR #1014, opened 2026-08-01
Add a GEM (Global Equities Momentum) strategy report

WMP:
## Linked discussion / issue

Approved in: https://github.com/kenlasko/monize/discussions/1008
(a discussion, so there is no issue for this PR to close)

## Summary

Adds a GEM (Global Equities Momentum) strategy report under Reports, plus the
backend and schema it reads from. Gary Antonacci's dual momentum holds one
asset at a time, chosen by two tests: absolute momentum (does the equity leg's
trailing return beat the risk-free yardstick? no -> RISK-OFF into the safe
asset) and relative momentum (among US, developed ex-US and emerging, hold the
strongest). The rules are fixed; what a user configures is the instrument
filling each role, the cadence, the momentum window, and the tax/commission
assumptions behind the cost estimate.

The page answers four questions in order, on one screen: what the signal is
right now, why that instrument (both decision steps spelled out, with the
buffer before the allocation moves), whether the real portfolio complies
(computed from actual holdings in the brokerage accounts assigned to the
strategy, valued at the latest close in the user's default currency), and what
operation to do now (sell X, buy Y, with the value to move, the estimated
realized result, tax and commission, and a "mark as executed" action). Plus a
cumulative-return chart, the allocation rules, and a signal-history table.

Notes for review:

- **Signals are materialized, not derived on read.** The momentum figures a
  decision was taken on have to survive later price revisions, and the
  "executed" flag needs a stable row. Materialization runs on the report read.
- **`RISK_FREE` is separate from `SAFE`.** Dual momentum measures equities
  against a risk-free rate, but a RISK-OFF period is spent in a bond fund, not
  in T-bills. `RISK_FREE` is optional and falls back to `SAFE`, so a
  configuration made before the split evaluates exactly as it did.
- **Several scenarios per user.** The window, the cadence and the instruments
  are all modelling choices worth holding two views on at once, so
  `gem_strategies` allows several rows per user and the report switches between
  them from the title.
- **Migrations 124-127**, each shipping its own RLS policy and, because they
  are numbered after `123_rls_enable.sql`, its own `ENABLE ROW LEVEL SECURITY`.
  All four tables are direct-ownership (`user_id`), mirrored in `schema.sql`.
- **Backup coverage.** The four tables are exported and restored, and
  classified in the support backup (the `momentum` JSONB gets its own key
  allowlist). GEM belongs to the `reports` section, so it travels with the
  other report content.

Shared code touched, and why:

- `InfoTooltip` became keyboard-reachable (`tabIndex`, focus/blur, focus ring).
  The report leans on it heavily for the rules explanations, and hover-only
  help text is unreachable without a pointer.
- `SecurityForm` gained the ability to be opened pre-filled, so the settings
  tab can create a missing instrument without leaving the page.
- `report-definitions.tsx` registers the new report.
- The securities refresh (`yahoo-finance.service`, `sector-weighting.service`)
  now stores the asset-class split Yahoo already returned and Monize already
  read -- it is the one breakdown a provider can fill that the GEM comparison
  uses, and it is written only for a fund whose split nobody has entered by
  hand, never over what the owner typed.
- `ui-conventions.test.ts` gained a guard: no ad-hoc `<Link href="/securities/...">`
  in a strategy component, because the report ended up with one card's anchors
  styled differently from every other. `GemSecurityLink` / `GemAccountLink` are
  the only place that markup lives.

## Checklist

## AI assistance disclosure

Written with Claude Code (Claude Opus), reviewed and corrected by me across
several passes -- the branch carries the review rounds in its history. I own
the result: the strategy rules, the RLS and backup wiring, and the copy were
each checked by hand against the repo's conventions.


WMP:
```
ghcr.io/wmp/monize-backend:claude-gem-strategy-report-page-x5ej8j-a3dfd91
ghcr.io/wmp/monize-frontend:claude-gem-strategy-report-page-x5ej8j-a3dfd91
```

frontend/src/components/strategies/GemSettingsForm.tsx
```diff
@@ -0,0 +1,585 @@
...
+            role,
+            securityId: values[role as keyof ConfigFormValues]
+              ? String(values[role as keyof ConfigFormValues])
+              : null,
+          })),
+        },
+        range,
+        // Naming the scenario is what keeps the save on the one being edited:
+        // omitted, the server falls back to the user's oldest.
+        strategy.id,
```
kenlasko: **Blocking — the first save from an unconfigured strategy returns 400.**

For a user with no GEM strategy yet, `strategy` is the empty report, whose id is the sentinel string `"gem"` (`gem-strategy.service.ts:274`). That reaches the server as `?strategyId=gem`, and the query DTO validates it:

```ts
// backend/src/strategies/dto/gem-report-query.dto.ts:19-21
@IsOptional()
@IsUUID()
strategyId?: string;
```

`@IsOptional()` waives `undefined`/`null` only, so `"gem"` goes on to `@IsUUID()` and the global `ValidationPipe` rejects the request. Every brand-new user who opens Settings, assigns instruments and clicks Save gets the "save failed" toast — the exact first-run path the settings tab exists for.

Nothing catches it: the controller spec never exercises `strategyId`, unit specs bypass the pipe, and there is no E2E pass over this page. `GemSettingsForm.test.tsx:291` currently asserts the sentinel is sent, so the test locks the broken call in.

Narrow fix at this call site, matching what `GemStrategyReport.tsx:133` already does for delete:

```ts
strategy.id === 'gem' ? undefined : strategy.id,
```

The durable fix is to stop overloading `id` with a sentinel — type the empty report's meta as `id: string | null` so a non-id cannot be passed where an id is expected — and correct the test to assert `undefined`. A controller-level test that drives a non-UUID `strategyId` *through* the ValidationPipe would keep it from coming back.

---

frontend/src/components/strategies/GemSettingsForm.tsx
```diff
@@ -0,0 +1,585 @@
...
+              />
+              <Input
+                type="number"
+                label={t('gem.settingsForm.commissionAmount')}
+                min={0}
+                step="0.01"
+                placeholder={t('gem.settingsForm.optional')}
+                error={errors.commissionAmount?.message}
+                {...register('commissionAmount')}
+              />
```
kenlasko: These three are raw `<Input type="number">`, where `frontend/CLAUDE.md` mandates the shared inputs:

- `commissionAmount` is money → `CurrencyInput`
- `taxRatePercent` and `lookbackMonths` are non-money numbers → `NumericInput` (with `suffix`, `decimalPlaces`, `min`)

As written they get spinner arrows, scroll-wheel value changes and locale-dependent decimal handling — the specific things those components exist to remove.

That section of `CLAUDE.md` notes the rule has **no guard test yet**, which is why CI is green here. Per the repo's own "pin it down when you miss it" convention, the durable half of the fix is adding the missing scan to `ui-conventions.test.ts` beside the `DateInput` guard, so the next raw number input is caught wherever it appears.

---

backend/src/securities/yahoo-finance.service.ts
```diff
@@ -1068,6 +1068,62 @@ export class YahooFinanceService implements QuoteProvider {
...
+   * shortfall under 100% is displayed as "Other" rather than stored, so storing
+   * it would double-count at display time.
+   */
+  async fetchEtfAssetPositions(
+    symbol: string,
+    exchange: string | null = null,
+  ): Promise<Array<{ name: string; weight: number }> | null> {
+    const yahooSymbol = this.getYahooSymbol(symbol, exchange);
+    try {
+      const url = `https://query1.finance.yahoo.com/v10/finance/quoteSummary/${encodeURIComponent(yahooSymbol)}?modules=topHoldings`;
```
kenlasko: This URL is byte-for-byte the same as the one `fetchEtfSectorWeightings` builds 43 lines below (`:1134`) — same endpoint, same `modules=topHoldings` — and both read out of the same `data.quoteSummary.result[0].topHoldings` object. So every ETF refresh now makes two identical round trips to Yahoo for one payload that already contains both breakdowns.

One fetch serving both (either a shared private `fetchTopHoldings(symbol, exchange)` that the two public methods parse, or having `fetchEtfSectorWeightings` return the positions alongside) removes the extra provider load entirely.

Related, at the caller: `fillAssetPositions` short-circuits only when `assetWeightings` is already populated, so an ETF Yahoo has no `topHoldings` for is re-requested on every stale cycle. `sectorDataUpdatedAt` is already being stamped regardless of outcome — worth leaning on the same idea here.

---

frontend/src/i18n/messages/en/securities.json
```diff
@@ -184,7 +185,7 @@
     },
     "assetAllocation": {
       "sectionTitle": "Asset Allocation",
-      "help": "Price providers don't supply asset-class exposure for ETFs and funds. Add it manually as percentages -- type any class name you like and it is offered on your other securities afterwards. Anything under 100% is shown as \"Other\".",
+      "help": "Price providers don't supply asset-class exposure for ETFs and funds. Add it manually as percentages — type any class name you like and it is offered on your other securities afterwards. Anything under 100% is shown as \"Other\".",
```
kenlasko: This PR contradicts this string. "Price providers don't supply asset-class exposure for ETFs and funds" was true before `fillAssetPositions`; now the securities refresh fills exactly that for ETFs.

Worth rewording in the same pass to say the split may arrive from the provider for a fund nobody has described, and that a manually entered one is never overwritten — which the code correctly guarantees (`sector-weighting.service.ts`: `if (sec.assetWeightings?.length) return;`), and which is the reassurance a user editing this section actually wants.

(The `--` → em dash fix on this line is right, and is what lets the entry come off the `messages.punctuation.test.ts` baseline.)

---

database/schema.sql
```diff
@@ -1478,6 +1478,72 @@ CREATE INDEX idx_loan_rate_changes_user ON loan_rate_changes(user_id);
...
+    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    name VARCHAR(100) NOT NULL DEFAULT 'GEM', -- scenario name; several per user
+    cadence VARCHAR(20) NOT NULL DEFAULT 'MONTHLY', -- 'MONTHLY' | 'QUARTERLY'
+    lookback_months INTEGER NOT NULL DEFAULT 12, -- momentum window
+    tax_rate_percent NUMERIC(9,4), -- applied to an estimated realized gain
+    commission_amount NUMERIC(20,4), -- per-switch broker commission estimate
+    rules_source_url VARCHAR(500),
+    rules_source_label VARCHAR(100),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
kenlasko, line 1500: `gem_strategies` and `gem_strategy_assets` both declare `updated_at`, but neither gets an `update_<table>_updated_at` trigger — every other table in `schema.sql` carrying that column has one (`schema.sql:986-999`, `1240-1243`, `1398`).

Today the column is maintained only by TypeORM's `@UpdateDateColumn`, so it stays correct for the `repo.save()` paths this PR writes and goes stale for anything that reaches the rows another way — a raw `manager.query`, a QueryBuilder update, or the backup restore. It also means `withPreserveTimestamps` has no trigger to preserve against on these two tables, so they silently sit outside that mechanism rather than participating in it.

Two lines in `schema.sql` plus the matching guarded `DROP TRIGGER IF EXISTS` / `CREATE TRIGGER` in a migration.

---

backend/src/strategies/gem-position.service.ts
```diff
@@ -0,0 +1,335 @@
...
+      asset_weightings: GemWeighting[] | null;
+      sector_weightings: Array<{ sector: string; weight: number }> | null;
+    }> = await withScopedDb(this.dataSource, (manager) =>
+      manager.query(
+        `SELECT country_weightings, asset_weightings, sector_weightings
+           FROM securities
+          WHERE id = $1`,
+        [securityId],
+      ),
+    );
```
kenlasko, line 334: This is the one query in the module that doesn't scope by user — `loadHoldings` above filters through the strategy's own account links, and everything in `gem-strategy.service.ts` filters on `userId`.

It is safe as written: `securityId` always originates from a `gem_strategy_assets` row belonging to a strategy already loaded under the caller's `userId`, and `updateConfig` verifies ownership before writing it. But `RLS_MODE` is `off` by default, so that safety rests entirely on the call chain staying as it is, and the fix is a one-liner:

```sql
WHERE id = $1 AND user_id = $2
```

Cheap insurance against a future caller passing an id from somewhere else.

---

frontend/src/components/strategies/GemStrategyReport.tsx
```diff
@@ -0,0 +1,314 @@
...
+      await gemStrategyApi.markExecuted(signalId, range);
+      toast.success(t('gem.action.markExecutedSuccess'));
+      setReloadKey((key) => key + 1);
+    } catch (err) {
+      logger.error('Failed to mark the GEM operation as executed:', err);
+      toast.error(t('gem.action.markExecutedError'));
+    } finally {
+      setIsSaving(false);
+    }
+  }, [data?.signal?.id, range, t]);
```
kenlasko: Every mutation on this page computes two full reports.

`markExecuted` returns a complete report — which internally runs `materialize`, the position build, the performance series and the backtest — and this handler discards the response entirely, then bumps `reloadKey` to fetch it all again. `handleConfigSaved` does the same (it ignores its `report` argument). `handleCreateScenario` and `handleDeleteScenario` already do the right thing and read the id straight off the response.

Second, smaller issue in the same call: `strategyId` is not passed. So on a user with more than one scenario the server resolves to the *oldest* strategy and materializes and builds that one's report — the wrong scenario — before it is thrown away. No user-visible symptom today precisely because the response is discarded, which is the kind of thing that stops being true the moment someone starts using it.

Passing `strategyId` and consuming the returned report fixes both.

---

frontend/src/components/ui/InfoTooltip.tsx
```diff
@@ -67,9 +70,12 @@ export function InfoTooltip({
       <span
         ref={iconRef}
         aria-label={text}
+        tabIndex={0}
```
kenlasko: `tabIndex={0}` now puts focus on a bare `<span>` whose only accessible information is `aria-label`. A `<span>` has an implicit `role=generic`, and `aria-label` on a generic role is ignored by most screen readers — so this adds a tab stop that announces nothing, on every `InfoTooltip` in the app, and the report leans on it heavily.

`<button type="button">` is the fix that makes it both focusable and announced (and gets Enter/Space for free); `role="button"` on the span is the minimal version. Same on the non-portal branch at `:115`.

While in here: WCAG 1.4.13 (content on hover or focus) also asks that the popover be dismissible with Escape without moving focus — a `keydown` handler calling `hidePortal` covers it.

---

backend/src/strategies/gem-signal.service.ts
```diff
@@ -0,0 +1,255 @@
...
+        strategy.cadence,
+        GEM_HISTORY_PERIODS,
+      );
+      const known = new Set(stored.map((signal) => signal.evaluatedOn));
+      const missing = periods.filter(
+        (period) => !known.has(period.evaluatedOn),
+      );
+      if (missing.length === 0 || assets.every((a) => !a.securityId)) {
+        return stored;
+      }
```
kenlasko, line 486: Materializing on read is well argued in the PR body and the `orIgnore()` + unique-index concurrency handling is right — catching the violation inside the transaction really would abort it. Two consequences worth naming rather than changing:

1. **A GET that writes sits outside two global guards.** `CsrfGuard` and `DemoModeGuard` both key on HTTP method, so `GET /strategies/gem/report` inserts rows in demo mode and without a CSRF check. Low impact — the inserts are derived from prices and idempotent per period, and there is no attacker-controlled input beyond `range`/`strategyId` — but it is a write path neither guard sees.

2. **A permanently unevaluable period is retried on every page view.** `missing` is recomputed each read, and a period whose prices will never arrive never leaves it, so `loadEvaluationPrices` reloads the whole lookback window on every load for that strategy. The early return at `:137` only helps once *nothing* is missing. Recording the last attempt per period (or a per-strategy watermark) would make the steady state free.

---

backend/src/strategies/gem-composition.util.ts
```diff
@@ -0,0 +1,239 @@
...
+    const weight = Number(entry?.weight);
+    if (!name || !Number.isFinite(weight) || weight <= 0) continue;
+    weights.set(name, (weights.get(name) ?? 0) + weight);
+  }
+  const total = [...weights.values()].reduce((sum, value) => sum + value, 0);
+  if (total > 1) {
+    for (const [name, weight] of weights) weights.set(name, weight / total);
+  }
+  return weights;
+}
```
kenlasko, line 154: The overlap is exact string equality on names (trimmed and lowercased), which is fine while both sides come from the same source — but this PR introduces a second source. `fetchEtfAssetPositions` writes English `"Stocks"` / `"Bonds"` / `"Cash"`, while the manual allocation editor lets a user type any class name they like (and the help text invites exactly that).

So a hand-entered `"Bond"`, `"Fixed income"`, or a non-English name scores **0** overlap against a provider-filled fund, and `buildPositionMath` then reports that as a confident low `compliancePercent` rather than falling back to the honest instrument comparison — because `hasWeightings(holding[dimension])` is true, just with names that never intersect.

Two options, either fine: canonicalise a small synonym set for the classes the provider emits, or treat "both sides described but zero names in common" as a signal to fall back to `byInstrument` with the existing caveat. The second is more conservative and reuses machinery already here.

---

kenlasko (changes requested):
## Review: GEM strategy report

Reviewed the full diff (189 files, +23k) against the base commit, read the `strategies` module, migrations 124-127, the backup wiring and the frontend components, and checked CI — all 27 checks green.

The engineering standard here is high. Pure rule logic is separated into testable utils (`gem-momentum.util`, `gem-position.util`, `gem-composition.util`) with no clock and no database; every DB access goes through `withScopedDb`; SQL is parameterized throughout; `updateConfig` verifies ownership of every supplied `accountId` and `securityId` before writing; path params take `ParseUUIDPipe`; the RLS policies ship in-migration with their own `ENABLE`, which is the right call for files numbered after 123. The `rls-setup.ts` change is a genuinely good catch — a post-M3 policy migration would otherwise have been read as an enable migration and dropped from the default apply set, leaving its table unpolicied while the suite stayed green. Backup export, restore ordering, delete, integrity refs and support-backup classification are all covered, and the `momentum` JSONB allowlist is the right instinct.

One blocking issue, three that should land in this PR, and a handful worth considering. Details are inline; summarised here.

### Blocking

1. **The first save from an unconfigured strategy returns 400.** `GemSettingsForm` sends the empty report's sentinel id `"gem"` as `?strategyId=`, which the query DTO validates with `@IsUUID()`. Every brand-new user's first Save fails. No test catches it — the controller spec never exercises `strategyId`, unit specs bypass the `ValidationPipe`, there is no E2E pass over the page, and `GemSettingsForm.test.tsx:291` asserts the broken call. (`GemSettingsForm.tsx:329`)

### Should fix

2. **Raw `<input type="number">` for three fields** where `frontend/CLAUDE.md` mandates `CurrencyInput` (money) and `NumericInput` (rates, counts). CI is green because that rule has no guard test yet — worth adding to `ui-conventions.test.ts` in the same pass. (`GemSettingsForm.tsx:383-411`)
3. **`securities:form.assetAllocation.help` is now contradicted by this PR** — it still tells the user providers don't supply asset-class exposure, which `fillAssetPositions` makes false for ETFs.
4. **Missing `updated_at` triggers** on `gem_strategies` and `gem_strategy_assets`; every other table in `schema.sql` with that column has one, and without it these two sit outside the `withPreserveTimestamps` mechanism.

### Worth considering

5. `fetchEtfAssetPositions` and `fetchEtfSectorWeightings` hit the identical Yahoo URL with identical modules and parse the same object — two round trips per ETF for one payload.
6. Asset-class overlap is exact string matching, and this PR introduces a second source of names (English, from the provider) alongside user-typed ones, which can produce a confident-looking wrong compliance figure.
7. Writes on a GET: `materialize` sits outside `CsrfGuard`/`DemoModeGuard` (both method-based), and a permanently unevaluable period reloads its full price window on every page view.
8. Every mutation computes two full reports — the returned one is discarded and refetched; `markExecuted` also omits `strategyId` and so builds the wrong scenario's report before throwing it away.
9. `InfoTooltip` now has a tab stop that announces nothing (focusable `<span>` with `aria-label` on an implicit `role=generic`), and no Escape dismissal.
10. `loadComposition` is the only query in the module not scoped by `userId` — safe via the call chain today, one line to make it not depend on that.

### Nits

`SecurityForm.tsx` leaves a double blank line where the old `useMemo` was; `StrategiesModule` exports `GemStrategyService` with no consumer; the `runBacktest` drawdown walk rescans the whole series per period where `priceAsOf`'s binary search is right there; and the `currencyNotConfigured` fix in `SecurityForm` is unrelated drive-by work against the PR's own single-concern checkbox — good change, just worth naming.

None of this undermines the design. The strategy rules, the RISK_FREE/SAFE split, the RLS wiring and the composition-based compliance model are all sound, and the file granularity and comment quality are better than most of what lands in this repo. Fix #1 before merge; #2-#4 are small and belong here rather than in a follow-up.

---

WMP:
I do /code-review few times, maybe it's needed to make it in new context. I will fix this and make cross review with codex

kenlasko:
Maybe I need to read some directions, but the UI leads me to a Settings page, where I accepted the suggestions and loaded 2 accounts. However, when I click `Save configuration`, I get the error: `strategyId must be a UUID`.  No corresponding error on backend.

image

WMP:
This is related with "Blocking — the first save from an unconfigured strategy returns 400."

WMP:
##  codex: Follow-up review: signal and trade correctness

I reviewed the GEM implementation at `8fd4c4f`, focusing on whether the materialized signal, portfolio-compliance calculation, and recommended operation remain correct under real portfolio data. The code is well structured and extensively tested, but I found several correctness issues that can produce a wrong signal or an operation that does not lead to the stated target allocation.

This is in addition to the already reported first-save blocker (`strategyId=gem`).

### Blocking

1. **Changing the strategy configuration reuses signals calculated under the previous configuration.**

   `GemSignalService.materialize()` treats `(strategy_id, evaluated_on)` as permanently complete and skips every known period. However, `updateConfig()` edits signal-driving inputs in place: `lookbackMonths`, `cadence`, role-to-security assignments, and the `SAFE` / `RISK_FREE` benchmark.

   After changing 12-month momentum to 6 months, replacing an ETF, or changing cadence, the report can continue using a row calculated with the old rules and old instruments. The presentation makes this more misleading: `toSignalView()` and `toHistoryEntry()` resolve the stored `targetRole` through the **current** role mapping, rather than through the stored `targetSecurityId`, so an old signal can appear to have selected the newly assigned instrument.

   Please either version/fingerprint the signal-driving configuration and materialize per revision, or explicitly invalidate and recompute affected signals when those fields change. Historical rows should also render the instrument that was actually stored on the signal.

2. **Momentum, performance charts, and the backtest use unadjusted closes.**

   `GemPriceService.loadSeries()` reads `close_price`, although the historical-price pipeline stores `adjusted_close`. Every historical calculation then consumes the same unadjusted `PricePoint.close`.

   This excludes distributions from momentum and can turn a stock split into a large apparent loss, potentially changing the GEM signal and corrupting CAGR and max drawdown. Historical series should use something equivalent to:

   ```sql
   COALESCE(adjusted_close, close_price)
   ```

   The latest market price used to value current holdings should remain the ordinary close. Please add regression tests covering a split and a distribution.

3. **A missing equity-market history still produces an actionable winner.**

   `rankEquities()` drops roles whose momentum is `null`, while `evaluate()` only requires the US leg and benchmark to be known. If US and ex-US have history but the assigned EM instrument does not, the report selects the strongest of the two known markets and presents a concrete switch, even though the missing market could have been the actual winner.

   An `INCOMPLETE_HISTORY` warning is not sufficient when the same response contains an actionable recommendation. If any assigned equity candidate required by the configured relative-momentum comparison is unknown, the current period should remain unevaluated/provisional and no operation should be recommended.

4. **The operation for a partially overlapping fund is not executable as calculated.**

   Composition overlap is useful for explaining exposure, but `buildPositionMath()` also treats the overlapping fraction as something that can be retained while selling only the non-overlapping fraction. For example, a 10,000 world ETF that is 20% EM against a 100% EM target produces an 8,000 transfer and realizes only 80% of its gain.

   Selling 80% of the ETF also sells 80% of its EM exposure. After buying 8,000 of the target, the portfolio is only 84% EM, not 100%. Because the target weight is 100%, any non-target instrument with overlap below 100% must be sold completely; the overlap can remain a diagnostic compliance figure, but it cannot be used directly as the fraction of ETF units to sell. Tax and transfer estimates must be based on the executable trade.

5. **Cash in the selected brokerage accounts is omitted from compliance and the proposed purchase.**

   `GemPositionService.loadHoldings()` reads securities from `holdings` only. The existing portfolio model separately carries linked investment-account cash balances, but GEM does not include them.

   An account containing 5,000 of the target ETF and 5,000 cash is therefore reported as 100% compliant, with no purchase required. If GEM requires the selected accounts to be 100% in one asset, available brokerage cash must count as off-target value, reduce compliance, increase the buy amount, and be included in the strategy notional.

### Should fix

6. **One fresh instrument hides stale or missing prices for the others.**

   `latestPriceDate()` returns `MAX(price_date)` across all mapped securities, and that single date drives both “prices as of” and the stale-price warning. A fresh US quote can therefore hide an ex-US quote that is weeks old or an instrument that has no price at all.

   Staleness should be evaluated per required role/security. The aggregate date should represent the oldest required input, or the response should identify the stale instruments explicitly.

### Suggested regression coverage

- Change lookback, cadence, benchmark, and a role assignment after signals exist; verify that the current signal cannot come from the old configuration and that history shows the originally selected security.
- Adjusted-close series containing a split and a distribution.
- One assigned equity role with insufficient history must not produce an operation.
- A mixed ETF with 20% overlap against a 100% target must require selling the whole mixed ETF.
- Target holding plus brokerage cash must not report 100% compliance.
- Fresh price for one role plus stale/missing prices for another must raise the appropriate warning.

The overall module design, RLS wiring, backup integration, and separation of pure arithmetic from data access are strong. The issues above are blockers because this page is explicitly presenting a current allocation signal and concrete monetary operation; those outputs need to remain correct after configuration changes and under common ETF/account conditions.

WMP:
Follow-up review after the correctness fixes

I reviewed the current head, "8d20d7b".

This is a substantial improvement. The first-save sentinel, shared numeric inputs, duplicate Yahoo request, adjusted-close handling, incomplete-market behavior, partial-overlap trade calculation, brokerage cash, per-security freshness checks, stored historical instrument references, and tooltip accessibility have all been addressed.

In particular, the adjusted-close, missing-market, full-sale, cash, and price-freshness fixes look correct. The configuration fingerprint also handles lookback and role-assignment changes where the evaluation dates remain the same.

I found two remaining correctness blockers.

Blocking

1. Changing cadence mixes signals from the old and new calendars

"materialize()" loads the newest 24 stored rows regardless of their fingerprint or cadence, then generates the 24 dates belonging to the current cadence.

A stale row is recomputed only when its "evaluatedOn" date also appears in the new calendar. This works for a lookback or instrument change because the period dates remain unchanged, but it does not work for a cadence change.

For example, after changing from monthly to quarterly:

- quarter-end monthly rows such as March 31 and June 30 are recomputed;
- the other monthly rows, such as February 28, April 30 and May 31, remain stored with the old fingerprint;
- the final query still returns the newest 24 rows without filtering by fingerprint or the current cadence;
- "getReport()" passes that mixed set directly to both the history table and the backtest.

The resulting quarterly scenario therefore replays old monthly decisions between the new quarterly decisions. The newest 24 monthly rows can also crowd older valid quarterly periods out of the returned history.

There is another edge case once the table contains more than 24 rows: an older quarterly date may already exist in the database but not be present in the initial "take: 24" result. The insert then reaches "orIgnore()", leaves the old-fingerprint row untouched, and never refreshes it.

Please reconcile signals against the exact set of dates in the current calendar and return only signals belonging to the current configuration revision. For example:

- load existing rows for the exact current-period dates rather than only the newest 24 rows;
- refresh every matching date whose fingerprint differs;
- exclude rows outside the current calendar from the current history and backtest.

If histories from previous configurations are meant to be retained, they should be associated with an explicit configuration revision and displayed separately rather than mixed into the current scenario.

A regression test should populate a monthly history, switch the strategy to quarterly, and verify that every returned and backtested period is a quarterly date carrying the current fingerprint.

2. "executed" belongs to the signal, but the operation is dynamic

The execution flag is stored on "gem_strategy_signals", while the actual operation is calculated from the currently selected accounts, holdings and cash balances.

This makes the flag stale as soon as the portfolio changes without the signal changing.

A reproducible example:

1. The user marks the current operation as executed.
2. The user adds another brokerage account to the scenario, deposits new cash, or acquires another off-target holding during the same signal period.
3. The report correctly calculates a new required operation.
4. The action still receives "executed: true" from the unchanged signal.
5. The UI displays “already executed” and hides both action buttons, even though the operation currently shown has never been executed.

Resetting the flag only when the target instrument changes is therefore not sufficient. Execution should be associated with the concrete operation or with an operation fingerprint containing at least the target and account set, and preferably the relevant position snapshot. The report should treat an execution as current only while that fingerprint still matches.

Resetting the flag when the account set changes would cover the deterministic settings case, but a stored operation snapshot is safer because holdings and cash can change independently of the strategy configuration.

Should fix

3. A stale signal can still become "previousRole" for a current signal

"byEvaluatedOn" is seeded with every stored signal, including rows whose fingerprint does not match the current configuration.

Consider an instrument change where an intermediate period cannot be recomputed because the new instrument does not yet have enough history, but a later period can be evaluated. The later period selects the stale intermediate row as "previous" and stores its old target as "previousRole".

The history can then claim that the strategy switched out of an allocation that was never produced under the current configuration.

Only signals matching the current fingerprint, together with signals successfully refreshed during the current pass, should participate in the "previousRole" chain.

4. Cash is labelled as “Unassigned” in the main holdings and action views

The backend now correctly marks the synthetic cash position with "isCash", but it deliberately has no symbol or security name. "assetFullLabel()" consequently renders it as “Unassigned”.

For an account containing only cash, the recommendation can therefore show:

«From: Unassigned»

The main portfolio holdings list has the same issue, although the detailed working table already special-cases cash correctly.

Please use the existing "isCash" field to render a localized “Cash” label, omit the security link, and avoid showing an unknown share quantity for that row.

Suggested regression coverage

- Populate 24 monthly signals, change the cadence to quarterly, and verify that history and backtest contain only quarterly dates with the current fingerprint.
- Keep a stale, unevaluable intermediate signal and evaluate a later period; verify that the stale row cannot supply "previousRole".
- Mark an operation as executed, then add an account or additional cash; verify that the newly required operation is not reported as already executed.
- Build a report for an account holding only cash and verify that the action names it as cash rather than an unassigned security.

Verdict

Request changes.

The original correctness findings have been addressed well, but cadence transitions still produce a mixed historical strategy, and the signal-level execution flag can suppress a newly required operation. Those two issues should be resolved before merging.

WMP:
Follow-up review at "7c03ed1"

The latest fixes are a clear improvement.

The cadence-specific calendar filtering now prevents ordinary monthly rows from appearing in a quarterly history. Seeding "previousRole" only from signals produced by the current configuration is also the correct direction. The cash labelling and sell-side changes are sensible, and the UI no longer hides a live operation merely because the signal was previously marked as executed.

I found three remaining correctness blockers.

Blocking

1. History and backtest can still mix signals from different configuration fingerprints

"materialize()" now limits the database reads to dates belonging to the current cadence, but it still returns every row on those dates regardless of "configFingerprint".

That matters whenever an old row cannot be recomputed under the new configuration.

For example:

1. A monthly strategy has an existing history.
2. The EM instrument is replaced with a recently listed ETF.
3. A June RISK-ON period cannot be recomputed because the new EM instrument has insufficient history, so the old June row remains.
4. A later July RISK-OFF period can be recomputed because that decision only needs the US and benchmark legs.
5. The returned array now contains fresh rows under the new fingerprint and the stale June row under the old fingerprint.

"getReport()" passes this mixed array directly to both the history view and the backtest.

This also reintroduces a history-display bug despite the new "previousRole" seeding:

- the July row correctly calculates "previousRole" from the last valid new-configuration signal, perhaps May;
- "toHistoryEntry()" nevertheless resolves the previous instrument from "signals[index + 1]";
- in the sorted array, that entry is the stale June row;
- the history can therefore display the June instrument as the asset July switched out of, even though "previousRole" was derived from May.

The returned history is not a coherent “actual history” either, because periods that can be recalculated are overwritten under the new configuration while periods that cannot be recalculated retain the old one. It is a hybrid of counterfactual and historical signals.

Please separate the two concepts:

- use only rows matching the current fingerprint for the current history, predecessor chain and backtest;
- if legacy configuration history should remain visible, expose it separately and identify its configuration revision;
- do not use positional adjacency across different fingerprints to resolve the previous instrument.

A regression test should create:

- a valid current-fingerprint May signal,
- an unevaluable stale June signal,
- a valid current-fingerprint July signal,

and verify that July resolves its "from" instrument from May and that June is not included in the current-configuration backtest.

2. One unpriced holding can make a non-compliant portfolio appear 100% compliant

"buildPositionMath()" filters out null market values when calculating "totalMarketValue", while treating the same missing values as zero in "targetValue".

It then trusts the resulting percentage when deciding "changeRequired".

Consider:

- 10,000 in the correctly targeted ETF;
- one additional off-target holding with no available price.

The calculation becomes:

- known total market value: 10,000;
- target value: 10,000;
- compliance: 100%;
- "changeRequired": false.

The report therefore says the portfolio is fully compliant even though it contains an off-target position that must be sold.

The same partial-known-data issue affects the operation amount. When several off-target positions exist and one has no price, "transferValue" sums only the priced positions and presents that subtotal as the amount to move.

Please treat missing valuation data as unknown rather than silently excluding it:

- if any relevant holding is unpriced, the total and compliance percentage should be unknown or explicitly partial;
- "changeRequired" should still be true whenever an off-target holding exists, even when its value is unknown;
- "transferValue" should be null if any position included in the transfer cannot be valued;
- the response should preferably contain a warning identifying the unpriced holdings.

Regression coverage should include:

- a priced target holding plus an unpriced off-target holding;
- two off-target holdings where only one can be valued.

3. A missing cost basis on one sale produces a partial gain and tax estimate presented as complete

The realized-result calculation filters out holdings whose market value or cost basis is unknown and sums the remaining ones.

For example:

- Fund A has a known 4,000 gain;
- Fund B is also being sold but its cost basis is unknown.

The report returns a realized gain of 4,000 and calculates tax on that amount. The actual gain for the whole operation is unknown and may be substantially higher or lower.

This is especially dangerous because the value is displayed as the estimated realized result and tax for the complete switch, not as a partial estimate covering only some sales.

The result should be all-or-nothing:

- if every security being sold has a known market value and cost basis, sum the gains and losses;
- if any sold security lacks either value, return "realizedGainLoss: null" and consequently "estimatedTax: null";
- cash should not affect this check because it is spent rather than sold.

Please add a regression test with two sold holdings where only one has a known cost basis.

Verdict

Request changes.

The latest commits correctly resolve the previously reported cadence, predecessor, cash-rendering and execution-visibility issues. The remaining problems are blockers because they can still produce a false compliance result, an understated operation amount, an understated tax estimate, and a backtest assembled from incompatible strategy configurations.

WMP:
Follow-up review at "50a15a4"

The latest changes correctly address the configuration-history issue.

"materialize()" now returns only signals carrying the current configuration fingerprint, so stale rows can no longer leak into the history, predecessor chain, or backtest. The compliance percentage also correctly becomes unknown when any holding cannot be valued.

The cadence-aware backfill range change looks reasonable as well.

There are still three correctness blockers.

Blocking

1. Missing prices still produce partial totals and a partial transfer amount presented as complete

The new "unpricedCount" check correctly prevents the portfolio from being reported as 100% compliant. However, the other monetary results still silently omit unpriced holdings.

"totalMarketValue" is calculated from "knownValues" only:

const knownValues = valued
  .map((holding) => holding.marketValue)
  .filter((value): value is number => value !== null);

const totalMarketValue =
  knownValues.length > 0 ? sumMoney(knownValues) : null;

Similarly, "transferValue" filters out unknown values and sums whatever remains:

const offTargetValues = offTarget
  .map((holding) => holding.marketValue)
  .filter((value): value is number => value !== null);

const transferValue =
  offTargetValues.length > 0 ? sumMoney(offTargetValues) : null;

For example:

- 10,000 in the target;
- 4,000 in a priced off-target fund;
- another off-target holding with no price.

The report returns:

- compliance: unknown;
- total market value: 14,000;
- estimated value to transfer: 4,000.

Both monetary figures are only known subtotals, but the UI labels them as the total market value and the estimated value of the complete operation. The partial total is also passed to the backtest as its notional, which makes the commission drag incorrect.

Please make these aggregates all-or-nothing:

- "totalMarketValue" should be "null" when any included position is unpriced, unless the API explicitly exposes a separate known subtotal;
- "transferValue" should be "null" when any off-target position cannot be valued;
- a warning identifying unpriced holdings would make the reason actionable.

Please add a regression test with one priced and one unpriced off-target holding.

2. Realized gain and tax are still calculated from only the positions with known cost basis

The previous tax blocker remains.

The implementation filters out positions with missing data and sums the rest:

const realizable = offTarget.filter(
  (holding) => holding.marketValue !== null && holding.costBasis !== null,
);

Therefore, when two securities are being sold and only one has a known cost basis, the report presents the gain and tax for that one security as the estimate for the whole switch.

There is an even more misleading cash case:

- an unpriced off-target security has an unknown result;
- cash is also off target and has equal market value and cost basis;
- cash enters "realizable";
- the result becomes zero;
- the report can show an estimated tax of zero even though the security being sold has an unknown gain.

The calculation should be based on "sold", not on all "offTarget" positions, and it should require complete information for every security being sold.

Suggested behavior:

- no securities to sell, cash-funded purchase only → realized result "0";
- every sold security has a market value and cost basis → sum the results;
- any sold security is missing either value → realized result and estimated tax are "null".

Please add tests covering both a mixed known/unknown cost basis and cash plus an unpriced security.

3. The backtest silently converts missing or stale prices into a 0% return

The backtest currently reports a concrete CAGR and drawdown even when it does not know what happened during part of the simulated period.

There are two paths to this:

1. When a target security has no series, "periodGrowth" is "null", the loop applies no return, and the full period still remains in the annualization timeline.
2. When a series exists but stops before the period end, "priceAsOf(series, endsOn)" carries the last available price forward indefinitely. The entry and exit can therefore be the same stale observation, producing an apparent 0% return.

The new regression test explicitly expects an unpriced period to be held flat. That is not a neutral treatment of missing data; it is an assumption that the asset returned exactly zero. The UI has no incomplete-data flag and presents the resulting CAGR and max drawdown as ordinary backtest results.

A missing six-month return can materially change CAGR, drawdown and hit rate. It should not silently become zero.

Please either:

- require valid price coverage for every period with a target security and return no backtest when coverage is incomplete; or
- add an explicit incomplete state, omit metrics affected by missing periods, and clearly disclose which periods were excluded.

Price coverage should also enforce a reasonable maximum age around the period boundaries rather than carrying an arbitrarily old quote forward.

Should fix

4. A request that loses every concurrent insert returns the stale pre-insert snapshot

"orIgnore()" correctly prevents concurrent materialization from aborting the transaction. However, when another request inserts all missing periods first, every "generatedMaps[0]" is empty, "written.length" remains zero, and this request returns "current(stored)".

"stored" was read before the competing request inserted the rows, so a first report load can temporarily return no signals and display a first-run or calculation-failed state even though the signals now exist in the database.

The current concurrency test explicitly expects an empty result, which locks in that stale response.

After any attempted insert that lost a conflict, the service should re-read the current calendar and fingerprint before returning. Always performing the final read after entering the materialization loop would be the simplest option.

Verdict

Request changes.

The configuration-history fix is correct, but the report can still present known subtotals as complete portfolio, transfer, gain and tax estimates. The backtest also produces definitive metrics by assuming missing returns were exactly zero. Those outputs should remain unknown or explicitly incomplete until all data needed for the calculation is available.

kenlasko:
While validating issue #1017, two internal inconsistencies in this PR's `docs/gem-strategy.md` were confirmed (finding 9 of that issue). They need a fix on this branch, since the file only exists here:

1. **The roles list omits `RISK_FREE`.** The "Rules" section says `Roles: US_EQUITY, EX_US_EQUITY, EM_EQUITY, SAFE`, but the PR description and the absolute-momentum rule both describe a separate optional `RISK_FREE` role that falls back to `SAFE`. The list should include `RISK_FREE` and state its fallback behaviour.

2. **The API section says `backtest` is always `null`, but later sections describe an implemented backtest.** "`backtest` is always `null` for now..." contradicts the sections that specify the backtest reads `COALESCE(adjusted_close, close_price)`, simulates the most recent unbroken stretch of priced periods, reports truncated runs gross, and computes `hitRatePercent`/`coveragePercent`. Whichever reflects the current code should stand, and the other section should be removed or rewritten.

The rest of #1017 (deleting the stale `.claude/skills/*` files, updating `CONTRIBUTING.md`, and adding `docs/financial-calculation-contract.md` + `docs/time-series-contract.md`) is addressed on `claude/issue-1017-review-5acpw0`. Once that lands, it may be worth cross-referencing the new contracts from this document.

---

WMP:
I think this is ready to merge, build to test:

```
ghcr.io/wmp/monize-backend:claude-gem-strategy-report-page-x5ej8j-d51661d
ghcr.io/wmp/monize-frontend:claude-gem-strategy-report-page-x5ej8j-d51661d
```

I need to add tutor for this and to resolve this one CI pipeline problem.

kenlasko:
## Full re-review at `83921a9`

Fresh whole-branch review of the current head (215 files, +39,593/−196), independent of the earlier rounds. Verified locally: backend suites for `strategies`/`backup`/`securities` (48 suites, 1,767 tests) and the frontend GEM/i18n/guard suites (130 files, 3,500 tests) all pass under `TZ=UTC`; `tsc --noEmit` and `i18n:check` are clean, so the CI failure at `d51661d` (missing `isLoading` prop in `GemPerformanceChart.test.tsx`) is genuinely fixed by the head commit. The prior rounds' fixes held up under re-verification: the fingerprint/cadence model, adjusted-close basis, all-or-nothing money aggregates, lost-race abandon path, backup fingerprint re-hash on restore, and the frontend request-key discipline are all in place and pinned by tests. The findings below are new.

### Blocking

1. **Migration number collision with `main`.** `main` now contains `124_scheduled_transaction_currency.sql` (merged via #1021), and this branch adds `124_gem_strategies.sql` — `ls database/migrations | awk -F_ '{print $1}' | sort | uniq -d` prints `124` at the merge result. `database/CLAUDE.md` requires unique prefixes, and this is only fixable before merge: once a deployment records the name in `schema_migrations`, renaming breaks re-application. The whole GEM chain (124–131) needs renumbering, together with the number references in `schema.sql`'s policy-count comment, `docs/gem-strategy.md`, and the cross-references inside migrations 125/126/129/130.

### Should fix — backend correctness

2. **A period can be materialized permanently from a boundary close that has not been ingested yet** (`gem-signal.service.ts:541-577` + `gem-momentum.util.ts` `spanCloses`). The 14-day boundary lag correctly tolerates weekends/holidays, but nothing distinguishes "market closed on the boundary" from "close not fetched yet", and a stored period is skipped forever under the same fingerprint/version. Open the report at 9am on Aug 1 before the quote job has stored Jul 31's close, and the Jul 31 signal is computed off Jul 30, persisted, and never corrected — even when the true month-end close flips the decision. A cheap closure in the module's own vocabulary: treat the exit boundary as settled only once the series contains an observation dated ≥ `evaluatedOn`, deferring exactly the way UNELAPSED periods already defer. If "decide with what you had" is deliberate policy, it should be written into `docs/gem-strategy.md`, because it silently diverges from the strategy's month-end-close rule for anyone who loads the report early.

3. **Legacy `price = 0` BUY/REINVEST rows are replayed as genuinely free acquisitions** (`portfolio-calculation.service.ts:824-846`). Before this PR, `create()` stored `price ?? 0` and the form allowed a blank price, so real databases contain zero-price BUYs that mean "unknown". The lot replay treats an explicit 0 as a known zero-cost lot (`basisKnown: true`), producing a confidently understated basis and overstated gain/tax — the exact class this PR closes elsewhere. Since the new `assertAcquisitionPriced` guarantees no legitimate zero-price acquisition can exist going forward ("a zero-cost purchase is not a concept this application has"), stored zeros on BUY/REINVEST can safely be folded into `unpriced_acquisition`.

4. **Those same legacy rows are now uneditable through the real client** (`investment-transactions.service.ts:2984-2990`). The guard's carve-out ("a row that was already unpriced stays editable in every other respect") keys on `updateDto.action !== undefined || updateDto.price !== undefined` — but `InvestmentTransactionForm` always sends `action` (and loads a legacy blank price as `0` and sends it back), so every edit of a legacy zero-price BUY throws 400, including a description-only change. That's exactly the stranding the comment says it avoids. Compare against the incoming *change* rather than the field's presence, and add the `> 0` rule to the form's Zod schema so it surfaces before submit.

### Should fix — frontend

5. **`InfoTooltip`'s Escape handler swallows Escape app-wide in modals** (`InfoTooltip.tsx:76-84`). `dismissOnEscape` calls `stopPropagation()` unconditionally, even when no popover is showing; `Modal` closes on a document-level keydown listener the stopped event never reaches. With focus on any tooltip trigger inside any modal, the first Escape dismisses the tooltip (correct per WCAG 1.4.13) but every subsequent Escape is also eaten, so the modal can't be closed by keyboard. Only stop propagation when the popover is actually visible.

6. **"Discard" can disarm the unsaved-changes guard without discarding anything** (`GemStrategyReport.tsx:508, 674-680`). On the Settings tab with unsaved edits, clicking the header's "Edit settings" runs `guarded()` with a no-op action; choosing Discard sets `settingsDirty = false` but the form is never reset or unmounted, so its edits remain and react-hook-form's `isDirty` stays `true` — meaning `useFormDirtyNotify` never refires. The next tab/scenario switch then unmounts a genuinely dirty form with no prompt: the silent data loss the guard exists to prevent.

7. **`GemStrategyTabs` hand-rolls a tablist beside the shared `ui/Tabs.tsx` and reintroduces the defect it fixed.** It sets `aria-controls` unconditionally while only the active panel is rendered, so four of five tabs reference a nonexistent element — `Tabs.tsx:124-128` sets it only for the selected tab, with a comment saying exactly why. This is also the project's "one tab bar" pattern rule. (`GemSignalHistoryTable.tsx:235, 330` has the same collapsed-row `aria-controls` issue.)

8. **The report body carries no stale/loading marking during a scenario or range switch.** While `isStaleSelection` is true, the previous scenario's signal, portfolio, transfer, allocation and history stay on screen with no visual stale marker and no `aria-busy` (only the chart and settings tab swap to skeletons). Mutations are correctly disabled, but `frontend/CLAUDE.md` requires the pixels *and* AT to say the data is stale — a user switching A→B currently reads A's numbers under B's selection with nothing saying so.

### Should fix — docs

9. **`docs/gem-strategy.md` describes three superseded designs.** The earlier flags (Roles list, "backtest is always null") are fixed, but: line 68 still claims the unique key is `(strategy_id, evaluated_on)` (migration 131 widened it to include `algorithm_version` — line 88 says so); line 74 says the fingerprint is "led by `GEM_SIGNAL_ALGORITHM_VERSION`" (the code deliberately excludes it, as line 87 and the code comment both state); lines 123-129 describe a lost-race *replace-in-place* that the code deliberately abandoned (`gem-signal.service.ts:696-707` breaks instead, with a comment explaining why replacing would be wrong). Each is an invitation to "fix" correct code. The API table also omits the shipped `POST /strategies/gem` and `DELETE /strategies/gem/:id` routes and the `strategyId` query parameter.

### Nits

- `gem-position.util.ts:471-476` — `COMPLIANCE_TOLERANCE_PERCENT` is dead: it's only evaluated when `offTarget` is empty, in which case `exactTargetPercent` is exactly 100 or null. Delete it or apply it where it was meant to bite.
- `gem-momentum.util.ts:247` — `spanCloses` guards the entry close for positivity but not the exit close; a typo'd 0 close (the DTO allows `@Min(0)`) becomes a "known" −100% momentum or a zeroed equity curve.
- `gem-backfill.service.ts:47-51` — `monthsAgo` uses `setUTCMonth` from day 31, which overflows into the next month and can shift `needFrom` past a close that would satisfy coverage; the correctly-clamping `addMonthsUtc` is already imported. Worst case is a permanently re-retried provider fetch, not a wrong number.
- `portfolio-calculation.service.ts:990-999` — `settlementCurrencies` applies the linked-cash currency redirect to every account, but the settlement resolution it claims to mirror skips the redirect for named funding accounts; the comment says "the two must be changed together" and they already differ.
- `portfolio-calculation.service.ts:660-668` — the method contract says every caller must compare lot quantity against the holding, but `getHoldings` (via `knownCostBasesIn`) doesn't; pre-existing behavior, but per the repo's own rule the claim and the call site should agree (or the claim should become a scan).
- `gem-strategy.service.ts:880-885` — `rulesSourceUrl` skips the `normalizeWebsite(...) ?? null` pattern used by payees: `""` is stored as `""` and schemeless URLs stay schemeless (frontend defends itself; API callers see the difference).
- `gem-strategy.controller.spec.ts` — `createStrategy`/`removeStrategy` have no controller-level test (including the DELETE route's `ParseUUIDPipe` wiring).
- `GemInstrumentSelect.tsx` — a hand-rolled `role="listbox"` with no arrow-key/Home/End/typeahead support, options reachable only by Tab; `Combobox`/`EntitySwitcher` are the house patterns.
- `GemScenarioSwitcher.tsx:176-194` — "Create" is enabled with an empty name (server round-trip to a generic toast) and Enter in the name field doesn't submit.
- `gem-strategy.contract.test.ts:30-43` — the parity guard omits `GemStrategyMeta`, `GemStrategyReport`, the step types and `GemStrategyRef`; they agree field-by-field today, but drift in the settings-form fields would be invisible.
- `FavouriteAccounts.tsx` — moving the statement line out of the navigation button (correct hydration fix) leaves it a dead click zone inside a card whose hover still advertises the whole card as clickable.

### Test coverage gaps

- **No Playwright coverage at all** for a feature this user-facing — settings save, scenario switch, mark-as-executed would each pay for themselves in `e2e/tests/`.
- The adjusted-close `COALESCE` query and the ON CONFLICT insert — the two most safety-critical DB interactions in the layer — are verified by SQL string matching only; `gem-price.service.spec.ts` itself concedes the semantics belong in an integration test. A small GEM case in `backend/test/integration` would close it.
- Also noting: GEM is wired into neither the AI assistant tool executor nor the MCP tools. That's consistent with the shared-tools rule (it forbids shipping to *one* surface only), but worth a deliberate decision on record.

### Verdict

This is the strongest-engineered feature branch I've reviewed in this repo — the missing-data discipline, the materialization protocol (advisory lock, fingerprint/version predicates, abandon-on-foreign-winner), the request-keyed async handling on the frontend, and the backup fingerprint re-hash are all careful, documented, and pinned by adversarial tests. Nothing above undermines the design. But the migration renumbering (1) is a hard pre-merge gate, the legacy zero-price pair (3, 4) mis-states real users' gains and locks their rows, and the Escape regression (5) degrades every modal in the app — so: **request changes**, with all of it fixable small.

---

PR #1043, opened 2026-08-04
01: Define the API scope vocabulary once (v.1.15.0)

WMP:
## Linked discussion / issue

None. Addresses audit finding **P1-004**. See the note at the end.

## Summary

`read` and `write` were spelled out in four places: the
`personal_access_tokens.scopes` comment in `schema.sql`, the regex in
`CreatePatDto`, the `monize:`-prefixed list `OAuthProviderService` advertises, and
`docs/future-plans/vat-support.md`.

Two of them named a third scope, `reports`, that no issuance path has ever
accepted. A client that read the schema and asked for it got a validation error
for a scope the repository said existed, and the planned MCP tool guarded by it
would have been unreachable to every PAT client.

`backend/src/auth/scopes.ts` now holds `API_SCOPES`, and the DTO's pattern, its
message, its Swagger description and `MCP_RESOURCE_SCOPES` all derive from it. The
schema comment points at the constant instead of restating it; the VAT plan asks
for `read`.

**Tests.** `scopes.spec.ts` enumerates from `API_SCOPES` rather than restating the
list, so adding a scope there is what makes the checks cover it. Beyond the
per-scope issuance and guard cases, two scanning guards — because the defect was
documents disagreeing with code, and prose does not stay in step on its own:

- `schema.sql`'s column comment must enumerate exactly `API_SCOPES`;
- no scope named as a `requireScope` argument anywhere under `docs/` may be one
  that cannot be issued.

Both fail when the original wording is restored. A third check ties the frontend's
`SCOPE_OPTIONS` to the same list, since a scope the backend accepts but the UI
never offers is unreachable for anyone not hand-crafting requests.

Both scanning guards read one exact construct rather than judging prose. A first
attempt scanned for quoted scope-shaped tokens near the word "scope" and
false-fired on a sentence explaining that `reports` had been removed.

Also clears two further stale claims in the VAT plan: it proposed a `QueryRunner`
that ESLint rejects in `src/`, and migration numbers `086`/`087` from a time when
the directory ended at `085`.

**Observation, not a change:** the DTO's validation message is a plain
`class-validator` string, not wrapped in `tr()`. That is true of every DTO message
in the repository and is left as-is.

## Checklist

- [x] New behavior has tests, and the existing suite passes. — 631 tests in the touched areas

**On the missing Discussion:** the audit offered a choice — remove `reports`, or
build it out across DTO, OAuth registration, guards, MCP and the PAT UI. This PR
takes the first, since nothing in the tree needs it. If you intend `reports`, say
so and I will do the second instead.

## AI assistance disclosure

result. Verified locally: `auth` and `oauth` suites pass, `tsc`, `typecheck` and
ESLint clean. Both scanning guards were confirmed to fail when the original
`schema.sql` comment and VAT-plan wording were restored.

### DIFF
--- a/backend/src/auth/dto/create-pat.dto.ts
+++ b/backend/src/auth/dto/create-pat.dto.ts
@@ -9,6 +9,7 @@ import {
 import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
 import { IsFutureDate } from "../../common/validators/is-future-date.validator";
 import { SanitizeHtml } from "../../common/decorators/sanitize-html.decorator";
+import { apiScopesList, apiScopesPattern } from "../scopes";
 
 export class CreatePatDto {
   @ApiProperty({ description: "User-assigned label for the token" })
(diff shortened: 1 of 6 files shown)


WMP:
@kenlasko It’s ready for merging, unless you were planning to extend the API’s scope to include `reports` operations; currently, it only supports read and write operations.

PR #1044, opened 2026-08-04
01: Ban DataSource.transaction() and make the documented ban list checkable (v.1.15.0)

WMP:
## Linked discussion / issue

None. Addresses audit finding **P1-003** and **RFR-004** of the follow-up review.
See the note at the end.

## Summary

The `createQueryRunner` ESLint rule never covered `dataSource.transaction(fn)`,
and contributor guidance recommended it. It is not equivalent: it opens a
transaction that does not know about the ambient scoped manager, so under
`RLS_MODE=enforce` it carries no identity GUCs and fails closed, at `off` it loses
defence in depth, and nested inside a caller's `withScopedDb` it commits
independently — the outer rollback cannot take it back.

ESLint now rejects it in `src/` outside `scoped-db.ts`, specs and test helpers. **No
existing call site needed changing** — `scoped-db.ts` is the only caller, which is
why the ban is safe outright rather than behind a ratchet.

Wording alone is not the fix: `CONTRIBUTING.md` had already drifted once, and a
document cannot be type-checked. `backend/src/common/db/lint-bans.spec.ts` reads
the ban list out of `eslint.config.mjs` and requires every banned call to be named
in the root `CLAUDE.md` and `CONTRIBUTING.md`, and no instruction file to
recommend one. Both directions fail: restoring the original "MUST use a
`createQueryRunner()` transaction" sentence, and dropping the new ban from the
documentation.

The classifier is its own module with table-driven fixtures rather than only a
scan of the live documents, because scanning proves that today's wording passes
and nothing more. It detects prohibited call forms **and** type names that are
never ordinary prose (`QueryRunner`, `InjectRepository`), with prohibitive or
historical framing taking precedence over the recommending verb — which is what
lets `use` count as a recommender without firing on "Never … or use a bare
`dataSource.query(...)`".

Worth stating because it is easy to get backwards: requiring only the *call* form
was a deliberate fix for the opposite mistake ("MUST run in a single transaction"
is a noun these files use constantly), and it missed **four of the six** positive
fixtures, including the exact historical sentence. Both halves are needed.

Also corrects `docs/future-plans/mny-import.md`, which described the
`@InjectRepository`/`createQueryRunner` counting ratchet as a live CI gate. It
reached zero and became these bans; the script, its baseline and both npm entries
are gone, so a reader was pointed at a check that could not fail.

## Checklist

- [x] New behavior has tests, and the existing suite passes. — 78 tests in the touched areas, 20 of them new
- [ ] No shared/core areas were refactored without prior agreement. — **`backend/eslint.config.mjs`**, see below

**On the shared config:** this adds one `no-restricted-syntax` entry to
`backend/eslint.config.mjs`. Nothing in `src/` violates it today, so it cannot
break the build, but it is shared infrastructure and a new ban is a policy
decision. Happy to split it into its own PR, or to drop it and keep only the
documentation guard, if you would rather agree it first.

## AI assistance disclosure

result. Verified locally: `common/db` suites pass, ESLint over `src` and `test`
clean with the new rule active, and the rule was confirmed to reject a fixture
file calling `dataSource.transaction()`. The documentation guard was confirmed to
fail in both directions.


WMP:
I think that this will need to be tested with `RLS_MODE=enforce` env

PR #1046, opened 2026-08-04
01: Gate migration prefix uniqueness and ordering in CI (v.1.15.0)

WMP:
## Linked discussion / issue

None. This is the regression gate recommended by **RFR-002** of the follow-up
review. See the note at the end.

## Summary

`database/CLAUDE.md` states both rules in prose. Prose did not hold, and I am the
worked example: a branch of mine added `133_import_jobs_one_active_per_user.sql`
while `main` had meanwhile taken `133` for `133_joint_account_grants.sql` and
shipped the same fix as `135_import_jobs_single_active.sql`.

Migrations are applied and tracked **by full filename**. On a database that had
already recorded `135`, the newly introduced `133_*` would be seen as pending and
applied *after* it — an obsolete body running against a schema the newer migration
had already corrected, with the order between the two `133_*` files decided by
alphabetical tie-breaking.

`scripts/check-migration-prefixes.mjs` checks two things:

1. no numeric prefix is used twice, except the historical pairs already recorded
   in `database/CLAUDE.md` (`022`, `068`, `075`, `116`, `117`, `124`). That list
   **may only shrink** — the script fails if an entry in it is no longer
   duplicated, so it cannot quietly re-permit a future collision;
2. when a base ref resolves, every migration the branch adds has a prefix
   strictly above the highest prefix on that base.

Restoring the `133_*` collision trips both. A synthetic `090_test_low.sql` trips
both as well.

Check 2 needs git history, hence `fetch-depth: 0` on that job's checkout. When no
base ref resolves it prints that it skipped rather than reporting OK — a check
that cannot run must not look like a check that passed.

## Checklist

- [x] New behavior has tests, and the existing suite passes. — the script is the test; both rules confirmed to fire

**Merge order:** PR 5 adds another step to the same `env-doc-check` job.
Whichever lands second needs a two-line rebase in `.github/workflows/ci.yml`.

**On `fetch-depth: 0`:** this makes that one job's checkout a full clone. If you
would rather not, the check still runs — it just skips its second half and says
so.

## AI assistance disclosure

result. Verified locally: passes on this branch (125 migrations, base maximum
`135`), and fails on both a restored `133` collision and a synthetic
low-numbered file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,6 +314,10 @@ jobs:
       - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
         with:
           persist-credentials: false
+          # The migration-prefix check compares this branch's migrations against
+          # the base branch's; a shallow clone has no base ref and the check
+          # reports that it skipped rather than passing silently.
+          fetch-depth: 0
       - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
           node-version: 24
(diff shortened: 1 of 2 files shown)


kenlasko:
@WMP you have a conflict in ci.yml

PR #1047, opened 2026-08-04
01: Type-check the test tree; four e2e suites had not compiled since 27 July (v.1.15.0)

WMP:
## Linked discussion / issue

None. Found while reviewing the files the audit listed; belongs to its Phase 7
(tests/CI) scope. See the note at the end.

## Summary

`backend/tsconfig.json` pins `rootDir` to `./src` and excludes `test/` because it
drives the build — and CI's `npx tsc --noEmit` uses it. ESLint's glob does cover
the directory, but a namespace import called as a function is a *type* error, not
a lint error. And CI runs only `test:unit` and `test:integration` (the latter
filtered to `test/integration/`), so nothing executes `test/*.e2e-spec.ts` either.

With no gate of any kind, `main.ts` moving to `import cookieParser from
"cookie-parser"` left three files under `test/` on `import * as cookieParser` and
four suites stopped compiling. `test/payee-detail.e2e-spec.ts` was edited three
days later, still broken, without anyone noticing — and that is the suite
`backend/CLAUDE.md` cites as what caught the raw-select transformer class of bug,
"because a unit spec with mocked query builders cannot".

Imports fixed and `npm run typecheck` (`backend/tsconfig.test.json`, `noEmit`, no
`rootDir` restriction) added to the backend lint job. Restoring the namespace
import fails it.

With the compile error gone, `payee-detail` **passes all 9 of its tests** — that
coverage was simply absent. The other three fail on rot that had accumulated
behind it, each from a different change that moved on without these files:

| Suite | State | Cause |
|---|---|---|
| `test/payee-detail.e2e-spec.ts` | passes (9) | — |
| `test/payees.e2e-spec.ts` | fails | calls services directly; never converted for RLS |
| `test/auth.e2e-spec.ts` | fails | `AuthController` gained a `TokenService` its test module lacks |
| `test/transactions.e2e-spec.ts` | fails | `DelegateTransferMaskInterceptor` gained a `CrossOwnerAccessService` |

Those three are documented in `backend/CLAUDE.md` with their exact causes and are
deliberately **not** fixed here — repairing them is separate work, and `test:e2e`
must stay out of CI until it is done or the job goes red. What they must not remain
is present, cited, and dead. Happy to take that on as a follow-up if you want it.

## Checklist

- [x] This PR addresses a **single concern**. — close the compile gap and gate it
- [x] New behavior has tests, and the existing suite passes. — the gate is the test; it fails on the restored import. Restores 9 previously-dead tests

**Judgement call for you:** the three broken suites could instead be *deleted* in
this PR. I left them, documented, because `payee-detail`'s history suggests they
encode real knowledge — but they are dead weight until repaired, and deleting them
is a defensible alternative.

## AI assistance disclosure

result. Verified locally: `npm run typecheck` clean, fails when the namespace
import is restored, and `test/payee-detail.e2e-spec.ts` passes 9/9 against a real
PostgreSQL 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,13 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npx tsc --noEmit
+      # tsconfig.json excludes test/ (it drives the build and pins rootDir to
+      # src/), so `tsc --noEmit` above never type-checked the test tree. Four
+      # `test/*.e2e-spec.ts` suites sat uncompilable for over a week as a result
+      # -- ESLint's glob covers the directory but a namespace import called as a
+      # function is a type error, not a lint error, and CI runs neither
+      # `test:e2e` nor a typecheck that reaches them.
+      - run: npm run typecheck
       # Migration idempotency: every migration body must be a no-op when
       # re-applied to an up-to-date database, or a half-applied migration
       # crash-loops the backend at startup (task B1). Self-test first, then lint.
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -8,7 +8,8 @@ NestJS API server. All commands run from this directory.
 npm run start:dev          # Dev server with HMR
 npm run build              # Production build
 npm run lint               # ESLint --fix
-npm run test               # All tests (unit + E2E)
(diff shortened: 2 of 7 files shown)


WMP:
It is possible that this test will reveal further typing errors in the code.

PR #1048, opened 2026-08-04
01: Add the audit and remediation-review response documents (v.1.15.0)

WMP:
## Linked discussion / issue

None. These *are* the written record of the audit and its remediation. See the
note at the end.

## Summary

Two documents under `docs/audits/`, answering an external audit of this repository
and the follow-up review of the remediation.

Each answers a finding with the verdict, what changed, which test asserts it, and
what fails when the original defect is restored. They also record the parts a
commit log does not carry:

- where a recommendation was deliberately **not** followed, and why — no ADR for
  the scoped-manager rule; no `helm lint` / `docker compose config` in the new CI
  job; not aborting the import migration on duplicate `running` rows;
- where the audit had already been overtaken by `main` — two of its observations
  no longer applied, which shrank finding P1-003 to the enforcement gap underneath
  them;
- which hypotheses were checked and **rejected**, including two that looked like
  privilege escalation and were not, because `AccountDelegateGuard` is registered
  globally and rejects acting tokens on any route without `@AllowDelegate`.

The one finding still open is written up in full rather than left implicit:
`totalMarketValue` sums an unpriced position as zero, so a portfolio holding one
security without a quote reports a loss equal to that security's cost basis, with
no signal that anything is missing. It gets a reproduction table, the contract
clause it violates (`docs/financial-calculation-contract.md` section 1, which names
the exact field and the `knownMarketValueSubtotal` that does not exist), the reason
a partial fix would be worse than none, and a suggested test matrix — including the
two cases that are easy to get backwards, an all-unpriced account and an empty one,
where `null` and `0` mean different things.

Not fixed there or here: the repository's own rule is that a financial change of
that size starts from an approved spec, and making the totals nullable ripples into
the frontend, the reports and the AI/MCP adapters.

## Checklist

- [x] This PR addresses a **single concern**. — documentation only, no code
- [x] New behavior has tests, and the existing suite passes. — no behavior; every cited line number, path and count re-checked against the tree
- [x] All user-facing strings are translated for every locale. — no strings

**Merge order:** this must land **after PR 7**. PR 5 adds a rule requiring every
`npm run <script>` named under `docs/` to exist, and these documents record having
run `npm run typecheck`, which PR 7 adds. PR 5 + this without PR 7 fails that
check.

**On the missing Discussion, and on whether this belongs in the tree at all:**
these are correspondence with the auditors, not project documentation. If you would
rather they lived in a Discussion, close this PR — nothing else in the series
depends on it, and I will move the content instead. I am including them because the
open finding above deserves a durable home, and because the "deliberately not
followed" reasoning is the part most likely to be re-litigated later.

## AI assistance disclosure

Written with Claude Code (Claude Opus 5). The audits were produced by other AI
reviewers; these responses, the code they describe and this PR body are also
AI-produced. I have reviewed them and own the result. The documents are held to the
same rule they apply to others: every identifier, path, line number and test count
they cite was verified against the tree before committing.

### DIFF
--- a/docs/audits/monize-audit-01-remediation-review-response.md
+++ b/docs/audits/monize-audit-01-remediation-review-response.md
@@ -1,0 +1,278 @@
+# Response to the Phase 1 Remediation Review
+
+**Responds to:** `monize-audit-01-remediation-review.md`
+**Reviewed branch head:** `b5712075` (pre-rebase)
+**Current branch head:** see `git log` — rebased onto `main` at `0a7db0ce`
+**Findings answered:** 5 of 5 (1 HIGH, 1 MEDIUM, 3 LOW) — all resolved
+**Verdict accepted:** yes, including "do not merge as-is". Every finding was
+reproduced before being acted on.
+
+---
+
+## RFR-002 — Branch stale, duplicates a fix already on main · MEDIUM · resolved
+
+Confirmed and decisive; this is why everything else in this response moved.
+
+`main` had independently shipped the same P1-001 fix as
+`135_import_jobs_single_active.sql`: the same partial unique index, under **the
+same name** `idx_import_jobs_one_active_per_user`, the same acquire-before-wipe
+ordering, and the same duplicate-retirement UPDATE. Meanwhile `133` had been taken
+by `133_joint_account_grants.sql`.
+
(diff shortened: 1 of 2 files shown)


WMP:
@kenlasko Decide whether you want to merge this

PR #1056, opened 2026-08-05
02: One audited allowlist for every user-profile response (v1.15.0)

WMP:
## Linked discussion / issue

## Summary

Makes `toUserProfile` (`backend/src/users/user-profile.ts`) the single, audited allowlist through which a `users` row leaves the API (P2-003). Five call sites used to spread the entity and delete the fields they remembered, so the shortest list decided what leaked — and none of them dropped `backupPasswordEnc`, the AES-encrypted copy of the user's backup password. Spreading also strips the class metadata `@Exclude()` needs, so the global serializer could not save those sites. A source-scanning guard test (`user-profile.spec.ts`) now fails for any new `{ passwordHash, ... } = user`-style sanitizer anywhere in `src/`.

Verified locally: backend typecheck, lint, and the users module suites (108 tests across `user-profile.spec.ts`, `users.controller.spec.ts`, `users.service.spec.ts`) on this branch alone against current `main`.

## Checklist

## AI assistance disclosure

---

### DIFF
--- a/backend/src/users/user-profile.spec.ts
+++ b/backend/src/users/user-profile.spec.ts
@@ -1,0 +1,142 @@
+import { readdirSync, readFileSync, statSync } from "fs";
+import { join } from "path";
+import {
+  PROFILE_FIELDS,
+  SELF_ONLY_PROFILE_FIELDS,
+  toDelegatedUserProfile,
+  toUserProfile,
+} from "./user-profile";
+import {
+  excludedUserFields,
+  fullyPopulatedUser,
+  nonProfileUserFields,
+} from "./user-profile.test-util";
+
+describe("user profile serialization", () => {
+  describe("toUserProfile", () => {
+    it("emits exactly the allowlist plus hasPassword", () => {
+      const profile = toUserProfile(fullyPopulatedUser());
+
+      expect(Object.keys(profile).sort()).toEqual(
+        [...PROFILE_FIELDS, "hasPassword"].sort(),
+      );
+    });
+
+    it("drops every @Exclude() column on the entity", () => {
+      const profile = toUserProfile(fullyPopulatedUser()) as Record<
+        string,
+        unknown
+      >;
+      const excluded = excludedUserFields();
+
+      // Guards the guard: an entity that lost its decorators would make the
+      // loop below vacuous and this suite would pass while leaking everything.
+      expect(excluded.length).toBeGreaterThanOrEqual(11);
+      for (const field of excluded) {
+        expect(profile).not.toHaveProperty(field);
+      }
+    });
+
+    it("drops every column the allowlist does not name", () => {
+      const profile = toUserProfile(fullyPopulatedUser()) as Record<
+        string,
+        unknown
+      >;
+      for (const field of nonProfileUserFields()) {
+        expect(profile).not.toHaveProperty(field);
+      }
+    });
+
+    it("serializes no value that came from a secret column", () => {
+      // The fixture marks every secret with a LEAK- prefix, so this catches a
+      // secret copied into a differently named field as well as one passed
+      // through -- neither of which a key-name assertion sees.
+      const serialized = JSON.stringify(toUserProfile(fullyPopulatedUser()));
+      expect(serialized).not.toContain("LEAK-");
+    });
+
+    it("reports hasPassword without the hash", () => {
+      expect(toUserProfile(fullyPopulatedUser())).toMatchObject({
+        hasPassword: true,
+      });
+      expect(
+        toUserProfile(fullyPopulatedUser({ passwordHash: null })),
+      ).toMatchObject({ hasPassword: false });
+    });
+
+    it("leaves a column absent when the caller selected a narrower row", () => {
+      const profile = toUserProfile({ id: "u1", email: "a@b.c" }) as Record<
+        string,
+        unknown
+      >;
+
+      expect(profile).toEqual({
+        id: "u1",
+        email: "a@b.c",
+        hasPassword: false,
+      });
+      expect(profile).not.toHaveProperty("role");
+    });
+  });
+
+  describe("toDelegatedUserProfile", () => {
+    it("keeps identification and drops the owner's credential state", () => {
+      const delegated = toDelegatedUserProfile(fullyPopulatedUser()) as Record<
+        string,
+        unknown
+      >;
+
+      expect(delegated).toMatchObject({
+        id: "owner-1",
+        email: "owner@example.com",
+        firstName: "Olivia",
+      });
+      for (const field of SELF_ONLY_PROFILE_FIELDS) {
+        expect(delegated).not.toHaveProperty(field);
+      }
+    });
+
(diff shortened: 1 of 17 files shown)


WMP:
@kenlasko : I havent idea how to test this, but i thinks that it needed

backend/src/users/user-profile.ts
kenlasko, line 1: Two things on this file.

**1. `SELF_ONLY_PROFILE_FIELDS` describes a property the system does not have after this PR.**

The docblock says "a delegate acting in the owner's context does not receive them". That is true of `GET /users/me`, which this PR converts, and not true of `GET /auth/profile`, which it does not. `auth.controller.ts` (the `getProfile` handler) is also `@AllowDelegate()`, also resolves `req.user.id` — the owner's id while acting — and calls `authService.sanitizeUser`, which is now plain `toUserProfile`. So an acting delegate still receives the owner's `hasPassword`, `mustChangePassword`, `isDelegateOnly` and `backupEncryptionEnabled` from that route.

It matters which route, too: `/auth/profile` is what the frontend actually calls (`frontend/src/store/authStore.ts`), while `userSettingsApi.getProfile` → `/users/me` has no production caller I can find — only tests. So as it stands the protection lands on the unused endpoint and not the used one.

This is the same shape as the defect being fixed: a rule that reads as global and holds at some call sites. Suggest routing `/auth/profile` through `toDelegatedUserProfile` when `req.user.isActing`, with a test asserting the two endpoints agree for an acting delegate. (`/auth/me-self` is correct as-is — it resolves `realUserId`, so the full profile is the caller's own.)

**2. `hasPassword` is asserted even when `passwordHash` was never selected.**

`toUserProfile` is careful to leave an absent column absent, then unconditionally sets `profile.hasPassword = !!user.passwordHash`. For a narrowed `select` that did not fetch the hash, the response says `hasPassword: false` — "this user has no password" — where the truth is "not fetched". `user-profile.spec.ts` locks that in as intended behaviour in the "leaves a column absent when the caller selected a narrower row" case.

No current call site hits it (`auth.service.ts` and `pat.service.ts` are the narrowed selects, and neither goes through the sanitizer), so it is latent rather than live. But this function is now the single door every future caller walks through, and `CLAUDE.md`'s missing-data rule is explicit that a known-unknown must not be reported as a settled `false`. Either omit the key when `passwordHash` is not `in user`, or say in the docblock why `false` is the right answer for an unfetched hash. One line either way.

---

backend/src/users/user-profile.spec.ts
kenlasko, line 1: The `@Exclude()`-metadata fixture and the `LEAK-` sentinel are better than the usual key-name assertions — the sentinel catches a secret copied into a differently named field, which a key check cannot see. Two notes on the guard itself.

**The source scan matches the shape of the code just deleted, and only that shape.** `AD_HOC_SANITIZER` requires `passwordHash` followed by a rest-spread within 600 characters. It does not catch `const { resetToken, passwordHash: _ph, ...rest } = user` (hash not first), a `delete copy.passwordHash`, an `omit(user, [...])`, or the same destructure with a longer field list between the two anchors. That is a narrower guarantee than "no ad-hoc sanitizer exists anywhere in `src/`", which is how the test name and the PR body both read.

Suggest a meta-test that runs candidate leak snippets through the same regex and asserts which are caught, so the boundary is written down and checked rather than inferred — and then either widen the pattern or record the limit in `backend/CLAUDE.md`. A guard's perceived coverage is what the next author relies on, which is the failure this PR is fixing one level up.

**Guard the walk, not just the metadata.** There is already a "guards the guard" assertion for the `@Exclude()` list; the file walk needs the same. If `SRC` ever resolves wrong (a move, a build layout change), `walk()` returns few or no files and the test passes vacuously. `expect(walk(SRC).length).toBeGreaterThan(500)` next to the existing one covers it. Worth a glance at runtime too — this is a `statSync` per entry across all of `backend/src` on every unit run.

---

backend/src/users/users.controller.ts
kenlasko, line 1: This handler is right, and the comment explaining why `@Exclude()` cannot save a spread is worth keeping. Two knock-on effects to check before merge.

**The delegated variant violates the frontend's `User` type.** `frontend/src/types/auth.ts` declares `hasPassword: boolean` and `mustChangePassword: boolean` as required, and `toDelegatedUserProfile` omits both. Nothing in production consumes `/users/me` today so this may not bite, but it means the response and the declared contract disagree. Worth a frontend typecheck plus `settings/page.delegate.test.tsx`, `SecuritySection.test.tsx` and `ProtectedRoute.test.tsx`.

**`AdminUser` currently survives the allowlist by coincidence.** Every field it declares happens to be in `PROFILE_FIELDS`; nothing fails if a later PR removes one. A type-level assertion (`const _: AdminUser = {} as UserProfile`) would make that a compile error rather than an admin-table regression found by hand — the highest-ranked check available for this particular mistake, per the `CLAUDE.md` preference ordering.

---

kenlasko (changes requested):
The core design is right, and the argument in the description is correct: an allowlist is the right shape for this and a removal list is not. Five divergent removal lists all missing `backupPasswordEnc` is exactly what that shape predicts, and deriving the test fixture from class-transformer's `@Exclude()` metadata rather than a hand-copied list is the part that will keep holding.

One blocker and one question, both small.

**Blocker — `/auth/profile` did not get the delegate treatment.** It is also `@AllowDelegate()`, also resolves `req.user.id` (the owner while acting), and still returns the owner's `hasPassword`, `mustChangePassword`, `isDelegateOnly` and `backupEncryptionEnabled` to an acting delegate. That is the endpoint the frontend actually calls; `/users/me`, which this PR does convert, has no production caller. So the docblock on `SELF_ONLY_PROFILE_FIELDS` asserts a property the system does not have — which is the same class of defect as the one being fixed. Details in the inline comment on `user-profile.ts`.

**Question — `hasPassword` on a narrowed row.** `toUserProfile` always emits it, so a `select` that skipped the hash yields `hasPassword: false`: a claim, where the truth is "not fetched". No live call site hits it, but this is now the single door, and the spec locks the behaviour in. Omit the key, or write down why `false` is right.

**Testing.** This changes five response shapes, and the guards are all unit-level on `toUserProfile` plus one mocked controller test — the claim being made is about HTTP responses. `backend/test/` already has supertest infrastructure; the `LEAK-` sentinel works just as well there (`expect(JSON.stringify(res.body)).not.toContain('LEAK-')`) across login, 2FA verify, both `/auth` profile routes, `/users/me`, `PATCH /users/profile` and the three admin surfaces. That is the check that would have caught `/auth/profile`, and it is the kind a unit test on the serializer structurally cannot make.

Related: how many of the 108 tests fail on `main` without the fix? If the suite was green before, the guards do not yet cover P2-003 in its original form, and per `CLAUDE.md` a behaviour change with no failing case is itself something to report and close in the same commit.

Happy to re-review once `/auth/profile` is routed through `toDelegatedUserProfile`; nothing here needs a redesign.

---

PR #1058, opened 2026-08-05
02: Bind an MCP session to the credential that opened it (v1.15.0)

WMP:
## Linked discussion / issue

## Summary

Binds an MCP session to the credential that opened it, and re-derives scopes from the credential presented *now* on every request (P2-004). A session used to store the scopes of the token that created it and match subsequent requests by user only — but one user holds many tokens with different scopes, so a read-only PAT that could reuse a write session could write. Cached authorization is not authorization: the session is now bound to one credential fingerprint, and a different token — even the same user's — cannot ride it.

Verified locally: backend typecheck, lint, and the full MCP module suite (27 suites, 479 tests) on this branch alone against current `main`.

## Checklist

## AI assistance disclosure

---

### DIFF
--- a/backend/src/auth/pat.service.spec.ts
+++ b/backend/src/auth/pat.service.spec.ts
@@ -253,6 +253,8 @@ describe("PatService", () => {
       expect(result).toEqual({
         userId: "user-1",
         scopes: "read",
+        // The credential fingerprint the MCP transport binds a session to.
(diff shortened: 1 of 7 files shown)


WMP:
It needs to test mcp before merge.

kenlasko:
Tested MCP auth and several queries. Worked fine

PR #1060, opened 2026-08-05
02: OIDC re-authentication and OAuth grant-integrity hardening (v1.15.0)

WMP:
## Linked discussion / issue

## Summary

OIDC re-authentication and OAuth grant-integrity hardening, shipped in final form (P2-005 + FV-001 + the OAuth consent grant-ownership check).

**The second factor for the four most destructive operations collapsed into possession of the first.** Account deletion, data deletion, backup restore and the .mny import's wipe-first mode "re-authenticated" an OIDC account by checking that a client-supplied string was non-empty — and the frontend sent the literal `"oidc-session-confirmed"`. The step-up service's OIDC branch accepted `oidcConfirmed: true`, a boolean the client asserted. A local account with no password (admin-provisioned, reset not completed) fell off the end of the `else if` and proved nothing at all. Replacing it: `GET /auth/oidc/reauth?purpose=X` sends the user to the identity provider with `prompt=login` and `max_age=0`; the callback mints a five-minute signed artifact bound to the user, the action and a one-time id; the destructive handler consumes it — signature, type, subject, action, freshness, single use.

**The artifact is bound to its own round trip (FV-001), not merely to the fact that one was started.** Without that, the pending marker outlived its redirect: start `?purpose=delete-account`, do not follow it, complete an ordinary `GET /auth/oidc` — no `prompt=login`, answered silently by a live SSO session — and the callback minted a valid delete-account artifact for a challenge that never happened. The marker now carries a hash of the `state` it was issued with, the callback compares it against the state it actually validated, and an ordinary login clears a pending marker rather than silently discarding it.

**OAuth consent adopts a grant only after checking whose it is.** The consent screen took `interaction.grantId`, loaded that grant and widened it, having verified only that the *interaction* belonged to the session — a grant naming another account or another client would have been widened on this user's click. Every field that has to match is compared (`accountId` **and** `clientId`), with a fresh object as the fallback.

Also: backup restore validates everything free before spending the single-use artifact (a mistyped backup password no longer costs a full IdP round trip), a dead ID-token claim check nothing called is deleted, and the two integration modules that construct `UsersService` now provide its new dependency. All strings this PR adds or rewords are translated for every locale in the same PR, pseudo-locales regenerated.

Verified locally: backend typecheck, lint, auth/users/backup/import/oauth/i18n suites (880 + 832 tests across the two runs) and frontend typecheck + callback/step-up/restore/danger-zone/i18n-parity tests, on this branch alone against current `main`.

## Checklist

## AI assistance disclosure

---


WMP:
Before merge we need to check SSO and backup restore operation.

PR #1062, opened 2026-08-05
02: Scope maintenance refreshes by the operation, not the caller (v1.15.0)

WMP:
## Linked discussion / issue

## Summary

Scopes the two maintenance refreshes by the operation, not the caller, and stops a global sweep from enumerating other tenants.

`refreshAllPrices` and `refreshAllRates` inherited their tenant scope from whoever called them: the cron wrapped them in a system context and the endpoints did not, which read as global at `RLS_MODE=off` and silently narrowed to the requesting user's own rows at `enforce` — a maintenance endpoint that stops doing maintenance the moment enforcement is switched on. Both now seed their own system context, so the reach is a property of the operation rather than of the call site. The scoping tests fail with the system context moved back into the cron only.

And `POST /currencies/exchange-rates/refresh`, which any authenticated user may call, returned the per-pair results: the set of currencies everybody else on the deployment transacts in, which on a small self-hosted instance is an inference about named people's finances. It returns counts now — all the UI ever read; the per-pair detail stays in the operator's log.

Verified locally: backend typecheck, lint, and the currencies + securities suites (33 suites, 1442 tests) on this branch alone against current `main`.

## Checklist

## AI assistance disclosure

---

### DIFF
--- a/backend/src/currencies/currencies.controller.spec.ts
+++ b/backend/src/currencies/currencies.controller.spec.ts
@@ -321,13 +321,48 @@ describe("CurrenciesController", () => {
   });
 
   describe("refreshRates()", () => {
-    it("delegates to exchangeRateService.refreshAllRates", () => {
-      mockExchangeRateService.refreshAllRates!.mockReturnValue("summary");
+    const summary = {
+      totalPairs: 3,
+      updated: 2,
+      failed: 1,
+      lastUpdated: new Date("2026-08-03T00:00:00Z"),
+      results: [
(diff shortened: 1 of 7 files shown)


WMP:
Before going live, you need to carry out a test – to see how the currency exchange rate refresh works now, and after this change. You need to check whether all exchange rates are being refreshed, or just individually.

kenlasko:
Didn't see any difference in behaviour, but I only have 2 currencies in total (CAD, USD)

PR #1063, opened 2026-08-05
02: Boot-time runtime role verification: attributes, ownership, membership (v1.15.0)

WMP:
## Linked discussion / issue

## Summary

Boot-time verification that the database role serving `RLS_MODE=enforce` traffic actually is unprivileged — attributes, ownership, membership, and every context the connection can reach — with the claims proven against a live PostgreSQL (P2-006, DR-02, DR-V2/DR-R1, RR3–RR7).

Selecting a role name and supplying its password says nothing about `rolsuper`, `rolbypassrls`, or what the role owns — and PostgreSQL exempts all three from every policy. The check asks the connection what it is and refuses to serve enforced traffic on a wrong answer. The predicate set is shipped here in final form, because each intermediate form passed in a case it did not think of, and several of those failures were *measured*, not argued:

- attributes are a `pg_has_role(..., 'SET')` question (not inherited), ownership a `'USAGE'` question (inherited) — a SET-only predicate reported `GRANT owner TO app WITH INHERIT TRUE, SET FALSE` as safe while the connection saw every tenant's rows;
- routes compose: `app --SET--> bridge --INHERIT--> owner` gives SET=false and USAGE=false from `app`, yet `SET ROLE bridge` bypasses every policy — so the check enumerates reachable contexts and asks what each can reach;
- the forbidden-attribute list is derived from the same `APP_ROLE_ATTRIBUTES` the provisioner writes, with a parity guard — `NOREPLICATION` was provisioned and never checked, so a `REPLICATION` role booted clean;
- membership in a predefined role grants what `ALTER ROLE` cannot strip: a `NOSUPERUSER` member of `pg_execute_server_program` wrote a host file, a member of `pg_signal_backend` killed another role's backend — both measured. Every predefined role is classified forbidden or known-safe, and an unclassified one fails the boot rather than defaulting to safe;
- `array_agg(rolname)` returns `name[]`, which node-postgres renders as the literal string `"{}"` — and `"{}".length > 0` is `true`, which refused every enforce-mode boot on a violation that did not exist. Found by the live run, not the unit suite, whose mock returned the array the code wanted.

Two new integration suites run against real PostgreSQL: `runtime-role.integration.spec.ts` (grant surface, transitive reachability, inherited ownership, live capability proofs, complete predefined-role classification) and `schema-entity-parity.integration.spec.ts` (56 of 113 FKs disagreed between entity metadata and `schema.sql` on their delete rule, baselined shrink-only; the one that mattered, `user_currency_preferences.currency_code`, is fixed by declaring the CASCADE the schema has).

Built from current `main` with no dependency on the other PRs in this series (verified by building and testing in isolation, not assumed).

Verified locally: backend typecheck, lint, `runtime-role-check.spec.ts` + `app-role.spec.ts` (58 tests) on this branch alone. The live integration suites need PostgreSQL — expected to run in CI.

## Checklist

## AI assistance disclosure

---

### DIFF
--- a/backend/src/common/db/app-role.spec.ts
+++ b/backend/src/common/db/app-role.spec.ts
@@ -9,6 +9,7 @@ import {
   SqlClient,
 } from "./app-role";
 import { DEFAULT_APP_USER } from "./rls-config";
+import { FORBIDDEN_RUNTIME_ATTRIBUTES } from "./runtime-role-check";
 
 function makeClient() {
   const calls: Array<{ text: string; params?: unknown[] }> = [];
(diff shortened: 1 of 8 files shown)


WMP:
This need simple test if application can start before merge.

kenlasko:
Built image from this branch and pushed to my prod. It started fine with nothing out-of-the-ordinary in the logs

PR #1075, opened 2026-08-05
03: Repository guards that turn three prose rules into checks (v1.15.0)

WMP:
**Merge order: 1 of 4.** Base `main`. Nothing depends on this being reviewed first, but it is the
cheapest and safest, and the three PRs above it are stacked on it.

## What this is

Three rules this repository already had in prose, each of which **was broken precisely because prose
does not fail a build**:

- a `CLAUDE.md` or top-level `docs/*.md` naming a file that does not exist (a plan doc sent readers to
  `backend/src/ai/query/` for a service that lives under `backend/src/transactions/`);
- a migration creating roles or grants outside the documented carve-outs;
- a `@Cron` handler nobody wrote down (six were missing from `docs/cron-jobs.md`).

Each is now a test. The fourth change is a comment in the .mny staging service that claimed a filter
the query does not have.

## How to test it (5 minutes, nothing to click)

```bash
cd backend
npm run migration:lint      # prints the file count, then OK
npx jest src/common/doc-paths src/common/cron-doc
node scripts/migration-lint.test.mjs
```

Worth doing once, because a guard nobody has watched fail is not evidence: add a line naming
`docs/nie-ma-mnie.md` to any `CLAUDE.md`, re-run the doc-paths test, and see it go red. Then revert.

## Risk

None at runtime — no application code changes. If CI is red here, a document is wrong, not the app.

## Not in this PR

`docs/future-plans/` is deliberately out of the doc-path guard's scope: naming files that do not exist
yet is what a plan is for. `docs/release-notes/` too, being a shipped record that must not be edited to
match a later tree.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity) -- no user-facing strings.

## AI assistance disclosure

---


database/CLAUDE.md
```diff
@@ -95,12 +95,22 @@ GUCs through `withScopedDb` and the policies compare each row's owner against th
...
-   `ALTER DEFAULT PRIVILEGES`.
+   `ALTER DEFAULT PRIVILEGES`. The `role-or-grant-statement` rule in
+   `backend/scripts/migration-lint.mjs` enforces this over every migration, so it is a CI failure
+   rather than a convention.
+
+   **`PUBLIC` is the exception, for the same reason and not in spite of it:** it is a keyword that
+   always resolves, so it cannot fail for a missing role. It has to be permitted, because
+   `CREATE FUNCTION` grants `EXECUTE` to `PUBLIC` implicitly -- revoking that anywhere but the
+   transaction that created the function leaves a window in which any role can execute a fresh
+   `SECURITY DEFINER` function. `133_currency_global_liveness.sql` is the case. This paragraph used
```
kenlasko: This carve-out is justified by a migration and a function that do not exist in this tree.

Checked against `refs/pull/1075/head`:

- migration 133 is `133_joint_account_grants.sql`; there is no `133_currency_global_liveness.sql`
- `grep -rn "currency_code_in_use_globally\|currency_global_liveness" .` — zero hits
- `grep -rn "SECURITY DEFINER" database/` — zero hits
- there is **no** `GRANT` or `REVOKE` statement anywhere in `database/migrations/` or `database/schema.sql`. Every hit for those words is inside a comment. All grants live in `APP_ROLE_GRANTS_SQL` in `backend/src/common/db/app-role.ts`, exactly as `111_rls_helpers_and_trigger.sql` line 8 says they do.

So the last sentence — "This paragraph used to read as an absolute ban while that `REVOKE` was already in the tree, which is a rule nothing checked disagreeing with the code it governs" — is itself a false claim about the source. The absolute ban was being obeyed. As written, this PR relaxes a rule the tree currently satisfies, to permit a pattern the tree does not use, on a premise that isn't there.

That matters more than usual here, because it is the failure mode the PR exists to prevent, and the PR's own guard cannot catch it: `PATH_IN_BACKTICKS` in `doc-paths.spec.ts` requires an interior `/`, so a bare filename claim like `` `133_currency_global_liveness.sql` `` is never checked.

Two ways out, both fine:

1. Keep the ban absolute and drop the exception. The tree satisfies it today, and `CREATE FUNCTION ... SECURITY DEFINER` is not something any migration currently does.
2. Keep the exception on its forward-looking merit — it is a real Postgres hazard — but delete every provenance claim: the migration name, the function name, and the "the code already disagreed" sentence, here and in the matching comments in `migration-lint.mjs` and `migration-lint.test.mjs`.

---

backend/src/common/doc-paths.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+ * file and grep `docs/` and every `CLAUDE.md` in the same commit -- and it was
+ * still broken: the cross-owner-transfers plan pointed at
+ * `backend/src/ai/query/transaction-tool-prep.service.ts` for a service that
+ * lives under `backend/src/transactions/`, so anyone following the path would
+ * find nothing and either hunt for it or create a second one. A rule in prose
+ * gets read, agreed with, and violated anyway; this is the version the machine
+ * checks.
+ *
+ * Scope is the binding contracts -- every `CLAUDE.md` plus the top level of
+ * `docs/` -- and deliberately not `docs/future-plans/`, whose whole job is to
```
kenlasko: The excluded directory is the one the motivating defect lived in.

Lines 9-11 above cite the cross-owner-transfers plan pointing at `backend/src/ai/query/transaction-tool-prep.service.ts` as the reason this guard exists. That file is `docs/future-plans/cross-owner-transfers.md` — out of scope here, so the guard would not have caught the bug it is named after. Three of the four doc-path fixes in this PR are in files the test never reads.

I ran this spec's exact scope rules against `main`. It finds exactly one broken path: `docs/ms-money-data-model.md: migration/ms-money-data-model.md`.

The stated reason for the exclusion — "naming files that do not exist yet is what a plan is for" — is true, but it does not cover the actual failure mode, which is a *moved* file rather than an unbuilt one. Those two separate cleanly on one condition: does the basename exist somewhere else in the tree?

Extending the **rooted** check into `docs/future-plans/` under that qualifier, measured against `main`:

- **5 findings** — the 3 this PR hand-fixed, plus **2 it missed**: `docs/future-plans/row-level-security-tasks.md` and `docs/future-plans/row-level-security.md` both name `` `database/seed.service.ts` ``, which is `backend/src/database/seed.service.ts`.
- **0 false positives.** All 10 genuinely-unbuilt paths those plans name stay silent, because no file of that basename exists anywhere.

That the manual pass caught 3 of 5 is the argument for making the machine do it — the same argument the PR makes for everything else in it. Roughly:

```ts
const movedOnly = (path: string): boolean =>
  filesByBasename.has(basename(path));  // named elsewhere => moved, not unbuilt
```

---

backend/src/common/doc-paths.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+ *  - A **rooted** path (`backend/src/...`, `database/...`) is an exact claim and
+ *    must resolve exactly. This is the one that was wrong.
+ *  - A **relative** path (`map/map-loans.ts` inside a section about one
+ *    directory) is a readable shorthand, so it only has to match some file's
+ *    path suffix somewhere in the tree. That still catches the failure mode that
+ *    matters -- a renamed or deleted file -- without forcing every mention to
+ *    carry six directories of prefix.
+ */
+describe("docs name files that exist", () => {
+  const REPO_ROOT = resolve(__dirname, "..", "..", "..");
```
kenlasko: This assumes a full-repo checkout, and fails badly rather than clearly when it doesn't have one.

`docker-compose.dev.yml` line 86 mounts `./backend:/app`. Inside the project's own dev container `__dirname` is `/app/src/common`, so `REPO_ROOT` resolves to `/` — and `walk(REPO_ROOT)` on line 100 then recurses the container root (`/proc`, `/sys`, `/usr`, …) before `readdirSync(join(REPO_ROOT, "docs"))` throws. Both `collectDocs()` and `allFiles()` run in the `describe` body, so what you get is a collection error after a long stall, not a readable assertion.

`cron-doc.spec.ts` has the same shape at its `DOC_PATH`, though it throws immediately instead of walking.

CI is unaffected — full checkout, and `ci.yml` has no path filters — so this only bites someone running `npm run test:unit` in the container the root `CLAUDE.md` tells them everything runs in.

Cheapest fix that keeps the guard honest: walk up for a marker (`.git`, or the root `CLAUDE.md`) and throw one explicit error naming the assumption when it isn't found. Not a silent skip — that would quietly disable the guard in CI if the marker check ever regressed.

---

backend/scripts/migration-lint.mjs
```diff
@@ -416,6 +416,48 @@ export const RULES = [
...
+      const messages = [];
+      if (/\b(?:CREATE|ALTER|DROP)\s+(?:ROLE|USER|GROUP)\b/i.test(text)) {
+        messages.push(
+          "CREATE/ALTER/DROP ROLE in a migration crash-loops startup where the role differs; roles live in db-init (common/db/app-role.ts)",
+        );
+      }
+      const grant = /\b(GRANT|REVOKE)\b([\s\S]*)$/i.exec(text);
+      if (grant) {
+        // The grantee follows TO (GRANT) or FROM (REVOKE).
+        const grantee = /\b(?:TO|FROM)\s+([A-Za-z_"][\w"$]*)/i.exec(grant[2]);
```
kenlasko: Only the head of the grantee list is read, and a grantee list is legal SQL:

```sql
REVOKE ALL ON t FROM PUBLIC, monize_app;   -- passes clean
```

`exec` returns the first match, `grantee[1]` is `PUBLIC`, and the named role behind it is never looked at. The rule needs every grantee after `TO`/`FROM` to be `PUBLIC`, not just the first — split on commas up to the next keyword and check them all.

Separately, the rule's scope is narrower than its rationale. The stated reason is "a statement naming that role fails where the role does not exist", but three statements that do exactly that are unflagged:

- `ALTER TABLE t OWNER TO monize_app` — no `GRANT`/`REVOKE` keyword, so the second branch never fires
- `SET ROLE monize_app` — not `CREATE|ALTER|DROP ROLE`
- `REASSIGN OWNED BY monize_app TO ...` / `DROP OWNED BY monize_app`

`ALTER ... OWNER TO` is the plausible one in practice.

`migration-lint.test.mjs` only exercises single grantees, so the multi-grantee case should land there in the same change.

---

backend/scripts/migration-lint.test.mjs
```diff
@@ -203,6 +203,51 @@ test("INSERT without ON CONFLICT is flagged; ON CONFLICT and WHERE NOT EXISTS pa
...
+  flags('GRANT SELECT ON t TO "monize_app";', "role-or-grant-statement");
+
+  clean("REVOKE ALL ON FUNCTION f(VARCHAR) FROM PUBLIC;");
+  clean("REVOKE ALL ON FUNCTION f(VARCHAR) FROM public;");
+  clean("GRANT SELECT ON TABLE t TO PUBLIC;");
+  // The word appearing in prose or an identifier is not a grant statement.
+  clean("COMMENT ON FUNCTION f() IS 'EXECUTE is revoked from PUBLIC';");
+});
+
+test("the shipped migrations satisfy the role rule", () => {
```
kenlasko, line 342: This test reads no shipped migration. It calls `clean()` on an inline string, which duplicates a `clean()` case from the test directly above it — and the string names a function (`currency_code_in_use_globally`) and a migration that do not exist in the repo (see the `database/CLAUDE.md` comment).

So the comment "Migration 133 revokes EXECUTE from PUBLIC on a `SECURITY DEFINER` function" describes something that isn't in the tree, and the stated purpose — "so a future tightening of the rule to an absolute ban fails loudly instead of silently making the directory unlintable" — is not what the test does: tightening the rule would fail this test whether or not any migration was affected, because nothing here touches `database/migrations/`.

If the intent is to pin the real directory, lint it:

```js
test("the shipped migrations satisfy the role rule", () => {
  const { findings } = lintDirectory(MIGRATIONS_DIR);
  assert.deepEqual(findings.filter((f) => f.rule === "role-or-grant-statement"), []);
});
```

That is a real assertion about the tree, it would have surfaced that no `REVOKE ... FROM PUBLIC` exists today, and it costs nothing since `npm run migration:lint` already walks the same directory in CI.

---

docs/cron-jobs.md
```diff
@@ -19,4 +19,14 @@ Cron jobs use the `@Cron()` decorator from `@nestjs/schedule`. They run in the A
...
 | `mny-staging.service` | Hourly | Delete expired staged import files (24 h TTL) |
 | `mny-import-job.service` | Every 5 min | Fail import jobs whose worker stopped heartbeating |
-| `auto-backup.service` | Hourly | Enrol every non-admin user on the default backup policy, then run the backups that are due |
+| `auto-backup.service` | Hourly | Enrol every non-admin user on the default backup policy, then write each user's due automatic backup, promote weekly/monthly copies, enforce retention |
+| `action-history.service` | Daily 3 AM | Delete undo-log entries past their retention window |
+| `holdings.service` | Every hour at :30 | Apply matured fixed-term investment holdings |
+| `emergency-access-monitor.service` | Daily 9 AM | Advance emergency-access requests past their waiting period and notify |
+| `updates.service` | Every 12 hours | Refresh the cached latest-release metadata for the What's New digest |
+
+Every row above is checked against the source by `backend/src/common/cron-doc.spec.ts`: a
```
kenlasko: "Every row above is checked against the source" is checked for *membership* only — never for what the row says. The table still contains a row that contradicts the source, and the new guard now certifies it.

Line 14: `` `accounts.service` | Midnight daily | Account maintenance ``

The handler is `@Cron("0 * * * *")` on `applyDueTransactionBalances` (`backend/src/accounts/accounts.service.ts:1602`) — hourly, and it applies due transaction balances. Wrong on both the schedule and the purpose.

The membership half is genuinely correct and worth having: 19 services declare `@Cron` (23 handlers) and the post-PR table lists exactly those 19. `auth.service` → `token.service` is a real fix. It's only the "every row is checked" sentence that overclaims.

Minimum: correct line 14. If you want the schedule checked too, carry the literal cron expression in the cell (`` `0 * * * *` ``) alongside the prose and compare it against the decorator argument — `@Cron` takes a string literal or a `CronExpression` member at every one of the 23 sites, so it parses exactly.

---

backend/src/common/cron-doc.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (statSync(full).isDirectory()) {
+          if (entry !== "node_modules") walk(full);
+          continue;
+        }
+        if (!entry.endsWith(".ts") || entry.endsWith(".spec.ts")) continue;
+        const source = readFileSync(full, "utf8");
+        const handlers = (source.match(/^\s*@Cron\(/gm) ?? []).length;
+        if (handlers === 0) continue;
+        // The doc names services as `foo.service`, matching the file stem.
+        found.set(basename(entry).replace(/\.ts$/, ""), {
```
kenlasko: Two problems with keying on the file stem.

**Collision.** Two files with the same basename in different directories collapse into one map entry, and the second silently overwrites the first — so the survivor's handler count is used and the other file vanishes from the comparison entirely. No collision exists today, but nothing here would tell you when one appears.

**A `@Cron` outside a `*.service.ts` makes the suite unfixably red.** Any file with a `@Cron` lands in `services`, but `documented` on line 51 is built from ``/^\|\s*`([a-z0-9-]+\.service)`/gm`` — so a handler added to a `*.controller.ts`, `*.job.ts` or `*.task.ts` can never be matched by a doc row. "documents every service that declares a @Cron handler" then fails, and the only way to make it pass is to write a row the doc regex cannot read. The next person hits a red test with no legal fix.

Either key on the repo-relative path and widen the doc column regex to `[a-z0-9-]+\.[a-z]+`, or — probably better, and closer to what the table means — add an explicit assertion that every `@Cron` file is a `*.service.ts`, so the constraint fails on its own terms with a message that says what to do.

---

backend/src/common/doc-paths.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    "docs/",
+    "helm/",
+    "scripts/",
+    "e2e/",
+    ".github/",
+    ".claude/",
+  ];
+
+  const EXTENSIONS = "ts|tsx|sql|mjs|cjs|js|json|sh|ya?ml|md";
+  const PATH_IN_BACKTICKS = new RegExp(
```
kenlasko: Minor, but two of these bear on findings elsewhere in this PR.

**A path with no interior `/` is never checked.** The pattern requires `…/…` , so `` `schema.sql` ``, `` `133_currency_global_liveness.sql` `` and every other bare filename claim is invisible. That is precisely how the nonexistent migration name in `database/CLAUDE.md` got past this PR's own guard. Checking bare filenames when the basename is unique in the tree would close it; if that's too noisy, the `CLAUDE.md` paragraph should say so rather than reading as full coverage.

**The `branch:path` escape hatch is an untested implicit contract.** `` `poc/import-from-dotmny:migration/ms-money-data-model.md` `` is ignored only because `:` happens to be absent from the character class on the next line. The root `CLAUDE.md` now documents that idiom, so it deserves an explicit case here — otherwise widening `EXTENSIONS` or the class later silently starts failing on it, in a doc nobody touched.

---

backend/src/common/doc-paths.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    // repository content -- CI never has it, a contributor's clone may.
+    ".claude",
+    "node_modules",
+    "dist",
+    "coverage",
+    ".next",
+    "test-results",
+    "playwright-report",
+    ".turbo",
+  ]);
```
kenlasko: Nit, and it removes this list entirely: drive both walks from `git ls-files` instead of the filesystem.

As written, the relative-path check on line 143 suffix-matches against whatever is on disk, so an untracked local file can satisfy a claim that would not resolve in CI — the guard's verdict depends on the contributor's working tree. `SKIP_DIRS` is also a hand-maintained denylist that has to grow every time a tool adds an output directory.

`git ls-files` gives the tracked tree exactly as CI sees it, makes `node_modules`/`dist`/`.next` skipping free, and drops this whole `Set`.

While you're in here: `collectDocs()` and `allFiles()` each walk the entire tree independently. One pass collecting both would do, and matters a bit more once the walk is 3.6k files.

---

kenlasko (commented):
Reviewed by checking each claim against the tree at `main` (`c605d2b`) and at `refs/pull/1075/head`, and by re-implementing both new guards as throwaway scripts to see what they actually flag.

## What's right

- The cron table is now correct on membership: 19 services declare `@Cron` (23 handlers), and the post-PR table lists exactly those 19. `auth.service` → `token.service` is a real fix.
- Both specs carry vacuity guards (`services.size > 10`, `files.length > 500`). That instinct is rare and worth naming.
- The `mny-staging.service.ts` comment fix is accurate — `findInfo` is `findOne({ where: { id, userId } })`, no `expires_at` predicate. The old comment did claim a filter that is not there.
- Splitting the migration-lint failure message into two remedies is a genuine usability improvement.

## The two that matter

**1. The `PUBLIC` carve-out is justified by a migration and a function that do not exist** (inline on `database/CLAUDE.md`). Migration 133 is `133_joint_account_grants.sql`; `currency_code_in_use_globally` appears nowhere; `SECURITY DEFINER` appears nowhere in `database/`; and there is no `GRANT`/`REVOKE` statement in `database/migrations/` or `schema.sql` at all — every hit is inside a comment. The absolute ban was being obeyed. This PR relaxes it on a premise that isn't in the tree, which is the failure mode the PR exists to prevent.

**2. The doc-path guard excludes the directory the motivating defect lived in** (inline on `doc-paths.spec.ts`). The cross-owner-transfers plan is under `docs/future-plans/`, which is out of scope, so the guard would not catch the bug it is named after. Running the spec's own scope rules against `main` finds one broken path. Extending the rooted check into `future-plans/` under a moved-file qualifier — basename exists somewhere else in the tree — gives 5 findings and 0 false positives on `main`: the 3 fixed here, plus 2 this PR missed (`row-level-security.md` and `row-level-security-tasks.md` both name `` `database/seed.service.ts` ``, which is `backend/src/database/seed.service.ts`). 3 of 5 by hand is the argument for the guard.

## Also inline

- `migration-lint.mjs` — only the first grantee is read, so `REVOKE ALL ON t FROM PUBLIC, monize_app;` passes clean; `ALTER ... OWNER TO`, `SET ROLE` and `REASSIGN OWNED BY` are out of scope despite matching the rationale.
- `migration-lint.test.mjs` — "the shipped migrations satisfy the role rule" reads no shipped migration.
- `cron-doc.spec.ts` / `doc-paths.spec.ts` — both resolve the repo root by relative traversal, which lands on `/` inside the dev container (`docker-compose.dev.yml:86` mounts `./backend:/app`), where `doc-paths.spec.ts` walks the container filesystem before throwing. CI is unaffected.
- `cron-doc.spec.ts` — keying on the file stem collides silently, and a `@Cron` outside a `*.service.ts` makes the suite red with no legal fix.
- `docs/cron-jobs.md` — line 14 still says `accounts.service` runs "Midnight daily / Account maintenance"; it is `@Cron("0 * * * *")` on `applyDueTransactionBalances`. The new guard checks that rows exist, not what they say.
- Two nits on `doc-paths.spec.ts`: bare filenames are never checked (which is how finding 1 slipped past this PR's own guard), and the `branch:path` escape hatch works only by accident of the character class.

## CI

"Verify PR checklist" is the only red check and the reason for `mergeable_state: blocked` — `pr-checklist.yml` applies to outside contributors, and the body leaves `- [ ] An approved discussion or issue exists…` unticked with "Approved in:" blank. Everything else is green.

Submitting as a comment rather than a change request — the direction is right and most of this is mechanical, but findings 1 and 2 both go to whether the change does what it says on the tin, so I'd want them resolved before merge.

---

kenlasko (changes requested):
Changing my verdict on the review above from a comment to a request for changes. The findings are unchanged and the inline threads still hold the detail — this only corrects the level I filed them at, since two of them go to whether the change does what it says rather than to how it does it.

The two that block:

1. **`database/CLAUDE.md:111`** — the `PUBLIC` carve-out cites `133_currency_global_liveness.sql` and `currency_code_in_use_globally`, neither of which exists in this tree, and there is no `GRANT`/`REVOKE` statement in `database/migrations/` or `schema.sql` at all. The absolute ban this relaxes was being obeyed, so the sentence "that `REVOKE` was already in the tree" is the same kind of false claim about the source that the rest of the PR is built to catch. Either keep the ban absolute, or keep the exception and drop the provenance.

2. **`backend/src/common/doc-paths.spec.ts:17`** — the guard excludes `docs/future-plans/`, which is where the defect it is named after lived, so it would not have caught it. A moved-file qualifier (basename exists elsewhere in the tree) gives 5 findings and 0 false positives on `main`, including 2 this PR does not fix.

The other seven comments are non-blocking — take them or leave them.

Worth saying plainly, because a request-for-changes reads harsher than I mean it: the idea here is right, and three of the four things this PR sets out to do it does correctly. The cron table is accurate on membership now, the vacuity guards in both specs are better practice than most repositories manage, and the `mny-staging` comment fix is exactly the kind of correction that usually never gets made. It is worth finishing.

---

WMP:
Will be fixed

kenlasko (approved):
Re-reviewed at `868e04c`. All nine findings are addressed, and I checked the head ref rather than reading the commit messages. Clearing the change request.

| # | What I checked | |
|---|---|---|
| 1 | `database/CLAUDE.md` now says "No shipped migration uses this yet (the tree holds no `SECURITY DEFINER` function, and every grant lives in `app-role.ts`)". The invented migration and function are gone from the doc, the rule comment and the test. | fixed |
| 2 | `planProblem` + `planDocs` bring `docs/future-plans/` in under the moved-file qualifier, unit-tested against a synthetic index. Both `seed.service.ts` references corrected, plus a third in `frontend/CLAUDE.md` I had not found. | fixed |
| 3 | `findRepoRoot` walks up for markers; `REPO_ROOT \|\| process.env.CI ? describe : describe.skip` skips locally and hard-fails on CI through `requireRepoRoot`. Precedence is right, and the CI branch does throw rather than pass. | fixed |
| 4 | Ran 20 adversarial statements through `lintSql`. Every case passes, including `FROM PUBLIC, monize_app`, `OWNER TO`, `SET ROLE`, `REASSIGN OWNED BY`, and `granted_to` in a policy predicate staying clean. The `CREATE/ALTER POLICY … TO` clause is beyond what I raised. | fixed |
| 5 | Cron column added, 23 rows for 23 handlers, `accounts.service` corrected to `0 * * * *` with an accurate purpose. `scheduleDiscrepancies` compares sorted multisets, so a surplus doc row fails too — the asymmetry I did not ask about. | fixed |
| 6 | `aggregateByStem` merges same-stem files instead of overwriting; the row grammar no longer requires `.service`; an unparseable decorator is a loud failure naming the file. | fixed |
| 7 | `git ls-files` replaces the walk, `SKIP_DIRS` is gone, and existence is generous (`--cached --others --exclude-standard`) while the gating docs are tracked-only. | fixed |
| 8 | `classifySpan` has a `bare` kind, pinned with `133_currency_global_liveness.sql` as its test case. | fixed |
| 9 | `branch:path` is an explicit `cross-tree` classification with tests, and `path:42` is checked with the lines stripped. | fixed |
| 10 | `lintDirectory` over the real `database/migrations`, guarded by `files.length > 100`. | fixed |
| 11 | One `git ls-files` call replaces both walks. | fixed |

Two things I verified independently rather than taking on trust: the source has 23 anchored and 23 unanchored `@Cron` occurrences, so the two regexes in `parseCronDecorators` agree today; and fence-stripping in `docSpans` skips 0 path-shaped spans across the contract docs, so that exemption costs no coverage.

Residual nits, none blocking and none worth another round on their own:

- `repo-tree.util.ts` builds a shell string for `execSync`; `execFileSync("git", [...])` skips the shell. Args are hardcoded, and this is the one non-spec file the change adds to `src/`.
- `CRON_DECORATOR` is unanchored while `anchors` is `^`-anchored, so a commented-out `@Cron(` would report "0 @Cron line(s), 1 parsed". Latent — the counts agree today.
- `GRANT`/`REVOKE` inside a string literal false-positives (`VALUES ('grant access to the delegate')`), though the `SET ROLE` branch was deliberately anchored against exactly that. The pragma covers it.
- `ALTER TABLE t OWNER TO CURRENT_USER` is flagged though `CURRENT_USER` always resolves — the same shape as the `PUBLIC` carve-out.
- `parseCronDocRows` excludes underscores from the stem, so `auto_backup.service.ts` could not be documented.

CI is green apart from "Verify PR checklist", which is the unticked approval box and the blank "Approved in:" line — yours to clear, not a code problem.

Worth saying: the two blocking findings were the kind that are easy to argue away, and both were answered by making the guard stronger rather than the claim weaker. The `PUBLIC` allowance now states plainly that nothing in the tree uses it yet, which is a harder and more useful sentence than the one it replaced.

---

PR #1076, opened 2026-08-05
03: Serialise startup database work, and refuse an unsafe runtime role (v1.15.0)

WMP:
> **Stacked on #1075, which is still open.**
> diff until then.

**Merge order: 2 of 4**, after PR 1 (#1075). Targets `main`, like the rest of the series, so there is
nothing to retarget -- this PR's diff narrows to its own commit the moment PR 1 lands.

## What this is

Two startup defects, plus the grant convergence both of them need.

**Startup work was not serialised.** With more than one backend replica, `db-init` and `db-migrate`
could run their DDL concurrently against the same database. An advisory lock now makes one replica do
the work while the others wait for it.

**The backend accepted a database role that can reach privilege it must not have.** The original check
tested role *inheritance* — but a role can also acquire power through `SET ROLE` without inheriting it,
and membership in a predefined role with host capability is a third route. An application able to
bypass its own row-level security is not something to warn about and continue from, so this refuses at
startup, naming the role and the reason.

**Grants converge after migrations**, not only in `db-init`: a function created by a migration in this
very boot would otherwise stay ungranted until the next restart.

## How to test it (10 minutes)

1. `docker compose -f docker-compose.dev.yml up` — the stack should start normally. The backend log
   mentions verifying the runtime role.
2. **The refusal is the interesting half.** In `.env` set `DATABASE_USER=postgres` (a superuser) and
   restart the backend. It should refuse to start, with a message naming the role and why. Restore
   `.env` afterwards.

If you run more than one replica, starting them together is worth a look: previously both could run
migrations at once; now one waits.

## Risk

**This is the only PR in the series that can stop a container from starting** — that is its job, but it
means an unsafe existing deployment will notice. If your `DATABASE_USER` is a superuser or owns the
schema today, this PR will refuse to boot until that is changed. That is the finding, not a side
effect; the message tells the operator what to fix.

## Relationship to the merged #1063

#1063 landed while this series was being split, so this branch is rebased on top of it and the two are
reconciled rather than competing. They check different things and both are needed:

- **#1063** verifies the attributes of the role it *provisions* -- not SUPERUSER, not BYPASSRLS, and so
  on -- and narrows the blanket grant so the runtime cannot write the migration ledger.
- **this PR** verifies the role the application actually *connects as* cannot reach that power by
  another route: membership in a privileged role, or `SET ROLE` without inheritance. Asking
  `pg_has_role(..., 'USAGE')` covers both, which is why it subsumes the inheritance-only check rather
  than replacing one gap with another.

Two conflicts were resolved toward #1063, because they were the same fix written twice and its wording
is the one already reviewed: the `CREATE ROLE` attribute list, and the migration-ledger revoke.

The second is a real difference and worth naming. #1063 revokes `INSERT, UPDATE, DELETE` and keeps
`SELECT`; this branch had revoked `ALL`. The narrower version stands — read access to a list of
migration filenames is not what the revoke protects against; forging or deleting a row is. Two tests
here asserted the wider revoke and now assert the merged one, with the reason written beside them.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity) -- no user-facing strings.

## AI assistance disclosure

---


kenlasko (changes requested):
## Review

**Scope note:** this PR is stacked on #1075, so the combined diff carries six commits. I reviewed only this PR's own work, commit `82f4157` (6 files, +517/−1). The other five belong to #1075.

**Verified locally** at `82f4157`: `npx tsc --noEmit` clean, `eslint` clean on all four source files, and `app-role.spec.ts` / `db-migrate.spec.ts` / `startup-logging.spec.ts` / `lint-bans.spec.ts` all green (86 tests).

### Overview

Three changes: a session-scoped advisory lock serialising `db-init` and `db-migrate` across replicas; a new `assertRuntimeRoleIsSafe` that refuses enforce-mode boot on an over-privileged runtime role; and a second application of the role grants after migrations.

The advisory lock is the strongest part — correct mechanism, right scope choice (session, so it releases on connection close with no unlock to forget), and it genuinely fixes the crash-loop-during-rollout it describes.

---

## Blocking findings

### 1. The new role check duplicates an existing one and contradicts it

`backend/src/common/db/runtime-role-check.ts` already exists on `main` and is wired into `main.ts:86`. Its `assertRuntimeRoleSafe` refuses enforce-mode boot on: SUPERUSER, BYPASSRLS, **REPLICATION, CREATEROLE, CREATEDB**, database ownership, owned policied tables, inherited owner roles, transitively SET-reachable unsafe contexts, and membership in ten forbidden predefined roles — with a `KNOWN_SAFE_PREDEFINED_ROLES` list and an integration guard that forces a decision on any new `pg_*` role.

The new `assertRuntimeRoleIsSafe` (`app-role.ts:271`) is strictly weaker and **disagrees on three attributes**: it downgrades CREATEDB / CREATEROLE / REPLICATION from refusal to `logger.warn`, with the comment "these grant no RLS exemption" — while the existing check refuses on exactly those, citing RR5-001 (a REPLICATION role can hold WAL). Two checks in the same directory, with names differing by one word (`assertRuntimeRoleIsSafe` vs `assertRuntimeRoleSafe`), giving opposite answers on the same input.

Either call the existing checker from `db-init` (parameterised by role name rather than `current_user`), or state in the code why a second, weaker pre-flight earns its keep — and at minimum stop contradicting the stricter one.

### 2. The commit message and PR body claim coverage the SQL does not have

> "The verifier now classifies every predefined role and refuses at startup" — commit message
> "membership in a predefined role with host capability is a third route" — PR body

`RUNTIME_ROLE_SAFETY_SQL` tests only `m.rolsuper OR m.rolbypassrls`. `pg_execute_server_program`, `pg_read_server_files` and `pg_write_server_files` are neither superuser nor BYPASSRLS, so membership in them passes the new check. The predefined-role classification the message describes lives in the *existing* `runtime-role-check.ts` — not in this diff.

This is the same defect class that commit `fa3ec47` in this very stack fixes elsewhere ("a comment asserting a guard the code lacks").

### 3. `applyAppRoleGrants`'s stated reason does not exist in the tree

`app-role.ts:357` justifies the second grant pass by naming "the SECURITY DEFINER `currency_code_in_use_globally`, whose EXECUTE migration 136 revokes from PUBLIC." Checked:

- `database/migrations/` ends at `135_import_jobs_single_active.sql` — there is no 136.
- `currency_code_in_use_globally` appears **nowhere** in the repo except that comment.
- `APP_ROLE_GRANTS_SQL` contains no `GRANT EXECUTE ON FUNCTION` at all — only tables and sequences.

The commit message compounds it: "The currency-function grant inside `APP_ROLE_GRANTS_SQL` is guarded by an existence check." There is no currency grant inside it.

So as shipped, the second pass cannot converge any function grant. The table case is thin too: `db-init` runs `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ... TO %I`, which already covers tables the migrator creates afterwards as the same owner. Either land the guarded function grant here, or reword to the reason that is actually true today.

### 4. The convergence run in `db-migrate` has no failure signal

`APP_ROLE_GRANTS_SQL` reports privilege failure via `RAISE WARNING` inside `EXCEPTION WHEN insufficient_privilege`, which reaches the client as a Postgres notice. `db-init.ts:59` registers `client.on("notice", ...)` and logs it. **`db-migrate.ts` registers no notice handler** — node-postgres drops unhandled `notice` events silently.

So the new call site's only diagnostic is discarded: a grant convergence that fails produces zero log output, on the exact path added to make grants converge. Add the same listener to `db-migrate`'s client.

---

## Non-blocking

**The lock is not "held across both phases."** `advisory-locks.ts` says "A session-scoped advisory lock held across both phases makes one process the migrator." `docker-entrypoint.sh` runs `node dist/db-init.js` and `node dist/db-migrate.js` as separate processes with separate connections, so it is taken and released twice. The serialisation is still **correct** — each phase's check-then-act happens entirely inside its own hold, and a follower re-reads after the winner commits — so this is a comment defect, not a bug. Flagging it because the repo's rule is that such a comment is itself the defect.

**No test asserts the lock is taken.** `db-migrate.spec.ts` adds one `mockResolvedValueOnce({ rows: [{}] })` per test plus a comment, nothing more. Deleting `acquireDbLifecycleLock` from `db-migrate.ts` would fail these tests only incidentally, via mock-offset drift, and the failure would read as an unrelated shape error rather than "the lock is gone." One explicit assertion fixes it:

```ts
expect(mockQuery).toHaveBeenNthCalledWith(1, "SELECT pg_advisory_lock($1)", [DB_LIFECYCLE_LOCK_KEY]);
```

**Two new units ship untested.** `advisory-locks.ts` has no spec file, and `applyAppRoleGrants` has no test — it appears in `src/` only at its definition and its one call site. `db-init.ts` has no spec either, so the enforce-only gating and the ordering (after provisioning, before the tables-exist early return) are unverified. Project rule: "Always create tests for any new functionality added."

**JSDoc attached to the wrong declaration.** `advisory-locks.ts:44-52` documents "Block until this session holds the database-lifecycle lock" — but sits directly above `const logger = new Logger("DbLifecycleLock")` on line 52, not above `acquireDbLifecycleLock` on line 54. Hovering the function shows nothing. Move the logger declaration above the doc block.

**The `MEMBER` vs `USAGE` comment argues against a phantom.** `app-role.ts` spends ~25 lines explaining that the predicate "was `'USAGE'`" and is now `'MEMBER'`. The query is new in this commit — there is no prior version. The `'USAGE'` being argued against is the neighbouring `runtime-role-check.ts`, which uses `'SET'` and `'USAGE'` as two deliberate questions and documents why at length (RR3-001, RR4-001). A reader who follows the argument to that file finds the opposite case made, with measured evidence.

**`owns_tables` is trivially false on a fresh database.** In `db-init` the check runs before `schema.sql` is applied, so on the boot that creates the schema there are no tables to own. Harmless given `main.ts` re-checks later — but it means the pre-flight is weakest on exactly the boot it is meant to guard, which is another argument for finding 1.

**No `lock_timeout`.** `pg_advisory_lock` blocks indefinitely, so a wedged migrator now hangs every replica's startup instead of crash-looping. That is the deliberate trade and the "Waiting for the database lifecycle lock..." log line makes it diagnosable — but a generous `SET lock_timeout` would turn a stuck lock into a named error rather than a pod that never goes ready.

---

## What's done well

Parameterised queries throughout; `format('%I')` identifier quoting preserved in the DDL. The `problems` array reports every reason at once rather than one per restart, and each message names the remedy — the "warns but starts" vs "refuses" distinction is explicitly tested. The advisory-lock rationale for session scope over transaction scope is correct and well argued. The test asserting the revoke is ordered *after* the blanket grant is a good catch.

## Recommendation

Findings 1–4 are all "the code does less than the prose says," which is precisely the defect class this audit series is about — and #4 is a live operational gap. The advisory lock itself is ready to merge.

---

WMP:
Thanks — all four blocking findings are addressed in 65d3a46, along with every non-blocking note. The branch is also rebased onto `main` now that #1075 has merged, so the PR shows only its own two commits and the review-scope caveat no longer applies.

## Blocking findings

**1. Duplicate and conflicting role check.** You were right that this was the core defect rather than a wording problem: `assertRuntimeRoleIsSafe` and its `RUNTIME_ROLE_SAFETY_SQL` are deleted from `app-role.ts` outright. `runtime-role-check.ts` now builds its facts query from one template instantiated for two subjects — `current_user` for the serving connection in `main.ts`, `$1` for the pre-flight — and `db-init` calls the new `assertRuntimeRoleSafeByName`, which runs the same `runtimeRoleViolations` list and throws through the same message builder. CREATEDB / CREATEROLE / REPLICATION are refusals again on both surfaces.

Two guard tests, because prose here has already failed once:
- `runtime-role-check.spec.ts` asserts `NAMED_ROLE_FACTS_SQL === RUNTIME_ROLE_FACTS_SQL.replace("r.rolname = current_user", "r.rolname = $1")` — any edit to one query that is not an edit to the template fails it.
- An `it.each` over ten unsafe shapes (each forbidden attribute, database ownership, policied-table ownership, inherited owner, forbidden predefined-role membership, reachable context) asserts *both* asserts reject the same input.

**2. Unsubstantiated coverage claim.** Once the pre-flight shares the classifier, the predefined-role classification is genuinely true of it — a `pg_execute_server_program` membership now stops the boot in `db-init`, not only in `main.ts`. The commit message is rewritten to describe what the diff does rather than what the audit branch did.

**3. Referenced code that doesn't exist.** The `currency_code_in_use_globally` / "migration 136" citation is gone. The real present-tense reason for the second grants pass is `schema_migrations`: on a first boot `db-init` grants before `schema.sql` has created the table, so the write-revoke finds nothing and the default privileges then hand the runtime role INSERT/UPDATE/DELETE on the ledger the moment it appears. The `db-migrate` pass is what takes those writes back. Both comments now say that and nothing more.

**4. Silent failure in db-migrate.** `db-migrate.ts` registers the same `notice` handler as `db-init.ts`, with a test that fails if it goes missing.

## Non-blocking notes

- **Lock comment.** Rewritten: the two scripts are separate processes on separate connections, so nothing holds the lock across both phases. Each takes the same key for its own run, and what makes that sufficient is that each re-reads the database's state *after* acquiring rather than trusting a decision made before the wait.
- **Untested lock acquisition** → new `advisory-locks.spec.ts` (100% of the file): the key and its parameterization, the waiting line ordered before the blocking call, the timeout reset ordered after it, the 55P03 path, and a non-timeout error passing through unwrapped.
- **Missing test coverage** → `applyAppRoleGrants` has a suite in `app-role.spec.ts` (including "never touches the role or its password"), and `db-init.ts` has `db-init.spec.ts`. `db-init.ts` is exported behind a `require.main` guard like `db-migrate.ts` and is off the `collectCoverageFrom` exclusion list, so it counts toward the global thresholds now.
- **Misplaced JSDoc** → moved onto the function.
- **Phantom `USAGE` argument** → deleted with the function whose comment it was.
- **Weak pre-flight timing** → the `enforce` check now runs after the schema work, so the ownership arms interrogate the database as it will actually be served rather than an empty one. The test asserts the order of the emitted queries; I verified it fails when the call is moved back.
- **No lock timeout** → `lock_timeout` is set to 600s around the acquisition and `RESET` once the lock is held, so a migration's own lock waits keep the server's configured behaviour. A 55P03 becomes a named error pointing at `pg_locks` and the key, rather than a startup that hangs with only the waiting line in the log. 600s is a judgement call: long enough that a large legitimate migration run never trips it, short enough that a wedged holder surfaces within a restart cycle.

## Two things outside the findings

The RLS plan and runbook both asserted "no session-scoped advisory locks anywhere in the backend (verified)". This PR made that false, and a document making a false claim about the source is the failure mode this repo has a guard for. Both now scope the constraint to the pooled runtime connections and name the startup scripts as the deliberate exception — including the consequence, that they must reach Postgres directly rather than through a transaction-mode pooler.

And per the "a defect an AI wrote is a missing rule" convention, `backend/CLAUDE.md` gains a short section naming `runtime-role-check.ts` as the single owner of the role-safety question, with the two guard tests as the enforcement.

## Verification

`npm run typecheck` and ESLint clean; `TZ=UTC npm run test:cov` green at 418 suites / 11172 tests with the global coverage thresholds met (11171 → 11172 plus the new suites; the count moved because `db-init.ts` joined the measured set).

---

Issue #1073, opened 2026-08-05
03: Three restore-path design risks: unbounded queue, pre-auth upload occupation, unmeasured memory constants (DR-F3RB-002..004)


WMP:
Three design risks in the restore path, grouped because each needs a product or measurement decision
rather than a bug fix, and because they interact: the memory constants decide the gate's capacity, the
gate decides who queues, and the queue is what an unauthenticated upload can occupy.

**DR-F3RB-002 — the restore wait queue is unbounded and not cancellation-aware.**
`restore-processing-gate.ts` stores waiters in a plain array of resolve callbacks: no queue limit, no
timeout, no removal when the request disconnects, no `AbortSignal`. A client can disconnect while
queued and its destructive restore still runs when a slot frees. Decide between fail-fast admission
(429/503 once the queue is full) and a bounded abort-aware queue.

**DR-F3RB-003 — pre-authentication upload occupation is bounded, not eliminated.**
The aggregate upload admission must run before body parsing, which is before Nest's guards, so an
unauthenticated request can hold part of the budget until the receive deadline expires — degrading a
legitimate restore to a retry. The concurrency and lifecycle halves are fixed; what remains needs
ingress-level limits, a short-lived authenticated upload token, or streaming to disk.

**DR-F3RB-004 — `PEAK_MULTIPLE` and the process baseline are estimates, not measurements.**
They are argued from allocation counting. **Correction to the original text of this issue, which said
the processing cap "is designed to survive them being wrong": it is not, and the arithmetic shows the
opposite.** The gate bounds concurrency only. Every ceiling in the chain is derived by dividing by
`PEAK_MULTIPLE`, so none of them can vouch for it: `safeDerivedUploadLimit` is literally
`container * share / PEAK_MULTIPLE`, and `computeRestoreProcessingSlots` divides by
`PEAK_MULTIPLE * expandedLimit`. On the default 400 MiB pod that yields expanded limit 100 MiB,
modeled peak 300 MiB, baseline 96 MiB — one slot with 4 MiB spare, and a *true* multiple above **3.04**
puts a single admitted restore over the container (3.20 on a 512 MiB or 1 GiB pod). So a legitimate
restore can still be OOM-killed at one slot, and zero-slot refusal protects only according to the
model. Settling this needs a cgroup-constrained peak-RSS test over representative payloads (encrypted
and plain, several compression ratios, concurrent ordinary traffic), and possibly a safety margin on
top of whatever it measures. Closing this would also let
[...]

Issue #1071, opened 2026-08-05
03: Destructive restore accepts any non-empty OIDC value as the second factor (F3RB-007)


WMP:
## What's wrong

Restore replaces the caller's entire financial dataset, and for an OIDC account its second factor
is a string the client chooses. The frontend sends the literal `oidc-session-confirmed`; the backend
checks `if (!input.oidcIdToken) reject` and nothing else. No signature, no expiry, no audience, no
purpose, no nonce, no single use, no `auth_time` freshness, no provider round trip. Any non-empty
value — `x` — satisfies it.

So possession of a valid application session is the whole authorisation. The extra confirmation the
UI and the API name both imply does not exist. This is the same class of defect as the step-up
service's OIDC branch, which accepts `oidcConfirmed?: boolean` — also a flag the client asserts.

## The mechanism already exists, in another series

The fix is not to invent anything here. Audit-02's PR #1060 ("OIDC re-authentication and OAuth
grant-integrity hardening") introduces `OidcReauthService`: a five-minute artifact signed with
`JWT_SECRET`, bound to the user, the action and a one-time id, mintable only by the OIDC callback
after a real `prompt=login` / `max_age=0` round trip and bound to the `state` of that specific
authorisation request. Once it lands, closing this is:

1. add a `backup-restore` purpose to `OIDC_REAUTH_PURPOSES`;
2. consume the artifact in the restore path (signature, type, subject, action, freshness, single
   use) and delete the sentinel from the DTO and the frontend;
3. do the same for the step-up service's `oidcConfirmed` branch;
4. solve the file-selection problem the redirect creates — the user has a large backup file
   selected, and the OIDC round trip loses it, so validate everything free *before* spending the
   artifact (PR #1060 establishes exactly that ordering rule for the local-password path).

Implementing a second minting/verification path on the audit-03 branch instead would put two
divergent implementations of cryptographic step-up in one repository and guarantee a conflict with
#1060. Hence: blocked on that PR by choice, not overlooked.

## Where

- `backend/src/backup/backup.service.ts` — `verifyAuthentication`, the `oidcIdToken` branch.
- `backend/src/backup/backup.controller.ts` — the `X-Restore-OIDC-Token` header.
- `backend/src/auth/step-up/step-up.service.ts` — the sibling `oidcConfirmed` branch.
- `frontend/src/components/settings/BackupRestoreSection.tsx`, `frontend/src/lib/backupApi.ts` — the
  sentinel.
[...]

Issue #1070, opened 2026-08-05
03: Export materialises whole tables and every attachment before the memory ceiling applies (F3RB-006)


WMP:
## What's wrong

The plain HTTP export is described as streaming, and it does stream *between* tables — but not
within one. Each table arrives as a complete array from `manager.query()` and is serialised with a
single `JSON.stringify(rows)`; every external attachment object is loaded and base64-encoded into
one `carried` array before serialisation begins; and the attachment/blob arrays stay live until the
export ends. So peak memory tracks total dataset size, not chunk size. The plain path also has no
ceiling of its own — the buffered ceiling deliberately does not apply to it.

Thirty 10 MiB attachments are ~300 MiB raw, ~400 MiB as base64 text, before counting source
buffers, row objects, the JSON strings, gzip buffers and ordinary process memory. The canonical
chart sizes the backend at 400 MiB. The users with the most to lose — the largest datasets — are
the ones who hit it, and the failure mode is an OOM kill with no artifact and no readable error,
while the repeatable-read snapshot and its pooled connection are held until the process dies.

## Why it is an issue rather than part of the audit-03 series

The fix is one cursor inside the repeatable-read snapshot: serialise rows in batches,
base64-encode one attachment at a time, apply a per-chunk budget, honour backpressure. For the
encrypted path it additionally needs a streaming authenticated-encryption container, because GCM
currently requires the whole payload up front to compute its tag. That is a feature-sized change to
the most sensitive file in the backup module, it collides with every other PR touching
`backup.service.ts`, and its main claim — bounded peak RSS — cannot be verified without the
cgroup peak-RSS harness this repository does not yet have (`DR-F3R6-002` / `DR-F3R7-003`).

Doing it properly would also let `PEAK_MULTIPLE` be replaced by a measured bound instead of an
allocation-counting estimate, which is why `docs/audit-phase-3-response.md` §7 item 3 calls it
**the single highest-value open item**.

## Where

- `backend/src/backup/backup.service.ts` — `streamExport` (the table loop and `JSON.stringify`),
  `collectGzippedExport`, `readTable`, `appendExternalAttachmentBytes`.
- Ceilings and sizing: `backend/src/backup/backup-limits.ts`, `helm/values.yaml`.
- Contract: `docs/backup-restore-contract.md`; response doc §7 item 3 and §14.

## Suggested verification

Not a unit test: run a large plain export and a large encrypted export through the real HTTP
[...]

Issue #1069, opened 2026-08-05
03: A partial automatic backup can replace a complete one and consume its retention slot (F3RB-001)


WMP:
## What's wrong

`AutoBackupService.exportToFile` always builds the ordinary date-only name
(`monize-backup-daily-<YYYY-MM-DD>.json.gz`, or `.mzbe` when encrypted) and publishes it with
`writeFileAtomic`, which deliberately replaces an existing file of that name. Completeness is only
consulted afterwards, in `applyBackupOutcome`: an incomplete artifact skips promotion and retention
**for its own run**, but it has already been written under a complete artifact's filename and it
matches `DAILY_FILE_PATTERN`, so every later retention pass counts it as a complete daily backup.

Two consequences, both losing recovery points a configured policy promises:

- **Same-day replacement.** With `every6hours`, a healthy 02:00 run writes a complete artifact; an
  08:00 run that cannot read one attachment writes a partial one onto the same path. The complete
  artifact is gone before `applyBackupOutcome` records `partial`. `Run Backup Now` after a
  successful scheduled backup does the same.
- **Retention counts partials as complete.** With `retentionDaily = 3`: days 1–3 complete, days 4–6
  partial (retention skipped on those runs), day 7 complete and runs retention over seven
  ordinary-looking daily names — it keeps day 7 plus partial days 6 and 5, and deletes complete
  days 1–3. Three complete recovery points become one.

Sharding by user (`<root>/<ab>/<cd>/<userId>/`) isolates *users*; this is a collision between the
complete and partial artifacts of the *same* user, so it is unaffected.

Completeness also does not survive the artifact: it lives in the settings row
(`lastBackupStatus`), not in the file or a catalog, so a copied-away artifact, a restart, or a
directory rescan cannot tell the two apart.

## Why it is not simply a bug fix

The plausible fix changes artifact naming (a distinct durable identity for partial artifacts, e.g.
`monize-backup-partial-<timestamp>.json.gz`) and adds completeness to the envelope or a sidecar
catalog, with retention enumerating only artifacts proven complete. That is a compatibility
decision about the on-disk format and about what existing ordinary-named partial artifacts mean, so
it wants a maintainer call rather than being folded into the audit-03 backup PR.

## Where

- `backend/src/backup/auto-backup.service.ts` — `exportToFile` (filename + `writeFileAtomic`),
  `applyBackupOutcome` (post-publication completeness), `enforceRetention` / `DAILY_FILE_PATTERN`.
[...]

PR #1077, opened 2026-08-05
03: Harden manual backup and restore: integrity, self-contained artifacts, bounded memory (v1.15.0)

WMP:
> **Stacked on #1076, which is still open.**
> diff until then.

**Merge order: 3 of 4**, after PR 2 (#1076). Targets `main`.

**Rebased onto the seven audit-02 PRs that merged while this was being split** (#1057–#1063). Three
collisions were reconciled rather than resolved by preference, and each is worth a reviewer's eye:
#1061's atomic write, #1060's OIDC step-up ordering, and #1061's mock-based filesystem tests. The
sections below say what won and why.

This is the large one. It is large **because the file it centres on cannot be cut up**: the final
`backup.service.ts` imports the memory limits, the processing gate, and the completeness report it
returns, so splitting it into separate PRs produces branches that compile while mixing concerns. I
tried the split three ways before concluding that. Per-commit history for everything here is on
`claude/detailed-error-review-4pbug7` if you would rather review it that way.

## What it fixes

**Restore ordering is declared, not remembered.** `restore-plan.ts` holds the insertion order and the
deferred foreign keys as data, and a spec proves it against `database/schema.sql`. A new FK between
backed-up tables now fails a test instead of failing a user's restore.

**Currency references are decided once.** The list of columns pointing at `currencies(code)` was
written out in four places and wrong in all four. Migrations 136/137 make it a SQL function the
database evaluates. Before this, a user with a budget in a custom currency was told the code was
unused, lost the activation row — and because that preference cascades, **other users' activations went
with it**.

**A backup carries its own attachment bytes, for every provider.** Previously `local` and `s3`
attachments were metadata only, so a restore onto a fresh instance produced rows pointing at nothing.
Bytes now travel base64 in the artifact, verified against size and SHA-256 at export and again at
restore, and staged before anything is deleted — so an unusable object is discovered while the user's
data is still there. A backup taken under one provider restores under another.

**An export cannot claim more than it holds.** Completeness is assessed before the first byte is
written; an incomplete artifact says so in its filename and headers instead of reporting success.
Missing and checksum-mismatched attachments are counted separately, because "could not find it" and "it
changed under us" are different answers.

**Ownership comes from the database.** A restore reads a legacy external object only after confirming
the caller already owns a matching `transaction_attachments` row. No id, key, size or checksum from the
uploaded file establishes entitlement.

**#1061's atomic write is subsumed, not duplicated.** It wrote to a temp name and renamed — atomic
against a crash *of the process*, but not against power loss: without an `fsync` before the rename, the
final name can point at data that never reached the disk. `atomic-file.ts` flushes the file, renames,
then flushes the directory entry, and cleans up stale temp files. Going the other way, #1061 found
something this branch had missed — a pid-based temp name is not unique across replicas sharing a
volume — so its `randomUUID` now lives in `atomic-file.ts` with the reasoning. If this merges, #1061 is
fully contained here.

**Memory is bounded.** A small gzip expands to a large payload, so the wire limit never saw the real
cost: four 1 MiB uploads could hold ~400 MiB decompressed on a 400 MiB pod. Upload admission now
budgets across the process, decompression has a ceiling, and processing capacity is computed from the
resolved expanded limit minus a process baseline. A container where one modeled restore does not fit
refuses with an actionable 503 rather than being OOM-killed mid-recovery.

## How to test it (20 minutes — this is the PR worth clicking through)

Use a test account, or take a database backup first: **restore is destructive.**

1. Attach a file to any transaction (Transactions → pick one → Attachments).
2. Settings → Backup & Restore → **Export**.
3. Unpack the `.json.gz` and look for `attachment_blobs` — it should contain data, not just metadata.
   That is the change: previously local/S3 bytes stayed behind.
4. **Restore that file** on the same account. Then open the transaction and download the attachment —
   it should open.
5. The honesty check: delete the attachment's file from disk or the bucket **behind the application's
   back**, then Export again. The download should be named `...-INCOMPLETE...` and the UI should say how
   many attachments are unusable. Before this, the export reported success.
6. The currency check: Settings → Currencies → add one (e.g. `XTS`); create a budget in it; try to
   delete the currency. It should refuse and say it is in use.

Config-level, if you want to see the limits work: set `BACKUP_RESTORE_LIMIT=1mb`, restart, and try to
upload a larger backup — you should get a clear refusal, not a hang or a restarted container.

## Risk

High, in the sense that it touches how backups are produced and consumed. Everything here is covered by
417 backend suites (11203 tests), but the cases that matter most — a real S3 bucket, a real fresh-instance
restore, a container under a memory limit — are not in that number. Worth exercising once in a staging
deployment before this goes to users.

## What this PR does NOT fix

Stated plainly, because each has an issue and a comment in the code saying so:

- **#1069** — a partial automatic artifact still has no distinct durable identity, so a same-date partial
  can occupy a complete artifact's retention slot. This PR prevents *promotion* of an incomplete outcome;
  it does not solve naming. Do not read it as retention being solved.
- **#1070** — the plain export still materialises whole tables and every attachment before the memory
  ceiling applies.
- **#1071 is no longer open, and this PR must not be read as reverting it.** PR #1060 merged while this
  series was being split, so `main` now consumes a real single-use, purpose-bound OIDC artifact for a
  destructive restore. This branch keeps that. The rebase reconciled the one place the two designs
  collided: #1060 validates the uploaded file *before* spending the artifact (it is single-use, and the
  round trip that mints it loses the file selection), while this work put the whole restore inside the
  processing gate. The gate stays outside and #1060's ordering inside, so nothing decompresses
  unbudgeted and no artifact is spent on a file that was never going to restore.
- **#1073** — the memory model's constants are argued, not measured. On the default 400 MiB pod the
  margin is 4 MiB, and a true multiple above 3.04 still overflows. The gate bounds concurrency; it does
  not prove one restore fits.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] This PR addresses a **single concern** (large work is split into a series). -- one concern, but a large diff: one concern, one inseparable file -- see the note at the top
- [x] All user-facing strings are translated for **every** locale (i18n parity) -- every backend message this PR adds is in all 18 locales here, because the parity spec would fail this PR otherwise; the frontend catalogs travel with PR 4 for the same reason.

## AI assistance disclosure

---


kenlasko (commented):
Reviewed against this PR's own four commits (`5ec08e61`, `bd7ab07d`, `9112b451`, `35f31999`) rather than the combined diff, per the note at the top.

The engineering is strong and the reasoning carried in the comments is unusually good. Three things need doing before merge, and there are two real defects in the new code.

## Must fix before merge

**1. The branch conflicts with `main`, and one resolution is dangerous.**

#1076 merged while this sat. A test merge conflicts in `backend/src/common/db/advisory-locks.ts` (add/add), `app-role.ts`, `app-role.spec.ts`, `db-init.ts`, `db-migrate.ts`, `db-migrate.spec.ts` and `database/CLAUDE.md`. This branch carries `82f41570` + `e468107b`; `main` took the further-evolved pair (`315f8614` + `65d3a46d`, "Give the startup role pre-flight the runtime check's classifier").

Keeping this branch's side on `app-role.ts` would reintroduce the hand-written role-safety query that #1076 deleted — the exact defect the new "One classifier decides whether a database role is safe" section of `backend/CLAUDE.md` was written about. Drop this branch's two copies and rebase the four backup commits onto `main`.

**2. `database/CLAUDE.md` on `main` now contradicts migration 136.**

Post-#1076 it reads: *"No shipped migration uses this yet (the tree holds no `SECURITY DEFINER` function, and every grant lives in `backend/src/common/db/app-role.ts`)"*. `136_currency_global_liveness.sql` ships both a `SECURITY DEFINER` function and a `REVOKE ... FROM PUBLIC`. That file is one of the conflicts above, so the merged text has to be this PR's version, not `main`'s.

**3. "Verify PR checklist" is the only red check** — the linked-discussion box is unchecked and the link line under "Linked discussion / issue" is empty. Everything else is green (unit, integration, E2E x4, lint, typecheck, schema drift, Bearer, Lighthouse).

## Defects

**4. A cancelled plain export leaks the `REPEATABLE READ` snapshot transaction.**

`backend/src/backup/backup.service.ts:363-370`:

```ts
const write = (chunk: string): Promise<void> =>
  new Promise((resolve, _reject) => {
    if (!gzip.write(chunk)) { gzip.once("drain", resolve); } else { resolve(); }
  });
```

If `gzip.write` returns `false` and the client then aborts the download, nothing drains the stream and nothing rejects — the promise never settles, so `inExportSnapshot`'s callback never returns and the transaction opened at `backup.service.ts:522` stays open for the life of the process.

This is newly consequential *because* of this PR. The comment at `:496-515` records that each table used to get its own short transaction; now an aborted download pins a pooled connection and holds an xmin against vacuum indefinitely. Cancelling a large backup download is the likeliest way a user reaches this path.

Reject pending writes on `res`/`gzip` `close` and `error`, or use the `gzip.write(chunk, cb)` callback form with an error listener attached.

**5. `loadOwnedAttachmentSources` is handed every UUID in the backup.**

`backend/src/backup/backup.service.ts:1696`:

```ts
const ownedSources = await this.loadOwnedAttachmentSources(userId, [...idRemap.keys()]);
```

`idRemap` holds every row id in the artifact — transactions, splits, prices, all of it. The only ids ever looked up are `oldIdOf.get(attachmentId)` for the attachment rows, so this builds a potentially several-hundred-thousand-element array out of an uploaded file and serialises it into one `id = ANY($2::uuid[])`.

It also runs unconditionally, including when every attachment carried its own bytes — the case in which the legacy path it feeds is never taken at all.

Narrow it to `rows.map(r => oldIdOf.get(String(r.id ?? "")))`, and skip the query entirely when no row needs the legacy path. It is the one place in the PR where an unbounded, upload-sized allocation slipped past the memory-bounding theme.

## Smaller points

**6. Stale identifier claims in the new code.** Worth calling out because commit `8648045` in this very PR adds the guard test for exactly this class of defect — and that guard only scans `CLAUDE.md` and `docs/*.md`, so none of these are caught:

- `backend/src/currencies/currency-reference-columns.ts:4` — "the two SQL functions (migrations 133 and 134)". They are 136 and 137; 133/134 are the joint-accounts migrations.
- Same file, `:10` — points at `currency-reference-columns.spec.ts`. The file is `currency-references.spec.ts`.
- `database/migrations/137_currency_codes_referenced_by_user.sql:43` — "unlike their sibling in 133" should read 136.
- `database/migrations/136_currency_global_liveness.sql:69` — "the runtime grant lives in `backend/src/db-init.ts`". It lives in `backend/src/common/db/app-role.ts:171`, as commit `35f31999` itself says.

The consistent move would be extending `doc-paths.spec.ts` to backticked paths in `database/migrations/*.sql` and `src/**/*.ts` comments — this PR is otherwise a model of "prefer the rule the machine can check".

**7. `RestoreProcessingGate.configure(0)` strands queued waiters.** `drain()`'s loop cannot run at capacity 0, and `acquire` only throws for *new* callers. Unreachable today — only `main.ts` configures, once, before traffic — but the class is exported and spec'd, and rejecting existing waiters there would make the zero case uniformly honest.

**8. `backup.service.ts` is 2544 lines against the 800-line ceiling in `CLAUDE.md`.** The argument in the description — that the file cannot be split across *PRs* — is sound, and I agree with it. It does not argue that the file cannot be split *within* this PR: export assembly, restore orchestration, attachment staging and the teardown are four separable concerns. A follow-up issue would be enough.

**9. Test step 5 over-promises.** "the UI should say how many attachments are unusable" — there are no frontend changes in this PR (the description says the catalogs travel with PR 4), so a reviewer clicking through gets the `-INCOMPLETE` filename and the `X-Backup-*` headers, and no UI copy.

## What is clearly right

`restore-plan.ts`-as-data plus the schema-parsing spec is the right shape for the `linked_loan_account_id` class of bug — the obligation becomes a property of three arrays instead of something you hold in your head while reading forty call sites.

Moving the currency reference list into SQL functions, and making the global one `SECURITY DEFINER`, correctly fixes a real cross-tenant data-loss path: an RLS-blinded `NOT EXISTS` feeding an `ON DELETE CASCADE` is a hard bug to see and a very bad one to have. The `FOR UPDATE` on the parent row closes the check-then-delete race *through* the FK's `FOR KEY SHARE` conflict rather than around it, which is the right mechanism and is explained as such.

The ownership argument in `stageAttachmentObjects` — that no unsigned value from an uploaded file can establish entitlement, so the bytes themselves have to travel — is correct, and keeping both documented wrong turns in the comment is worth the lines.

`writeFully` handling short writes, and `attachments.service.ts` moving external byte deletion to after the commit, are both real bugs found and fixed the right way round.

---

WMP:
@kenlasko : Do you know why checks dont start?

kenlasko:
Github is having issues with actions right now: https://www.githubstatus.com/

WMP:
So, github is down, I've run out of tokens before get /code-review output, time to go to bed...

kenlasko:
The AI Gods have spoken.

PR #1078, opened 2026-08-05
03: Tell the user what a backup actually contains, in every language (v1.15.0)

WMP:
> **Stacked on #1077, which is still open.**
> diff until then.

**Merge order: 4 of 4**, after PR 3 (#1077) -- it reads completeness headers that PR 3's backend sets, so it
must not merge first. Targets `main`.

## What this is

The frontend half of the backup hardening, plus the localization for both halves.

**The export used to report success whichever bytes it managed to gather.** It now reads the
completeness headers, marks an incomplete download in its filename, and says how many attachments are
unusable — counting missing and checksum-mismatched together, because "0 of 1 missing" for a corrupted
attachment reads as nothing wrong. An absent header means complete, so an older backend degrades to the
previous behaviour rather than to a false warning.

**Automatic backups say there is no storage configured before you set a schedule**, rather than after
the first failed run. And a schedule that has started failing can now be switched off — unavailable
storage used to block disabling it too, which left the user with a warning and no way to act on it.

**Restore results name skipped attachments** instead of folding them into the restored count: a number
of things you did not do does not belong in the total of things you did.

All copy is in all 18 locales. That is not politeness — the parity tests require a key to exist
everywhere in the commit that introduces it, so a PR adding English only is red.

## How to test it (10 minutes, all in the UI)

1. Settings → Backup & Restore → **Export**. Normal case: file downloads, success message.
2. Delete an attachment's file from storage behind the app's back, then Export again. The filename
   should carry an INCOMPLETE marker and the message should count the unusable attachments — not report
   success.
3. Restore a backup that is missing an attachment. The result should list skipped attachments
   separately from restored rows.
4. Settings → Automatic Backups with no storage path set: the warning should be there **before** you
   configure anything. Set a schedule, then make storage unavailable — you should still be able to turn
   the schedule off.
5. Switch the interface language (Settings → Preferences) and repeat step 2 — the new messages should be
   translated, not English fallbacks.

## Risk

Low. No backend behaviour changes here; the worst case is wording.

## What changed about #1071 while this was in flight

The earlier draft of this description said the OIDC confirmation shown before a destructive restore is
not a real second factor. That was true of the branch and is **no longer true of `main`**: PR #1060
merged, so the backend now consumes a signed, single-use, purpose-bound artifact. Nothing in this PR
weakens it. Worth saying explicitly because the opposite claim was written down first, and a stale
"known defect" note is how a fixed thing gets re-reported.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity) -- all 18 locales, both frontend and backend catalogs.

## AI assistance disclosure

---


kenlasko:
## Review

Checked out `pr/03-04-backup-ui` at `8d57c16` and reviewed it against `main` at `8ebca94` (which already contains #1077, so the diff narrows to this PR's own work).

### Verification

| Check | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npx eslint .` | clean (1 pre-existing warning in `Combobox.tsx`) |
| `npm run i18n:check` | clean |
| `npx vitest run` (full frontend) | **648 files / 12642 tests passed** |
| `npx jest src/backup/auto-backup.service.spec.ts` | 98 passed, 2 skipped |

**The backend spec fix is more urgent than the description suggests.** I ran `main`'s copy of `auto-backup.service.spec.ts` in a separate worktree today (2026-08-07, a `WEEKLY_DAY`) and it **fails 10 tests**. `main` is red right now, and the second commit here is what makes it green. That commit is independently mergeable and probably shouldn't wait on the rest of this PR.

I also confirmed the header channel works end to end: `frontend/src/proxy.ts:182-188` copies every response header except `transfer-encoding`, so the `X-Backup-*` markers survive the same-origin proxy.

Contract compliance looks right — `skippedAttachments` sits beside `restored` rather than inside it, an absent header degrades to "complete" rather than to a false alarm, and missing + checksum-mismatched are counted together so a corrupt attachment doesn't read as "0 of 1".

### Findings

Findings 1 and 2 were confirmed empirically with a throwaway probe test against this branch (deleted afterwards), not just read off the diff.

**1. (medium) Capability is read once and never refreshed, so validating a writable alternate folder leaves the toggle dead.** With `available: false` for the stored path: type `/mnt/nas/backups`, click Validate, get "Folder is valid and writable" on screen — the enable toggle is *still* `disabled`, and the banner still names the old path. `capability` is only re-read by `loadSettings`, which runs on mount and after Run Now — and Run Now is itself disabled in exactly this state (`AutoBackupSection.tsx:616`).

The backend's `describeCapability` probes the *stored* folder specifically to support a writable secondary root from `BACKUP_ALLOWED_ROOTS` (its own comment cites F3RB-003), so this is a supported configuration that the new UI now blocks. Recovery exists but is undiscoverable: save the folder without arming, reload the page, then arm. Re-fetching capability after a successful `validateFolder` and after a successful save would close it.

This is the same shape as the defect the PR set out to remove — a warning with no way to act on it.

**2. (medium) A failed capability read disables the toggle with no on-screen reason.** When `getAutoBackupCapability` rejects, the toggle is `disabled` and no banner renders — the guard at `AutoBackupSection.tsx:258` gates only on `storageUnavailable`, not on `capabilityLoadFailed`. The sole signal is a transient toast, so after a reload the user sees a greyed-out control and nothing explaining it.

It's also the opposite of the degrade policy chosen one file over for the export headers, which deliberately fails permissive on a missing signal. Either render a banner for the `capabilityLoadFailed` branch, or fail open.

**3. (low) `Number(header(...) ?? 0)` yields `NaN` for a malformed header** (`backupApi.ts:229-233`), and the toast then renders "NaN of NaN attachment(s)". A finite check costs one line.

**4. (low) `X-Backup-Attachments-Included` is sent by the backend and ignored here.** `missing + inconsistent` is correct only while those two are the exhaustive failure set; `expected - included` is the same number today and stays right if a third category is ever added.

**5. (low) Reformat noise.** `backupApi.ts` and `AutoBackupSection.tsx` are converted single→double quotes and re-wrapped wholesale, while `BackupRestoreSection.tsx` in the same directory keeps single quotes. There's no prettier config and no quote lint rule in this repo, and the frontend convention is overwhelmingly single quotes (787 files vs 60). This roughly doubles the reviewable diff and leaves the settings directory mixed style. Worth reverting the cosmetic churn.

**6. (nit) Adjacent hardcoded English on lines this PR rewrites:** `return "Never"` (`AutoBackupSection.tsx:38`) and the browse panel's `Cancel` button (`AutoBackupSection.tsx:334`). Both pre-existing, both cheap while you're in the file.

### Summary

The substance is good: the completeness headers are read correctly, the disable-an-armed-schedule case is handled via persisted rather than form state, the i18n is complete across all locales including the pseudo-locale, and the new tests target the specific behaviours rather than the calls. Findings 1 and 2 are the two I'd want addressed before merge — both leave a user looking at a disabled control with no route forward, which is the failure mode this PR exists to eliminate.

---

PR #1096, opened 2026-08-07
04: Make balance and transaction writes concurrency-safe (v1.15.0)

WMP:
> **Stacked on `pr/04-01-concurrency-foundations`, which is itself stacked on #1078.**
> This branch is exactly one commit above Phase 4 PR 1 (`187b555236fdc3d577dab7f397867994736763e2`).
> Until the lower PRs land, GitHub's combined diff will contain their commits too. Review the unique
> commit `fb20158618888a0c5f3c9f42c7d14c26344f62f3` for this PR's own work.
>
> After the prerequisite PRs merge, rebase this branch onto the resulting `main` before final review.

**Phase 4 merge order: 2**, after `pr/04-01-concurrency-foundations`. Targets `main`.

## What this fixes

This PR applies the Phase 4 lock and transaction contract to the core financial writers. The goal is not to change financial formulas; it is to ensure that the value committed to derived state is based on the row version the write actually replaces.

### Transaction update and delete use locked committed state

Transaction edit/delete paths now lock and re-read the transaction before deriving an old amount, account, date or status from it.

Previously two requests could both act from the same stale snapshot. For example, with an opening balance of `100.00` and one `-10.00` transaction, concurrent edits to `-20.00` and `-30.00` could both reverse the original `-10.00`; the ledger could finish at `-30.00` while the stored balance finished at `60.00` instead of `70.00`.

The write paths now derive reversals from the `FOR UPDATE` row that is actually being replaced.

Deletion follows the same rule. `removeLockedTransactionLeg` conditionally deletes the locked row and reverses its balance contribution only when that delete actually removed a row. A second concurrent delete therefore cannot reverse the same amount twice.

### Absolute balance recomputation composes with atomic deltas

`AccountsService.recalculateCurrentBalance` now locks the account row **before** reading the ledger and writing an absolute total.

That closes the `READ COMMITTED` window where:

1. the recomputation reads the ledger;
2. another request commits an ordinary balance delta;
3. the recomputation writes an older absolute total and erases the committed delta.

The same account-lock protocol is used by other absolute derived-state writers added in this PR, including action-history balance rebuilds and net-worth snapshot recomputation.

### Closed accounts cannot receive a late balance mutation

`AccountsService.updateBalance` now performs the arithmetic and the `is_closed = false` check in the same guarded SQL `UPDATE`.

A transaction mutation that waited behind `close()` can no longer read the account as open and then add a delta after the close committed. A zero-row result is distinguished inside the same transaction as "account gone" versus "account closed", and the exception rolls the surrounding ledger mutation back.

### Holdings updates share one account-level exclusion

Holdings are derived from investment transactions, so a holdings-row lock alone cannot serialize a new trade against a rebuild that deletes and recreates holdings.

All holdings read-modify-write and rebuild paths now use the shared holdings advisory-lock namespace before reading the state they will replace. That covers ordinary quantity/cost-basis updates, stock splits, share adjustments, deletion, rebuilds and account-wide wipes.

### Transfers, splits, bulk operations and reconciliation use locked rows

The same stale-read rule is applied to the surrounding transaction machinery:

- transfer updates/removals lock both legs and compare against the committed pair;
- split parent/leg operations use the parent-before-leg ordering and include regression coverage for the prior deadlock shape;
- split-linked transaction removal uses the single conditional delete-and-reverse path;
- bulk status updates lock every affected transaction before deriving VOID balance movements;
- bulk deletes lock primary and linked rows before deriving reversals and count rows actually deleted;
- reconciliation transitions resolve their guards and VOID balance effects against the locked current status, not a pre-transaction snapshot.

### Net-worth snapshot repair is recoverable

Net-worth recalculation now locks the account before reading the source ledger and replacing snapshots. The process-memory debounce remains a latency optimization, but a periodic stale-snapshot sweep derives missed work from durable database timestamps so a killed pod or failed timer does not leave stale snapshots indefinitely.

### Existing account-service coverage is preserved

The earlier split of this work accidentally deleted a large portion of `accounts.service.spec.ts`. This branch keeps the existing behavior suite and adds the new concurrency cases instead; the file's diff is additive overall rather than replacing the suite with a small subset.

## User-visible errors and localization

New refusal/error messages are carried through the backend i18n catalogs for all supported translated locales, and the pseudo-locale is regenerated. The localization change is included here so this behavioral PR can satisfy the repository's merge-time parity rule independently of later Phase 4 PRs.

## Why

These races are financially material because the authoritative ledger and its derived values could disagree even though every individual request returned success. The failures were not floating-point noise: they were whole committed deltas, duplicated reversals or stale holdings state.

The required invariant is:

> A read-modify-write or delete-and-reverse operation must derive its side effects from the row version it owns under the same transaction and lock as the mutation.

This PR applies that invariant across the financial write paths that share balances, holdings and transaction state.

## Review focus

The highest-value review points are:

1. every reversal comes from a locked current transaction row, not a caller snapshot;
2. a delete reverses only when that request actually deleted the row;
3. absolute balance/snapshot recomputation locks before its source read;
4. closed-account refusal is part of the balance write, not a preceding check;
5. multi-row locks are deterministic and split parent/leg ordering is consistent;
6. holdings writers share one advisory-lock namespace before read-modify-write;
7. rejected operations roll back the ledger mutation and all derived-state changes together;
8. existing account/transaction behavior tests remain present alongside the new concurrency coverage.

## Test evidence

The branch commit records the following as green on its Phase 4 PR 1 base:

- backend typecheck;
- backend lint;
- balance/transaction unit specs: **1,860 tests**;
- backend i18n parity: **212 tests**;
- transfer integration spec: **15 tests**.

The branch also contains explicit integration regression coverage for the split parent/counterpart deadlock scenario.

The PR has not yet produced GitHub Actions status checks because it has not been opened. The numbers above are therefore the branch author's reported local verification and should be confirmed by PR CI before merge.

## Risk

**High.** This PR deliberately touches the core transaction, balance, holdings and net-worth write paths. The changes are defensive and transactional rather than formula changes, but a mistake in this layer can affect stored balances or portfolio state. The focused concurrency tests and separation from the later cron/import/attachment work are intended to keep that risk reviewable.

## What this PR does not do

- no database migration or schema change;
- no change to tax, gain, FX or rounding formulas beyond preserving existing monetary precision at the write boundary;
- no Phase 4 cron/job-claim protocol;
- no MNY import lifecycle changes;
- no attachment lifecycle changes;
- no emergency-access changes.

Those concerns remain in later Phase 4 PRs.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] This PR addresses one concern: concurrency safety of core financial writes and their derived state.
- [x] New behavior has regression tests; the branch author reports the test sets above as green.
- [x] New backend user-facing errors are localized and the pseudo-locale is regenerated.
- [ ] Shared/core-area scope has maintainer sign-off in the linked discussion or issue.
- [ ] The branch is rebased on the latest `main`. This becomes actionable after the prerequisite stacked PRs merge.

## AI assistance disclosure

Implemented with Claude Code from findings produced by the Phase 4 repository audit. The branch was subsequently reviewed for stack ancestry, transaction boundaries, stale-read/double-reversal failure modes, lock ordering, holdings exclusion, reconciliation/bulk behavior, preservation of the existing account-service test suite and localization completeness before this PR description was prepared.

### DIFF
--- a/backend/src/accounts/accounts.service.spec.ts
+++ b/backend/src/accounts/accounts.service.spec.ts
@@ -360,12 +360,25 @@ describe("AccountsService", () => {
   });
 
   describe("updateBalance", () => {
+    /**
+     * The guarded UPDATE reports whether it matched a row.
+     *
+     * `is_closed = false` is a predicate of the write, not of a read that came
+     * before it: a separate check let a mutation of an existing transaction read
+     * the account as open, wait behind `close()`, and then add its delta to a row
+     * that had since been closed at zero (audit P4-008). So the double has to
+     * answer the way the statement does -- a row, or nothing.
+     */
+    function stageBalanceUpdate(matched: boolean): void {
+      mockQueryRunner.query.mockImplementation(async (sql: string) =>
+        String(sql).includes("UPDATE accounts") && matched
+          ? [[{ id: "account-1" }], 1]
+          : [[], 0],
+      );
+    }
+
     it("adds positive amount to balance", async () => {
-      accountsRepository.findOne.mockResolvedValue({
-        ...mockAccount,
-        currentBalance: 1000,
-      });
-      accountsRepository.query.mockResolvedValue(undefined);
+      stageBalanceUpdate(true);
       accountsRepository.findOneOrFail.mockResolvedValue({
         ...mockAccount,
         currentBalance: 1500,
@@ -373,19 +386,16 @@ describe("AccountsService", () => {
 
       const result = await service.updateBalance("account-1", 500);
 
-      expect(mockQueryRunner.query).toHaveBeenCalledWith(
-        `UPDATE accounts SET current_balance = ROUND(CAST(current_balance AS numeric) + $1, 4) WHERE id = $2`,
-        [500, "account-1"],
-      );
+      const [sql, params] = mockQueryRunner.query.mock.calls[0];
+      expect(sql).toContain("current_balance AS numeric) + $1");
+      expect(sql).toContain("is_closed = false");
+      expect(sql).toContain("RETURNING");
+      expect(params).toEqual([500, "account-1"]);
       expect(result.currentBalance).toBe(1500);
     });
 
     it("subtracts negative amount from balance", async () => {
-      accountsRepository.findOne.mockResolvedValue({
-        ...mockAccount,
-        currentBalance: 1000,
-      });
-      accountsRepository.query.mockResolvedValue(undefined);
+      stageBalanceUpdate(true);
       accountsRepository.findOneOrFail.mockResolvedValue({
         ...mockAccount,
         currentBalance: 700,
@@ -393,14 +403,15 @@ describe("AccountsService", () => {
 
       const result = await service.updateBalance("account-1", -300);
 
-      expect(mockQueryRunner.query).toHaveBeenCalledWith(
-        `UPDATE accounts SET current_balance = ROUND(CAST(current_balance AS numeric) + $1, 4) WHERE id = $2`,
-        [-300, "account-1"],
-      );
+      expect(mockQueryRunner.query.mock.calls[0][1]).toEqual([
+        -300,
+        "account-1",
+      ]);
       expect(result.currentBalance).toBe(700);
     });
 
     it("throws NotFoundException when account not found", async () => {
+      stageBalanceUpdate(false);
       accountsRepository.findOne.mockResolvedValue(null);
 
       await expect(service.updateBalance("nonexistent", 100)).rejects.toThrow(
@@ -409,6 +420,9 @@ describe("AccountsService", () => {
     });
 
     it("throws BadRequestException for closed accounts", async () => {
+      // No row matched and the account exists: it is closed. Told apart so the
+      // caller gets 400 rather than one ambiguous error.
+      stageBalanceUpdate(false);
       accountsRepository.findOne.mockResolvedValue({
         ...mockAccount,
         isClosed: true,
@@ -419,12 +433,26 @@ describe("AccountsService", () => {
       );
     });
 
-    it("rounds to 4 decimal places to match DB schema precision", async () => {
+    it("refuses the delta when close() committed while the ledger write waited", async () => {
+      // The regression guard for P4-008: an unvoid read the account as open, the
+      // close committed at zero, and the delta then landed on the closed row --
+      // a closed account holding -10.00. Re-evaluating the predicate after the
+      // row lock refuses instead, and the throw rolls the ledger mutation back
+      // with it.
+      stageBalanceUpdate(false);
       accountsRepository.findOne.mockResolvedValue({
         ...mockAccount,
-        currentBalance: 10.1,
+        isClosed: true,
       });
-      accountsRepository.query.mockResolvedValue(undefined);
+
+      await expect(
+        service.updateBalance("account-1", -10),
+      ).rejects.toBeInstanceOf(BadRequestException);
+      expect(accountsRepository.findOneOrFail).not.toHaveBeenCalled();
+    });
+
+    it("rounds to 4 decimal places to match DB schema precision", async () => {
+      stageBalanceUpdate(true);
       accountsRepository.findOneOrFail.mockResolvedValue({
         ...mockAccount,
         currentBalance: 20.3,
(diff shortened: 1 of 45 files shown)


kenlasko (changes requested):
Reviewed this PR's own commit (`fb20158`) against its stated base, and ran the backend suites locally.

**The core work is sound.** The invariant it sets out to enforce — derive every reversal from the row version the write actually replaces, under the same lock — is applied correctly and consistently across transfers, splits, bulk, reconciliation and holdings. Two changes are better than the description claims: folding `is_closed = false` into the guarded `UPDATE` in `updateBalance`, and switching `recalculateCurrentBalance` from `repo.save(account)` to a targeted one-column `UPDATE` — the old form wrote back every column from a pre-transaction snapshot, so a balance recalculation would silently revert a concurrent rename or opening-balance edit. The advisory-before-row ordering rule in `locks.ts` is the right shape.

Findings below, ordered by severity.

## 1. CI is red — two suites fail (verified locally)

Full unit run on this branch: **2 failed / 429 passed suites, 2 failed / 11,583 passed tests**. Both failures are this PR's, so the "green on this base" evidence in the description does not hold.

- **`backend/src/common/cron-doc.spec.ts`** — `sweepStaleSnapshots` adds `@Cron(CronExpression.EVERY_30_MINUTES)` at `backend/src/net-worth/net-worth.service.ts:201`, and `docs/cron-jobs.md` has no `net-worth.service` row. The guard reports `undocumented: ["net-worth.service"]`. The service's own comment cites `docs/cron-jobs.md` for the idempotent-predicate mechanism, which makes the missing row the more surprising.
- **`backend/src/accounts/rls-context-smoke.spec.ts:66`** — the new `lockAccountsForBalanceWrite` inside `applyDueTransactionBalances` issues an extra `m.query`, shifting that spec's fixed `mockResolvedValueOnce` chain by one. The balances query now receives `undefined` → `Cannot read properties of undefined (reading 'length')`. `accounts.service.spec.ts` was updated for the extra query; this one was missed.

Typecheck, lint and i18n parity (249 tests, all locales) are green — the localization pass is genuinely complete.

## 2. A lock-order inversion of exactly the shape this PR set out to fix

`locks.ts` states the rule as *parent before legs, by role not by id*, and `removeTransferFromSplitInTransaction` (`transaction-transfer.service.ts:920`) implements it carefully. `TransactionsService.remove` does the opposite:

- `transactions.service.ts:2272` locks the leg (`id`) first, then `removeParentTransaction` locks the parent at `:2335` → **leg → parent**
- `transaction-split.service.ts:712` (`removeSplit`) locks the parent, then `removeLinkedLeg` (`:690`) locks the leg → **parent → leg**

Both are live endpoints (`DELETE /transactions/:id` at `transactions.controller.ts:1358`, `DELETE /transactions/:id/splits/:splitId` at `:1506`), and `remove()` explicitly handles the split-counterpart case via its `parentSplit` lookup — so this is a reachable `40P01` on the same two rows, in opposite orders.

The new integration regression covers `removeSplit` vs `removeTransfer`, which was already correct, and not this pair. The fix is to lock the parent before the leg in `remove()`, and to extend `transfers.integration.spec.ts` to the third pairing.

## 3. Two comments make claims the source does not support

Both are the failure mode `CLAUDE.md` names: a doc that names an identifier is making a claim about the source.

- `remove-transaction-leg.ts:26` says "`derived-state-writers.guard.spec.ts` fails on a balance reversal beside an unconditional `remove()`". **That file does not exist anywhere in the repo.** Given the module's own premise is that this sequence was open-coded four times and one copy drifted, the guard is the part most worth actually writing — and it is a source-scanning test, the highest rung of the "prefer the rule a machine can check" ladder.
- The commit message and description call `removeLockedTransactionLeg` "the single sanctioned delete-and-reverse path". It is not: `transactions.service.ts` (`remove`, plus two copies inside `removeParentTransaction`), `investment-transactions.service.ts` (`removeLinkedCashTransaction`) and `transaction-bulk-update.service.ts` all still hand-roll it. And they disagree — `removeLockedTransactionLeg` calls `recalculateCurrentBalance` for a future-dated row, while `removeLinkedCashTransaction` silently skips it. That divergence is precisely what the helper was introduced to prevent.

## 4. The net-worth sweep's premise is false for one edit class

`net-worth.service.ts:189` grounds the whole derived-staleness mechanism on: *"Every write that can change a snapshot also moves the account's `current_balance` in the same transaction."*

It does not. In `transactions.service.ts:2199-2226`, a past→past date change with unchanged amount and account takes no branch — no `updateBalance`, no `recalculateCurrentBalance`, so no write to `accounts` and no `updated_at` bump. But moving a transaction between months does change `monthly_account_balances`. If the in-memory debounce is lost — the exact crash the sweep exists to recover from — that account can never be found by `a.updated_at > s.computed_at + grace`.

Either widen the predicate, or narrow the comment to the classes it actually covers. The mechanism is a good idea; it is the unqualified claim that will be built on later and be wrong.

## 5. Minor

- `transactions.service.ts:2234` and `:2314` still trigger the net-worth debounce on the **pre-transaction snapshot's** `accountId`, while the balance work now correctly uses `locked.accountId`. If a concurrent edit moved the row, the account that actually lost the amount never gets its snapshots recomputed. Small, and the debounce is best-effort, but it is an inconsistency this PR introduced by fixing one of the two and not the other.
- `holdings.service.ts` `remove()` re-reads the holding with `findOne({ where: { id } })`, dropping the `userId` scoping the outer `findOne(userId, id)` had. Safe today — ownership was checked and cannot transfer, and `RLS_MODE=off` is the default — but it is the unscoped-read pattern the codebase bans elsewhere.

## Verdict

Requesting changes. Items 1 and 2 are blocking: 1 is mechanical, 2 is a real liveness bug in the area the PR is about. Items 3 and 4 are corrections to claims rather than to behaviour, but on a high-risk PR in this layer a comment asserting a guard exists is worse than no comment at all. I would also ask that item 3's guard test actually be written, since the "one sanctioned path" property is currently enforced by nothing.

---

PR #1097, opened 2026-08-07
04: Harden auth counters, refresh-session revocation, and 2FA preference writes (v1.15.0)

WMP:
> **Stacked on #1096 (`pr/04-02-balance-transaction-safety`).**
> This branch is exactly one commit above #1096's current head (`fb20158618888a0c5f3c9f42c7d14c26344f62f3`).
> Until the lower PRs land, GitHub's combined diff will contain their commits too. Review the unique
> commit `754d4866aeeb63e29441e8a83b8efbae9a8627b5` for this PR's own work.
>
> After the prerequisite PRs merge, rebase this branch onto the resulting `main` and rerun the full
> CI suite before final review.

**Phase 4 merge order: 3**, after #1096. Targets `main`.

## Scope of the unique diff

The unique commit changes **8 files** (`+665/-144`) and is limited to authentication/session concurrency plus the preference-column writer used by 2FA:

- `backend/src/auth/auth.service.ts`
- `backend/src/auth/auth.service.spec.ts`
- `backend/src/auth/token.service.ts`
- `backend/src/auth/token.service.spec.ts`
- `backend/src/auth/two-factor.service.ts`
- `backend/src/auth/two-factor.service.spec.ts`
- `backend/src/users/user-preference-writer.ts`
- `backend/src/test-helpers/user-preference-testing.ts`

There is no migration, schema change, locale/catalog change, emergency-access change, or `auth-email.service.ts` change in this commit.

### Refresh-token rotation and family revocation use one lock protocol

This PR closes the Phase 4 logout-versus-refresh race (`P4-011`) by making refresh rotation and family revocation coordinate on the same transaction-scoped family advisory lock.

`refreshTokens()` now:

1. performs an unlocked lookup only to learn the immutable `family_id`;
2. acquires `lockTokenFamily()` **before** any refresh-token row lock;
3. re-reads the presented token under `FOR UPDATE` after the family lock is held;
4. performs replay detection, revocation and replacement insertion while that family remains excluded.

`revokeTokenFamily()` acquires the same family lock before revoking family members. `revokeRefreshToken()` keeps token lookup and family revocation inside one `withScopedDb` scope; the nested call is intentionally re-entrant and therefore joins the same transaction/manager supplied by the Phase 4 concurrency foundations.

This removes the prior `READ COMMITTED` statement-snapshot window where logout could update the old token after a concurrent rotation committed yet fail to see the replacement inserted by that rotation.

For whole-user session invalidation, `revokeAllUserRefreshTokens()` discovers active families, locks them in deterministic ascending order, revokes active rows, and performs bounded additional passes rather than relying on a single statement snapshot.

### Failed-login increments use the value PostgreSQL committed

The old login path calculated:

```text
user.failedLoginAttempts + 1
```

from an entity read before the bcrypt comparison. Five parallel invalid passwords could therefore all read `0` and all write `1`, allowing the persistent counter to lag far behind the actual attempt rate.

The new path executes:

```sql
UPDATE users
   SET failed_login_attempts = failed_login_attempts + 1
 WHERE id = $1
 RETURNING failed_login_attempts
```

and derives the lockout threshold from the returned committed count. The successful-login reset is also issued as a guarded SQL update rather than being skipped because of a stale entity snapshot.

This fixes the lost-increment half of `P4-012` and ensures the threshold decision no longer uses the pre-bcrypt read.

### 2FA preference writes no longer save a stale whole preferences entity

`user_preferences` is one row written by unrelated settings. A read/mutate/`save(entity)` cycle can write stale values from that entity back over columns another request changed in the meantime.

This PR introduces `patchUserPreferences()` and `ensureUserPreferencesRow()`:

- the row is materialized with `INSERT ... ON CONFLICT DO NOTHING` when absent;
- only the columns present in the requested patch are updated;
- concurrent changes to unrelated preference columns are left untouched.

2FA confirmation and disablement now use that writer for `twoFactorEnabled` rather than saving the whole preferences entity. The tests explicitly cover both a missing preferences row and preservation of an unrelated existing preference.

## Important pre-merge concurrency issue

**The current head should remain draft until this is corrected.**

The failed-login counter increment is atomic, but the `lockedUntil` write is committed in a **separate scoped transaction after the increment returns**. That leaves a serializability gap against a successful login reset.

One realistic interleaving is:

1. the account has four failed attempts;
2. request A verifies a wrong password, increments `4 -> 5`, receives committed count `5`, and commits that increment;
3. request B verifies the correct password and resets `failed_login_attempts = 0, locked_until = NULL`;
4. request A then performs its separate lockout update and sets `locked_until` into the future.

The final row can therefore be:

```text
failed_login_attempts = 0
locked_until          = <future timestamp>
```

That state is consistent with neither serial order: if A happened before B, B should have cleared the lock; if B happened before A, A should have incremented from zero and should not have locked the account.

The same split transaction can also let a lower-count parallel failure overwrite a longer progressive lockout chosen by a later committed count.

Before this PR is marked ready, the counter increment and any resulting lockout update should be one row-serialized database unit: either one guarded SQL statement/CTE or one `withScopedDb` transaction that retains the row lock from the arithmetic increment through the `locked_until` decision. The successful-login reset should compete with that whole unit, so the final state always corresponds to one valid serial order.

A regression test should race a valid login against the threshold-crossing invalid login through independent PostgreSQL connections and assert that no final state has `failed_login_attempts = 0` with a future `locked_until`. Parallel threshold/progressive-lockout tests should also prove that lockout duration cannot move backwards.

## Scope correction after the rebase

The commit message mentions password-reset/email-change token consumption. Those changes are **not part of this branch's unique diff** on the current stack.

`AuthEmailService.resetPassword()` and email-verification consumption are already conditional `UPDATE ... WHERE` operations in the parent tree, and `backend/src/auth/auth-email.service.ts` has the same blob on this branch and on #1096's head. This PR therefore does not claim those changes.

Emergency-access claim hardening is also intentionally absent here; it belongs to the later emergency-access PR in the Phase 4 series.

## Why

Authentication state has the same concurrency rule as financial derived state: a security decision must be based on the row version the write actually owns, not on an application snapshot that can become stale while expensive work or another request runs.

The key invariants for this PR are:

> After logout returns, a refresh-token replacement created by a racing rotation in the same family must not remain usable.

> Failed-password accounting must reflect committed attempts, and lockout/reset state must correspond to a valid serial order.

> Toggling a security preference must not revert unrelated preference columns written concurrently.

## Review focus

The highest-value review points are:

1. `lockTokenFamily()` is acquired before refresh-token row locks everywhere rotation/revocation interact;
2. `revokeRefreshToken()` really stays in one re-entrant scoped transaction rather than opening a second connection;
3. after a logout/refresh race, no active token remains in the presented family;
4. multi-family acquisition in `revokeAllUserRefreshTokens()` is deterministic;
5. failed-login threshold decisions use the count returned by SQL, not the pre-bcrypt entity value;
6. the pre-merge `lockedUntil` race described above is closed before requesting final review;
7. 2FA changes update only `twoFactorEnabled` and materialize a missing preferences row safely;
8. no password-reset, email-verification, emergency-access, migration, schema or localization behavior is accidentally attributed to this unique commit.

## Test evidence

The branch commit records the following as green on its #1096 parent:

- backend typecheck;
- backend lint;
- focused auth/session unit specs: **419 tests**.

The unique tests cover:

- SQL-based failed-login counting and threshold decisions;
- guarded successful-login counter reset;
- family-lock ordering for refresh rotation and revocation;
- deterministic multi-family session revocation and bounded retry passes;
- 2FA preference-row materialization;
- 2FA targeted-column updates that preserve unrelated preference values.

There is currently **no GitHub Actions run for commit `754d4866aeeb63e29441e8a83b8efbae9a8627b5`**, so the test count above is branch-author-reported local evidence, not CI evidence.

The current tests are primarily unit-level lock/protocol simulations. Before merge, the security-sensitive concurrency invariants should also be exercised against real PostgreSQL with independent connections, especially:

- logout racing refresh rotation in one family (`P4-011`);
- threshold-crossing invalid login racing a valid login reset (`P4-012`).

## Risk

**High until the remaining `P4-012` lockout transaction gap is fixed; Medium after that correction.**

The diff is focused and has no schema migration, but it changes authentication lockout and refresh-session revocation behavior. A mistake can either leave a session usable after logout or deny access to a legitimate user, so the final rebased tree needs full CI plus real-database concurrency coverage before merge.

## What this PR does not do

- no database migration or `database/schema.sql` change;
- no new user-facing strings or locale/catalog changes;
- no password-reset/email-verification consumption change unique to this commit;
- no emergency-access claim/grant hardening;
- no financial balance/transaction changes beyond those inherited from #1096;
- no cron/job-claim protocol;
- no MNY import lifecycle changes;
- no attachment lifecycle changes.

Those concerns are either already in the parent tree or belong to other Phase 4 PRs.

## Linked discussion / issue

Approved in: 

## Checklist

- [x] This PR addresses one concern: concurrency safety of authentication/session state and the 2FA preference write it depends on.
- [x] Focused regression tests are present; the branch author reports **419** auth/session unit tests as green.
- [x] No new user-facing strings are introduced, so no localization change is required.
- [ ] Close the remaining `P4-012` increment/`lockedUntil` transaction gap described above.
- [ ] Add or run real-PostgreSQL race coverage for the security-sensitive interleavings.
- [ ] Full GitHub Actions CI is green on the final rebased tree.
- [ ] Shared/core auth scope has maintainer sign-off in the linked discussion or issue.
- [ ] The branch is rebased on the latest `main` after the prerequisite stacked PRs merge.

## AI assistance disclosure

Implemented with Claude Code from findings produced by the Phase 4 repository audit. The branch was subsequently reviewed read-only against its exact parent and current head for stack ancestry, unique-diff scope, refresh-family lock ordering, nested scoped-transaction behavior, failed-login counter semantics, 2FA preference-write isolation, test coverage and stale claims carried over from the earlier PR handoff. The remaining `P4-012` transaction gap above was identified during that review and must be resolved before the PR is considered merge-ready.


kenlasko (commented):
## Review: `04-03: Auth and session hardening`

Reviewed the unique commit `a5231a1` only (8 files, +665/-144) against its stacked parent `e355221`, not the combined diff.

**Verified locally on the branch:** `tsc --noEmit` clean; `eslint` clean on all changed files; `TZ=UTC npx jest src/auth/auth.service.spec.ts src/auth/token.service.spec.ts src/auth/two-factor.service.spec.ts` -> 3 suites, 245 tests passing; `src/users` + `src/updates` -> 16 suites, 240 tests passing (no collateral breakage from the preference-writer change).

The core direction is right, and the family-lock ordering in `token.service.ts` is correct as far as I can trace it: `refreshTokens` takes family-then-row, `revokeAllUserRefreshTokens` takes families-then-rows in ascending order, and `revokeRefreshToken`'s nested `revokeTokenFamily` genuinely joins the ambient scoped transaction rather than opening a second one. No lock-order cycle between those three paths.

Findings below, most severe first.

---

### 1. Blocking: the increment and the `lockedUntil` write are still two transactions

`backend/src/auth/auth.service.ts` (~L425-465)

The PR body identifies this and says *"The current head should remain draft until this is corrected"* — but the PR is marked ready for review, not draft, and the checklist item is unticked. Restating it here so it is a review finding rather than a self-note, plus two details the body does not mention.

`incrementFailedAttempts()` runs its own `withScopedDb` and **commits**. The `locked_until` write is a separate `scoped(User, ...)` transaction issued afterwards. Two interleavings produce states that match no serial order:

- **A legitimate user is locked out despite a correct password.** Request A increments 4->5 and commits; request B then verifies the correct password and resets `failed_login_attempts = 0, locked_until = NULL`; A then writes its `locked_until` into the future. Final row: `failed_login_attempts = 0` with a future `locked_until`. B got its tokens, and every subsequent login is a 403 until the lock expires.
- **Lockout duration moves backwards.** With enough parallel failures, the request holding committed count 10 computes `2^1` = double duration, and a request holding count 5 computes `2^0` and can commit last, shortening the lock.

Two things worth noting that the body does not:

- The gap is **not** microseconds. Between the committed increment and the `locked_until` write sit an awaited `resolveUserEmailLocale()` — a whole separate `withScopedDb` round trip — plus `emailTranslator` and `accountLockedTemplate` rendering.
- The `if (Object.keys(updateFields).length > 0)` guard means the sub-threshold path now issues no second statement at all. That part is a genuine improvement; it just makes the threshold path the only one carrying the gap.

Suggested shape: fold the threshold decision into the increment as one guarded statement, e.g.

```sql
UPDATE users
   SET failed_login_attempts = failed_login_attempts + 1,
       locked_until = CASE
         WHEN failed_login_attempts + 1 >= $2
           THEN now() + make_interval(secs => $3 * power(2, floor((failed_login_attempts + 1)::numeric / $2) - 1))
         ELSE locked_until
       END
 WHERE id = $1
 RETURNING failed_login_attempts, locked_until, (SELECT locked_until FROM users WHERE id = $1) AS previous_locked_until
```

or, if the progressive-lockout arithmetic is better kept in TypeScript, one `withScopedDb` that does `SELECT ... FOR UPDATE` and holds that row lock through the `locked_until` decision — with the success-path reset competing for the same row lock so it cannot slot in between.

### 2. The lockout email now fires per failed attempt above the threshold, not per lockout transition

Same block. The `if (newAttempts >= this.MAX_FAILED_ATTEMPTS)` branch sends "Account Temporarily Locked" for **every** committed count at or above 5. Under parallel failures all requests clear the pre-bcrypt `lockedUntil` check before any of them has written a lock, so counts 5, 6, 7... each render and send a mail to the victim's address.

This is a direct consequence of fixing the counter: the old (broken) arithmetic had every parallel request compute `1`, so parallel bursts sent *zero* emails. The per-IP throttle (5 per 15 min on `/auth/login`) bounds a single source to roughly one, but it does not bound a distributed one.

The fix composes with #1: have the guarded statement `RETURNING` the pre-update `locked_until` and send the mail only when the write actually moved the row from not-locked to locked. That makes it exactly one email per lockout, which is what the feature means.

### 3. `user-preference-writer.ts` documents itself as "the one way", and five call sites still do the thing it exists to prevent

`backend/src/users/user-preference-writer.ts:7`

The docblock opens with *"The one way to write a few columns of a user's preferences row"* and names the motivating bug explicitly: *"Dismiss a tour in one tab while switching the theme in another and the theme reverts."*

That exact path is not migrated. Still doing `findOne` -> mutate -> `repo.save(entity)`:

- `backend/src/users/users.service.ts:261`
- `backend/src/updates/tours.service.ts:98`
- `backend/src/updates/updates.service.ts:252`
- `backend/src/updates/whats-new.service.ts:91` and `:120`

Only the two 2FA call sites use the writer. Per the root `CLAUDE.md` — *"A doc that names an identifier is making a claim about the source"* and *"Prefer the rule the machine can check"* — either migrate those five in this PR, or narrow the docblock to what is true today and add a source-scan guard test that fails on `save(<UserPreference>)` outside the writer, so the next agent inherits the correction. Right now nothing stops the sixth instance.

### 4. New shared module ships without its own spec

There is no `backend/src/users/user-preference-writer.spec.ts`. `patchUserPreferences` / `ensureUserPreferencesRow` are exercised only indirectly, through `createUserPreferenceRepoMock` in the 2FA specs. `CLAUDE.md`: *"Always create tests for any new functionality added."*

Specifically untested:
- the empty-patch branch (`Object.keys(patch).length === 0`) that materializes the row and returns — a documented behaviour with no case;
- that `ensureUserPreferencesRow` leaves an existing row untouched (the mock models `ON CONFLICT DO NOTHING` correctly, but nothing asserts the *writer's* contract directly);
- that `buildDefaultPreferences` is the source of the materialized baseline, so a divergence from the registration path fails somewhere.

I did confirm `user_preferences.user_id` is the PRIMARY KEY, so the untargeted `.orIgnore()` is safe — no second unique constraint for it to swallow.

### 5. Every concurrency claim in this PR is proven against a mock

`createScopedDbMocks` gives a jest `EntityManager`; the family advisory lock is asserted as *"a `query()` call whose SQL contains `pg_advisory_xact_lock` was recorded, in this argument order"*, and *"loops until a pass revokes nothing"* is a scripted `mockImplementation` returning a canned family list. None of that can demonstrate that two real connections serialize.

This is the same rule as `CLAUDE.md`'s *"A mocked filesystem cannot demonstrate a filesystem property"* — a mocked `EntityManager` cannot demonstrate an advisory-lock property. The PR body already commits to real-PostgreSQL coverage for `P4-011` and `P4-012` before merge; flagging it so it stays a merge gate. The two that matter most:

- logout racing rotation in one family, asserting no active token remains in the family afterwards;
- threshold-crossing failure racing a successful login, asserting no final state has `failed_login_attempts = 0` alongside a future `locked_until`.

### 6. `revokeAllUserRefreshTokens` can revoke a session created after it started

`backend/src/auth/token.service.ts:297`

Pass 1 essentially always affects > 0 rows (the user has active tokens — that is why we are here), so a second pass always runs. That pass re-runs an **unscoped** `update({ userId, isRevoked: false }, { isRevoked: true })` — not restricted to the families discovered in pass 1 — so a login that lands between the two passes is silently revoked.

`AuthEmailService.resetPassword` (`auth-email.service.ts:157`) calls this immediately before the user is expected to sign in again with their new password, which is exactly the window. Narrow later passes to the family set discovered in pass 1, or gate on `created_at <= <pass-1 start>`, so convergence only ever mops up replacements rather than new sessions.

Separately: the loop's justification in the comment is slightly off. Because all family locks are held *before* the UPDATE and the UPDATE takes a fresh READ COMMITTED statement snapshot, a rotation in an already-discovered family cannot hide from pass 1. The case the loop actually exists for is narrower — a family whose only unrevoked row is a replacement from a rotation that had already revoked its predecessor before the discovery SELECT ran, so the family was never in the lock set. Worth correcting in the comment, since the comment is what the next reader will reason from.

### 7. The commit message claims work that is not in the commit

```
- Password-reset and email-change tokens are consumed via a single conditional
  UPDATE, so exactly one request wins and a token cannot be replayed.
```

`backend/src/auth/auth-email.service.ts` is not in `a5231a1`. The PR body has a "Scope correction after the rebase" section acknowledging this, but the body is not what lands in `main`'s history — the commit message is. Please amend it before merge. The trailing `(P4-003, P4-006, P4-011)` on the 2FA bullet is also mis-attributed relative to the body, which assigns the counter work to P4-012.

### 8. Nit: mixed identifier styles in the reset predicate

`backend/src/auth/auth.service.ts` (~L497)

```ts
.where("id = :id", { id: user.id })
.andWhere("(failed_login_attempts <> 0 OR locked_until IS NOT NULL)")
```

The first line uses the entity property name, the second raw column names. It works — `UpdateQueryBuilder` has no alias and passes unrecognised identifiers through verbatim, and both columns carry explicit `@Column({ name })` mappings — but the two lines now have different failure modes if the entity mapping ever changes. Consider `"(failedLoginAttempts <> 0 OR lockedUntil IS NOT NULL)"` for consistency with the line above it.

---

### Summary

Findings 1 and 2 should block. 3, 4 and 5 are the `CLAUDE.md` obligations this change incurs and are cheap to satisfy now. 6 and 7 are small but land in `main` if not caught here.

The refresh-token family-lock work is the strongest part of the diff and I did not find an ordering defect in it. The failed-login counter is genuinely fixed for the lost-increment half of `P4-012`; it is the half that decides `locked_until` from that count that still needs to become one row-serialized unit.

---

PR #1099, opened 2026-08-07
Make import job reaping demand-driven to end user lockouts

kenlasko:
## Linked discussion / issue

Closes #
Approved in: 

## Summary

Moves import job reaping from a 5-minute cron to the request path, so users whose import workers die are no longer locked out waiting for the schedule to fire. A stale job row blocks all future imports via a partial unique index and cannot be cleared by the client (since `discard` only works on `pending` jobs), making it a lockout rather than litter.

**Demand-driven reaping:** `reapStaleJobsForUser` now runs inside the transactions of the two requests that care:
- `create` (POST /import/mny/start): clears stale jobs before attempting the insert, so the user can immediately retry
- `findOne` (GET /import/mny/jobs/:id): clears stale jobs before reading, so the wizard's poll shows the failure instead of a frozen progress bar

**Advisory check fixed:** `hasActiveJob` now excludes stale jobs from its count, so it agrees with what `create` will actually allow (previously it could throw a 409 for a row the very next statement would reap).

**Cron reduced to hourly backstop:** `reapStaleJobs` still runs cross-user on a schedule, but now hourly instead of every 5 minutes, since the request path handles the waiting user. It catches jobs abandoned by users who closed the tab and never asked again.

**Shared staleness logic:** Extracted `STALE_ACTIVE_JOB_CONDITION` and `reapStatement` so the three callers (`create`, `findOne`, `hasActiveJob`) all use the same predicate—a single source of truth for what counts as stale, preventing subtle drifts that would let one path see a job as active while another reaped it.

## Checklist

- [x] This PR addresses a **single concern** (demand-driven reaping to end lockouts).
- [x] All user-facing strings are translated for **every** locale (no new user-facing strings added).

## AI assistance disclosure

https://claude.ai/code/session_01WHdq5aJFiggMiwJteLYDP7

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -377,3 +377,27 @@ previous backups are encrypted and silently downgrading is worse than failing).
 Cron jobs use `@Cron()` from `@nestjs/schedule` and run **in the API process** -- `ScheduleModule.forRoot()` is registered in `app.module.ts`; there is no separate scheduler process (on k8s with more than one backend replica, every replica fires every cron). For the full schedule, see `docs/cron-jobs.md` or grep `@Cron(`.
 
 Every `@Cron` handler is an out-of-request entry point, so its body must seed its own RLS context (tasks C2-C4): the cross-user fan-out under `withSystemContext`, each per-user body under `withUserContext(userId)`. A handler that reaches the DB with no ambient context throws `DB access outside request/user/system context` in every `RLS_MODE`, including `off` -- the per-module `rls-context-smoke.spec.ts` specs are the pattern for proving a cron runs clean.
+
+### Cleanup somebody is blocked on belongs on the request path
+
+Before choosing an interval, ask what the stale row *does* while it sits there.
+If it is only untidy, a schedule is the whole answer. If it **refuses the user's
+next request** -- a slot, a lock, a uniqueness guard -- then the interval is a
+lockout the user cannot end, and picking a smaller number only makes the outage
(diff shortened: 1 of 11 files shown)


kenlasko:
**Backend Unit Tests** is red on this PR, and it is not this diff. All 10 failures are in `AutoBackupService`, which this PR does not touch.

They reproduce on `main` — I ran `src/backup/auto-backup.service.spec.ts` at `a3d609b8` (current `main` HEAD) and at `8ebca949` (this branch's base): 10 failed, 88 passed, identical set both times.

**Cause: the suite is day-of-month dependent.** `WEEKLY_DAYS = [7, 14, 21, 28]` (`backend/src/backup/auto-backup.service.ts:68`), and `copyToWeeklyIfNeeded` reads the wall clock via `new Date()`. Today is the 7th, so the weekly promotion fires and every assertion expecting exactly one artifact in the folder sees two:

```
Expected length: 1
Received length: 2
Received array:  ["monize-backup-daily-2026-08-07.json.gz", "monize-backup-weekly-2026-08-07.json.gz"]
```

So this job passes on 24 days a month and fails on 4. It will go green again tomorrow on its own and break again on the 14th, 21st and 28th — which is why it has not been noticed. Worth fixing rather than waiting out: a suite that is green because of today's date is not evidence about the code.

Fixing it here would put an unrelated module in this PR's diff, so I am doing it on a separate branch and PR instead. This PR's own suites — the `mny` unit specs and `mny-import-job` / `mny-import` / `rls` integration specs — are green.

---

PR #1111, opened 2026-08-07
04: Coordinate cron jobs with durable claims and owned leases (v1.15.0)

WMP:
> **Draft / stacked Phase 4 PR — not ready for final review yet.**
>
> The current branch head is `e41d66264acc2f6fb09b8ad6b83eb11b78bdaf08`, exactly one commit above
> `a5231a1095cacd89b37e7d3e8375362f38ad7e5c`, a historical `04-03` commit.
> The live `pr/04-03-auth-session-safety` branch has since moved to
> `dac59f589d9bf98aafe0047963c5c9835b3d1fbe`, and upstream `main` has moved to
> `249390cedbc10d2c0c4c167e2611d999cff90b12`.
>
> Phase 3 PR #1078 and Phase 4 PR #1096 are now merged. This branch therefore needs to be
> restacked/rebased onto the current Phase 4 chain and latest `main` before final review. In
> particular, the backup changes need to be reconciled with the newer streaming-export work that
> landed upstream in #1109.
>
> Until that restack is done, review the unique commit
> `e41d66264acc2f6fb09b8ad6b83eb11b78bdaf08`, not the combined branch-vs-`main` diff.

**Phase 4 merge order: 4**, after the auth/session safety PR. Targets `main` after restacking.

## Summary

Every backend replica runs the same NestJS `@Cron()` handlers. Process-local booleans, `Set`s,
`Map`s and read-then-write checks therefore do not coordinate scheduled work across replicas.
This PR introduces database-backed ownership and arbitration for the scheduled and background
work that can otherwise double-send, double-post, race a destructive maintenance operation, or
write stale derived state.

The main mechanisms are:

- a shared `JobClaimService` with permanent claims, expiring leases and per-attempt lease tokens;
- a durable delivery record that distinguishes **"I own this attempt"** from **"the side effect
  actually happened"**;
- database enforcement preventing a stale or previous-release worker from mutating a lease it no
  longer owns;
- durable occurrence/fingerprint keys for work whose uniqueness belongs in the domain table;
- guarded upserts and row locks for cron writers that update shared database state;
- a per-user maintenance protocol for whole-account replacement operations;
- source-scanning regression guards for concurrency mistakes that previously reappeared in several
  unrelated services.

This is intentionally not a single "put a mutex around every cron" change. Each job uses the
smallest durable mechanism that matches its failure semantics: a permanent key, an expiring lease,
a unique domain row, an idempotent predicate, a row lock, or an atomic upsert.

## Why

`ScheduleModule.forRoot()` runs in the API process, so a deployment with two backend replicas
normally executes a scheduled handler twice. Before this series, several paths implicitly assumed
there was only one process:

- bill and mortgage reminders could send the same email from every replica;
- AI insight generation could call the provider twice for the same user and account usage twice;
- two demo resets could interleave a wipe and re-seed and leave a partial demo dataset;
- a scheduled transaction could be posted manually while the cron posted the same occurrence;
- two budget processors could both pass application-level de-duplication and both send an alert;
- market-data refreshes could both observe a missing row and race the same insert;
- long-running whole-account replacement could overlap background work derived from the dataset.

The invariant for this PR is:

> If scheduled/background work is not safe to execute independently on every replica, ownership or
> idempotency must be represented in durable shared state, and the side effect must be derived from
> the state that won that arbitration.

### 1. Durable job claims with owned leases

`backend/src/common/jobs/job-claim.service.ts` is the shared coordination primitive.

It supports three distinct contracts rather than treating every claim as "done":

- `claimOnce(type, userId, key)` — permanent uniqueness for work where the claim itself is the
  durable fact and a second execution must never happen;
- `claimLease(type, userId, key, ttl)` — temporary exclusion for work that may need to be retaken
  after a worker dies;
- `markDelivered(...)` / `wasDelivered(...)` — a delivery record for work with an external side
  effect, written only after that side effect succeeds.

A lease returns a random UUID token identifying **the attempt**, not merely the work. `releaseLease`
and `markDelivered` require that token. This closes the stale-holder race where worker A stalls past
its expiry, worker B legitimately retakes the lease, and A later resumes and deletes or completes B's
lease by addressing only `(claim_type, user_id, claim_key)`.

Migration 141 adds database enforcement for the same property. The current holder declares its token
through a transaction-local `app.job_claim_lease_token` GUC; attempts that try to mutate a live
tokenized lease without owning it are rejected by PostgreSQL. That matters during a rolling deploy,
where a previous-release pod cannot be trusted to know about the new token predicate.

`backend/test/integration/job-claim-lease-rollout.integration.spec.ts` exercises the literal old
untokenized `DELETE` / delivery-update shapes against the migrated database rather than mocking the
property.

### 2. Reminder delivery is recoverable and multi-replica safe

Bill and mortgage reminder jobs now take a per-user lease before sending.

The delivery protocol is deliberately **at-least-once**:

1. take a bounded lease;
2. check whether that exact reminder fingerprint was already delivered;
3. send the email;
4. only after SMTP succeeds, record `delivered_at` using the winning lease token;
5. release the lease on a pre-send refusal or failure.

This fixes two opposite failure modes:

- healthy replicas no longer both send the same reminder;
- a worker killed after claiming but before SMTP no longer leaves a permanent "already handled"
  record that suppresses the reminder forever.

A crash after SMTP acceptance but before `delivered_at` can still cause a retry. That duplicate is an
intentional at-least-once tradeoff: a duplicate reminder is preferable to silently losing the only
renewal/bill warning.

Primary paths:

- `backend/src/notifications/bill-reminder.service.ts`
- `backend/src/accounts/mortgage-reminder.service.ts`
- their focused unit specs.

### 3. AI insights keep the local shortcut but move exclusion to PostgreSQL

`AiInsightsService` still has its in-process `generatingUsers` set because it is a cheap same-process
shortcut, but that set is no longer the concurrency mechanism.

A durable per-user `ai_insight_generation` lease now surrounds aggregate computation, provider calls
and persistence. A killed replica eventually loses the lease by expiry; a normal completion releases
it early by token. This prevents duplicate provider spend and duplicate usage accounting across pods
without replacing the existing recent-insight cooldown.

### 4. Demo reset and intraday generation use different contracts intentionally

The destructive daily demo reset takes an expiring lease. Two wipe-and-reseed passes are not harmless:
one replica can delete rows while the other is still seeding them.

The deterministic intraday generator uses a permanent window claim instead. Every replica computes
the same date/hour window; generating a window twice is worse than losing part of one failed window,
so the claim itself is the durable "this window was consumed" fact.

This distinction is intentional and documented in `DemoResetService`; it is not a single generic retry
policy applied to both jobs.

### 5. Scheduled transaction occurrences become one transactional unit

Posting a scheduled transaction now has a durable occurrence identity:

`(scheduled_transaction_id, original_due_date)`

The posting path locks the schedule, inserts into `scheduled_transaction_postings`, and then performs
the financial mutation, override consumption and schedule advancement in **the same transaction**.

That closes the materially different routes to the same duplicate posting:

- two replicas firing the hourly cron;
- a user manually posting while the cron is running;
- a crash after the financial transaction committed but before `next_due_date` advanced.

The unique occurrence key survives process death and turns the losing poster into a conflict instead
of a second financial transaction.

### 6. Budget period and budget-alert races are moved into database arbitration

Budget-period creation uses the existing `(budget_id, period_start)` uniqueness as the serialization
point through `INSERT ... ON CONFLICT DO NOTHING RETURNING`. The loser reads and returns the period the
winner created rather than surfacing a unique-constraint 500.

Closing a period now locks the open period before deriving its actuals and freezes the period plus its
next-period creation in one transaction. This prevents two monthly cron executions from independently
closing the same snapshot.

Budget alerts keep their application-level candidate filtering as an optimization, but the deciding
operation is now the database fingerprint introduced by migration 138. Only the request whose
`INSERT ... ON CONFLICT DO NOTHING RETURNING` actually created the alert is allowed to treat it as new
and send an immediate critical-alert email.

### 7. Exchange-rate and security-price writers use atomic database writes

The market-data cron paths are normal multi-replica writers, so "find a row, then insert it" is not a
safe creation protocol.

The exchange-rate path now upserts the `(from_currency, to_currency, rate_date)` key in one statement,
and persists the forward and inverse directions inside one transaction. This removes the duplicate
insert race that could abort the pair transaction and leave the reverse direction stale.

The security-price refresh path is similarly moved away from stale entity-save/check-then-act shapes
toward guarded database updates/upserts, with focused specs covering the write contract.

### 8. Backup scheduling gets its own domain claim

Automatic backup already has a durable schedule row, so it does not need a second job-claim row for
the due window.

The claim is the schedule advance itself:

`UPDATE auto_backup_settings ... WHERE next_backup_at <= now RETURNING id`

Only the replica that advanced the schedule runs the backup. The outcome is then written with a
narrow update of only `lastBackupAt`, `lastBackupStatus` and `lastBackupError`, so a stale cron snapshot
cannot overwrite a concurrent administrator edit to folder/frequency/retention settings.

This branch also preserves the Phase 3 backup completeness contract:

- incomplete attachment exports remain visible as incomplete;
- partial automatic backups are not promoted or used to drive retention;
- export reads use one `REPEATABLE READ` snapshot;
- the long-lived export snapshot is bounded with `idle_in_transaction_session_timeout`;
- restore keeps staged attachment bytes and displaced-key cleanup ordering.

**Important:** upstream `main` now contains the later streaming-export work from #1109. This branch
must be rebased and the backup section re-reviewed against that implementation before merge; this
description does not claim that the pre-restack backup diff is the final one.

### 9. Whole-account replacement has a shared maintenance concept

Backup restore and self-service "delete my data" use `UserMaintenanceService` so destructive operations
can refuse overlap instead of independently replacing the same user's dataset.

An active `.mny` import is also treated as maintenance state, because a wipe-first import can leave the
account intentionally empty between the committed wipe and the later import write.

The restore itself runs under the maintenance lease and keeps its database replacement transactional.
This closes several direct overlap cases, but the current acquisition protocol still has an important
cross-table race documented under **Known blockers before ready-for-review** below.

### 10. Source-scanning guards make recurring mistakes red

`backend/src/common/db/derived-state-writers.guard.spec.ts` adds repository-level checks for mechanical
patterns that had reappeared in multiple services:

- unsanctioned direct `accounts.current_balance` writers;
- balance-adjusting services that do not use the shared locked-read protocol;
- bare entity removal beside a balance reversal;
- process-local `Set` / `Map` state used as a cron exclusion mechanism;
- open-coded `.length` checks on tuple-shaped raw `UPDATE` / `DELETE` results;
- `ActionHistoryService.record` calls placed inside an active database transaction even though the
  recorder deliberately swallows its own failures.

Reviewed exceptions are explicit allowlist entries with reasons. A new occurrence fails the suite
instead of depending on a reviewer remembering the same historical race.

## Database migrations

This PR introduces four migrations, reconciled to the current numbering on this branch:

### `138_concurrency_integrity_guards.sql`

Adds six database-enforced guards:

1. `job_claims` durable uniqueness;
2. `import_jobs` status constraint, `data_committed` checkpoint and one-active-import-per-user index;
3. `scheduled_transaction_postings` occurrence uniqueness;
4. budget-alert fingerprint uniqueness;
5. the supporting transaction-attachment index used by the concurrency-safe cap check;
6. `attachment_blob_tombstones` plus the deletion trigger for external attachment objects.

The migration includes preflight repair for legacy states that would otherwise make the new
constraints fail at deploy time: invalid import statuses, multiple active imports for one user and
duplicate budget-alert fingerprints.

`backend/test/integration/migration-138-preflight.integration.spec.ts` seeds those legacy states and
exercises the migration rather than assuming production data is already clean.

### `139_job_claim_delivery_record.sql`

Adds the delivery record needed to distinguish a claimed reminder from a successfully delivered one.
Existing permanent claims are backfilled as delivered so an upgrade does not suddenly re-send all
previously consumed windows.

### `140_job_claim_lease_token.sql`

Adds the per-attempt UUID token used by owned leases.

### `141_job_claim_lease_ownership_enforcement.sql`

Enforces lease ownership in PostgreSQL, including mixed-version rolling deployments where a previous
binary still issues the old untokenized statements.

`database/schema.sql` is updated to carry the resulting schema, indexes, RLS policies and trigger
state.

## Known blockers before ready-for-review

The branch has moved substantially since the first draft of this PR description, but the following
material issues are still reproducible on head `e41d66264acc2f6fb09b8ad6b83eb11b78bdaf08`.
They should be fixed before converting this PR out of Draft.

### HIGH — maintenance and `.mny` import do not acquire one atomic exclusion

**Paths**

- `backend/src/common/jobs/user-maintenance.service.ts`
- `backend/src/import/mny/mny-import-job.service.ts`
- `backend/src/import/mny/mny-import.service.ts`

`withMaintenanceLease()` first checks `import_jobs` and only afterwards takes a `job_claims`
maintenance lease. `MnyImportJobService.create()` independently inserts the active `import_jobs` row.
Those are two different tables with no shared lock or atomic predicate spanning them.

A realistic interleaving is:

1. restore/delete-data checks and sees no active import;
2. an import creates its `pending` row;
3. restore/delete-data takes the independent maintenance lease;
4. both sides now believe they own the user's dataset transition.

For an import that proceeds to write, a destructive replacement can commit while the import worker is
still active, and the worker can then write into the replaced dataset.

**Required correction:** make acquisition of the import slot and maintenance exclusion share one
database serialization mechanism. A check in one transaction followed by a claim in another is not a
guarantee.

**Regression test:** two real connections attempting maintenance acquisition and import-slot creation
in both orders; exactly one protocol may win and the loser must perform no destructive/data writes.

### HIGH — backup/maintenance exclusion is still a TOCTOU pre-check

**Path:** `backend/src/backup/auto-backup.service.ts`

Both manual and automatic backup ask `isUnderMaintenance(userId)` and then start the backup later.
`UserMaintenanceService` correctly documents `isUnderMaintenance()` as a **hint, not a guarantee**.
Maintenance can begin after that read and before the export snapshot is established.

This is especially dangerous once the `.mny wipeExistingData` path is repaired: the import may commit
its wipe and then rebuild the account over time, while the backup that passed the earlier pre-check can
start inside that window. A valid gzip/envelope containing the transitional dataset is not a safe
backup merely because the pre-check was false a moment earlier.

**Required correction:** make backup admission and maintenance acquisition mutually exclusive with a
shared lock/lease protocol that remains valid until the export snapshot is established (or otherwise
make the database snapshot itself prove it did not overlap the destructive transition).

**Regression test:** two real connections/futures that pause between the maintenance check and backup
snapshot while the other begins maintenance; assert no artifact is produced from the replacement
window and no retention/promotion runs.

### HIGH — emergency-access delivery is still uncoordinated across replicas

**Path:** `backend/src/emergency-access/emergency-access-monitor.service.ts`

The branch defines `EmergencyAccessReminder` and `EmergencyAccessGrantNotify` claim types, but the
emergency-access monitor does not use `JobClaimService` or an equivalent guarded domain update.

For reminders, every replica can read the same old `lastReminderSentAt`, send the same email and only
then save the marker.

For grants the impact is worse:

1. replica A and B both observe `grantedAt === null`;
2. each generates a different raw claim token;
3. each saves its token hash and sends its own email;
4. the later save can replace the hash that validates the earlier email's link.

The user can receive duplicate grant notifications and at least one delivered claim URL can be invalid
by the time it is clicked.

**Required correction:** claim the owner/contact delivery before generating/sending, and ensure token
state plus the delivery marker follow one reviewed retry contract. The existing owned-lease mechanism
is suitable only if the domain writes are also guarded against a superseded attempt.

**Regression test:** run two monitor instances against the same real database state; assert one grant
attempt owns each contact and every delivered URL still matches the stored token hash.

### MEDIUM — `.mny` "wipe existing data" currently refuses itself

**Paths**

- `backend/src/import/mny/mny-import.service.ts`
- `backend/src/users/users.service.ts`

`MnyImportService.start()` creates the pending import job first, then calls:

`usersService.deleteData(userId, ..., "import-wipe")`

But `UsersService.deleteData()` has a fourth `initiator` parameter whose default is
`"user-request"`. That default takes the maintenance lease, whose active-import check sees the pending
job the same request just created and returns `409`.

The caller therefore discards the pending job and `wipeExistingData` cannot start successfully.

**Required correction:** pass the import-specific initiator deliberately, but do so as part of the
HIGH-severity shared exclusion fix above; simply bypassing the maintenance lease without replacing the
cross-protocol guarantee would trade the obvious 409 for a race.

**Regression test:** start a wipe-first import through the real service boundary; assert the wipe and
worker start, while an actually concurrent restore/delete-data request is refused without writi

### DIFF
--- a/backend/src/accounts/mortgage-reminder.service.ts
+++ b/backend/src/accounts/mortgage-reminder.service.ts
@@ -13,6 +13,11 @@ import { emailTranslator } from "../i18n/email-translator";
 import { DEFAULT_LOCALE } from "../i18n/config";
 import { withSystemContext, withUserContext } from "../common/db/with-context";
 import { withScopedDb } from "../common/db/scoped-db";
+import {
+  JobClaimService,
+  JobClaimType,
+} from "../common/jobs/job-claim.service";
+import { createHash } from "node:crypto";
 
 /**
  * Service for handling mortgage term renewal reminders
@@ -22,6 +27,21 @@ import { withScopedDb } from "../common/db/scoped-db";
  * notificationEmail preference). Also logs each detected renewal so ops
  * can see what was processed even when SMTP is unavailable.
  */
+/**
+ * How long one replica holds the right to send this user's mortgage renewal notice.
+ *
+ * A lease, so a replica killed between claiming and sending costs a retry window
+ * rather than the delivery. It only has to outlast an SMTP round trip; whether the
+ * work is still owed is decided by the delivery record, not by the lease
+ * (audit RV4-006).
+ *
+ * The resulting contract is at-least-once: a process killed after SMTP accepted
+ * but before the record committed re-sends next run. For a reminder that is the
+ * right trade -- a duplicate nudge is an annoyance, a missed mortgage renewal is
+ * not. See docs/cron-jobs.md.
+ */
+const REMINDER_LEASE_MS = 10 * 60 * 1000;
+
 @Injectable()
 export class MortgageReminderService {
   private readonly logger = new Logger(MortgageReminderService.name);
@@ -31,7 +51,29 @@ export class MortgageReminderService {
     private emailService: EmailService,
     private configService: ConfigService,
     private readonly i18n: I18nService,
+    private readonly jobClaims: JobClaimService,
   ) {}
+
+  /**
+   * Release a lease a send did not use, so the next run can retry.
+   *
+   * Addressed by the lease token, not only by the work: a stalled worker must not
+   * delete a lease another replica has retaken (audit DR-RRV4-01).
+   */
+  private async releaseClaim(
+    userId: string,
+    claimKey: string,
+    leaseToken: string,
+  ): Promise<void> {
+    await this.jobClaims
+      .releaseLease(JobClaimType.MortgageReminder, userId, claimKey, leaseToken)
+      .catch((error: unknown) =>
+        this.logger.warn(
+          `Failed to release mortgage-reminder claim for user ${userId}: ` +
+            `${error instanceof Error ? error.message : String(error)}`,
+        ),
+      );
+  }
 
   /**
    * Run daily at 8:00 AM to check for upcoming mortgage renewals
@@ -85,6 +127,44 @@ export class MortgageReminderService {
     let skipCount = 0;
 
     for (const [userId, mortgages] of mortgagesByUser) {
+      // Claim before sending -- see BillReminderService for the reasoning. Every
+      // replica fires this cron, so without a durable record two healthy pods
+      // both email the same renewal notice (audit P4-018).
+      const claimKey = buildMortgageReminderClaimKey(mortgages);
+      // A **lease**, not a permanent claim, plus a durable delivery record.
+      //
+      // `claimOnce` was taken before the send and was the only record that the
+      // send was owed, so a replica killed in between left a permanent row that
+      // every later run read as "already handled" -- the mortgage reminder was never
+      // sent and nothing could notice, because the row could not tell "I sent
+      // this" from "I intended to" (audit RV4-006). The lease bounds *doing* the
+      // work; `delivered_at`, written after the send and re-read here, is what
+      // says it was done.
+      const claimed = await this.jobClaims.claimLease(
+        JobClaimType.MortgageReminder,
+        userId,
+        claimKey,
+        REMINDER_LEASE_MS,
+      );
+      if (!claimed) {
+        skipCount++;
+        continue;
+      }
+      const leaseToken = claimed;
+      if (
+        await this.jobClaims.wasDelivered(
+          JobClaimType.MortgageReminder,
+          userId,
+          claimKey,
+        )
+      ) {
+        // Already sent, durably. Hand the lease back rather than holding it for
+        // its whole TTL.
+        skipCount++;
+        await this.releaseClaim(userId, claimKey, leaseToken);
+        continue;
+      }
+
       try {
         // RLS (task C2): per-user reads run under the user's own context.
         const prefs = await withUserContext(userId, () =>
@@ -96,6 +176,7 @@ export class MortgageReminderService {
         );
         if (prefs && !prefs.notificationEmail) {
           skipCount++;
+          await this.releaseClaim(userId, claimKey, leaseToken);
           continue;
         }
 
@@ -108,6 +189,7 @@ export class MortgageReminderService {
         );
         if (!user || !user.email) {
           skipCount++;
+          await this.releaseClaim(userId, claimKey, leaseToken);
           continue;
         }
 
@@ -138,8 +220,17 @@ export class MortgageReminderService {
               );
 
         await this.emailService.sendMail(user.email, subject, html);
+        // The delivery record, after the side effect and never before it.
+        await this.jobClaims.markDelivered(
+          JobClaimType.MortgageReminder,
+          userId,
+          claimKey,
+          leaseToken,
+        );
         sentCount++;
       } catch (error) {
+        // A transient SMTP failure returns the day rather than consuming it.
+        await this.releaseClaim(userId, claimKey, leaseToken);
         this.logger.error(
           `Failed to send mortgage reminder to user ${userId}`,
           error instanceof Error ? error.stack : error,
@@ -216,4 +307,28 @@ export class MortgageReminderService {
       })),
     };
   }
+}
+
+/**
+ * The delivery key for one user's mortgage-renewal notice: today's date plus a
+ * digest of which mortgages and which term-end dates it names.
+ *
+ * Same shape as the bill reminder's key, and for the same reason: the date alone
+ * suppresses a legitimate second notice when another mortgage enters the window
+ * today, and the digest alone would let the same notice repeat tomorrow.
+ */
+export function buildMortgageReminderClaimKey(
+  mortgages: readonly Account[],
+): string {
+  const today = new Date();
+  const date = `${today.getFullYear()}-${String(today.getMonth() + 1).padStart(2, "0")}-${String(today.getDate()).padStart(2, "0")}`;
+  const fingerprint = [...mortgages]
+    .map((m) => `${m.id}:${formatDateYMD(m.termEndDate!)}`)
+    .sort()
+    .join("|");
+  const digest = createHash("sha256")
+    .update(fingerprint)
+    .digest("hex")
+    .slice(0, 32);
+  return `${date}#${digest}`;
 }
--- a/backend/src/backup/auto-backup.service.ts
+++ b/backend/src/backup/auto-backup.service.ts
@@ -1,4 +1,9 @@
-import { Injectable, Logger, BadRequestException } from "@nestjs/common";
+import {
+  Injectable,
+  Logger,
+  BadRequestException,
+  ConflictException,
+} from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
 import {
   DataSource,
@@ -10,6 +15,7 @@ import {
   Repository,
 } from "typeorm";
 import { withScopedDb } from "../common/db/scoped-db";
+import { affectedRowCount } from "../common/db/query-result";
 import { Cron } from "@nestjs/schedule";
 import { promises as fs, readdirSync, unlinkSync } from "fs";
 import { randomUUID } from "crypto";
@@ -33,6 +39,7 @@ import { User } from "../users/entities/user.entity";
 import { DemoModeService } from "../common/demo-mode.service";
 import { isShardableId, shardedSegments } from "../common/shard-path.util";
 import { withSystemContext, withUserContext } from "../common/db/with-context";
+import { UserMaintenanceService } from "../common/jobs/user-maintenance.service";
 import {
   UpdateAutoBackupSettingsDto,
   AutoBackupFrequency,
@@ -193,6 +200,7 @@ export class AutoBackupService {
     private readonly backupService: BackupService,
     private readonly backupEncryption: BackupEncryptionService,
     private readonly demoMode: DemoModeService,
+    private readonly maintenance: UserMaintenanceService,
     config: ConfigService,
   ) {
     this.defaultFolderPath = this.resolveConfiguredFolderPath(
@@ -598,6 +606,18 @@ export class AutoBackupService {
   async runManualBackup(
     userId: string,
   ): Promise<{ message: string; filename: string }> {
+    // The cron defers in this state; a manual run has a user watching, so it says
+    // so instead. Writing the file anyway would produce an empty backup and then
+    // rotate the last good one out to keep the retention count.
+    if (await this.maintenance.isUnderMaintenance(userId)) {
+      throw new ConflictException(
+        tr(
+          "errors.maintenance.inProgress",
+          "Another operation is currently replacing this account's data. Wait for it to finish and try again.",
+        ),
+      );
+    }
+
     // A user who has never opened the auto-backup settings still gets a working
     // manual run: the row is seeded with defaults here and persisted by the
     // save at the end of this method.
@@ -693,6 +713,74 @@ export class AutoBackupService {
     );
   }
 
+  /**
+   * Claim one due backup by advancing its own schedule.
+   *
+   * Every replica fires this cron, so without a claim a two-replica cluster
+   * writes every user's backup twice -- and worse, one replica's
+   * `enforceRetention` can delete the file the other is still writing.
+   *
+   * The claim is the schedule advance itself: `next_backup_at` is a column this
+   * job owns, so `UPDATE ... WHERE next_backup_at <= now RETURNING` re-evaluates
+   * the predicate after taking the row lock and exactly one replica gets a row
+   * back. No claim table and no lease to expire -- the row already carries the
+   * fact. Advancing *before* the export also means a crash mid-backup skips this
+   * window rather than retrying forever, which is the behaviour the failure path
+   * already had.
+   */
+  private async claimDueBackup(
+    settings: AutoBackupSettings,
+    now: Date,
+    nextBackupAt: Date,
+  ): Promise<boolean> {
+    const rows = await withUserContext(settings.userId, () =>
+      withScopedDb(this.dataSource, (manager) =>
+        manager.query(
+          `UPDATE auto_backup_settings
+              SET next_backup_at = $1
+            WHERE user_id = $2
+              AND enabled = true
+              AND next_backup_at IS NOT NULL
+              AND next_backup_at <= $3
+            RETURNING id`,
+          [nextBackupAt, settings.userId, now],
+        ),
+      ),
+    );
+    return affectedRowCount(rows) > 0;
+  }
+
+  /**
+   * Write only the columns describing how the run went.
+   *
+   * `repo.save(settings)` would write back every column of the snapshot this
+   * sweep read at the top, so a user who changed their folder, frequency or
+   * retention through the UI while the sweep was running would find those edits
+   * reverted. The outcome columns are the only ones this job is entitled to
+   * write; `next_backup_at` was already set by the claim.
+   */
+  private async recordBackupOutcome(
+    userId: string,
+    at: Date,
+    status: "success" | "partial" | "failed",
+    error: string | null,
+  ): Promise<void> {
+    await withUserContext(userId, () =>
+      this.scoped(AutoBackupSettings, (repo) =>
+        repo
+          .createQueryBuilder()
+          .update(AutoBackupSettings)
+          .set({
+            lastBackupAt: at,
+            lastBackupStatus: status,
+            lastBackupError: error,
+          })
+          .where("user_id = :userId", { userId })
+          .execute(),
+      ),
+    );
+  }
+
   @Cron("0 * * * *")
   async handleAutoBackupCron(): Promise<void> {
     const now = new Date();
@@ -714,6 +802,35 @@ export class AutoBackupService {
     this.logger.log(`Auto-backup cron: ${dueSettings.length} backup(s) due`);
 
     for (const settings of dueSettings) {
+      const nextBackupAt = this.calculateNextBackupAt(
+        settings.frequency as AutoBackupFrequency,
+        settings.backupTime,
+        settings.timezone,
+        now,
+      );
+
+      // Never back up a dataset that is mid-replacement. A `.mny` import with
+      // "start fresh" commits its wipe and then writes rows for minutes, so an
+      // hourly backup landing in that window would export the empty dataset,
+      // save it as today's file, and enforce retention -- rotating the last good
+      // backup out to make room for one containing nothing. Skipping without
+      // claiming leaves `next_backup_at` in the past, so the next hour retries
+      // (audit DR-04-02).
+      if (await this.maintenance.isUnderMaintenance(settings.userId)) {
+        this.logger.log(
+          `Auto-backup deferred for user ${settings.userId}: their data is being replaced`,
+        );
+        continue;
+      }
+
+      const claimed = await this.claimDueBackup(settings, now, nextBackupAt);
+      if (!claimed) {
+        // Either another replica took this window, or the user disabled or
+        // rescheduled the backup after this sweep read its snapshot. Both mean
+        // "not ours to run".
+        continue;
+      }
+
       try {
         settings.folderPath = this.resolveFolderPath(settings.folderPath);
         const userFolder = await this.resolveUserFolder(
@@ -737,15 +854,16 @@ export class AutoBackupService {
           timezone,
         );
 
-        settings.lastBackupAt = now;
-        settings.nextBackupAt = this.calculateNextBackupAt(
-          settings.frequency as AutoBackupFrequency,
-          settings.backupTime,
-          settings.timezone,
+        // applyBackupOutcome set lastBackupStatus/Error to reflect a complete
+        // ("success") or incomplete ("partial") artifact; record exactly that,
+        // never a hardcoded success, so a partial backup is not persisted as a
+        // full one (F3R7-001). The write is the targeted outcome UPDATE, not a
+        // whole-row save, so it cannot revert a concurrent settings edit.
+        await this.recordBackupOutcome(
+          settings.userId,
           now,
-        );
-        await withUserContext(settings.userId, () =>
-          this.scoped(AutoBackupSettings, (repo) => repo.save(settings)),
+          report.complete ? "success" : "partial",
+          settings.lastBackupError,
         );
 
         this.logger.log(
@@ -755,17 +873,11 @@ export class AutoBackupService {
         this.logger.error(
           `Auto-backup failed for user ${settings.userId}: ${error.message}`,
         );
-        settings.lastBackupAt = now;
-        settings.lastBackupStatus = "failed";
-        settings.lastBackupError = String(error.message).slice(0, 1024);
-        settings.nextBackupAt = this.calculateNextBackupAt(
-          settings.frequency as AutoBackupFrequency,
-          settings.backupTime,
-          settings.timezone,
+        await this.recordBackupOutcome(
+          settings.userId,
           now,
-        );
-        await withUserContext(settings.userId, () =>
-          this.scoped(AutoBackupSettings, (repo) => repo.save(settings)),
+          "failed",
+          String(error.message).slice(0, 1024),
         );
       }
     }
--- a/backend/src/budgets/budget-period.service.ts
+++ b/backend/src/budgets/budget-period.service.ts
@@ -6,6 +6,7 @@ import {
 import { tr } from "../i18n/translate";
 import { DataSource, EntityManager } from "typeorm";
 import { withScopedDb } from "../common/db/scoped-db";
+import { returnedRows } from "../common/db/query-result";
 import { Budget } from "./entities/budget.entity";
 import { RolloverType } from "./entities/budget-category.entity";
 import { BudgetPeriod, PeriodStatus } from "./entities/budget-period.entity";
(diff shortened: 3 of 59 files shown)


backend/src/scheduled-transactions/scheduled-transactions.service.ts
```diff
@@ -1618,66 +1630,126 @@ export class ScheduledTransactionsService {
...
+        // so a categorized scheduled transfer behaves like a one-off one
+        // (#743). Same precedence as the non-transfer branch: inline override >
+        // stored occurrence override > the schedule's own category.
+        const transferCategoryId = hasInlineCategoryId
+          ? postDto.categoryId
+          : storedOverride?.categoryId !== null &&
+              storedOverride?.categoryId !== undefined
+            ? storedOverride.categoryId
+            : scheduled.categoryId || undefined;
+        await this.transactionsService.createTransfer(userId, {
```
kenlasko, line 1711: **Blocking — every scheduled transfer post now throws.**

This commit moves `createTransfer` *inside* the new `withScopedDb(this.dataSource, async (m) => ...)` block; previously it ran outside it. The call chain is:

`transactionsService.createTransfer` → `TransactionTransferService.createTransfer` (`transaction-transfer.service.ts:192`) → `crossOwnerAccess.accountAccessFor` → `loadAccountForAuthorization` (`cross-owner-access.service.ts:282`), which is `withSystemContext(() => withScopedDb(...))`.

That inner `withScopedDb` joins the ambient transaction, but its identity is `{system: true}` while the outer transaction was opened under the user's identity. `sameIdentity()` returns false and it throws `IDENTITY_MISMATCH_MESSAGE`.

So both the hourly cron and manual posting fail for every transfer schedule. The specs don't catch it because they mock the transactions service.

The occurrence-claim insert genuinely does need to be atomic with the financial mutation, so the fix isn't just moving the call back out — the authorization read needs to happen *before* the transaction opens, with the resolved account passed down, or `loadAccountForAuthorization` needs to accept an ambient manager instead of asserting system identity.

---

backend/src/notifications/bill-reminder.service.ts
```diff
@@ -90,6 +132,52 @@ export class BillReminderService {
...
+      // A **lease**, not a permanent claim, plus a durable delivery record.
+      //
+      // `claimOnce` was taken before the send and was the only record that the
+      // send was owed, so a replica killed in between left a permanent row that
+      // every later run read as "already handled" -- the bill reminder was never
+      // sent and nothing could notice, because the row could not tell "I sent
+      // this" from "I intended to" (audit RV4-006). The lease bounds *doing* the
+      // work; `delivered_at`, written after the send and re-read here, is what
+      // says it was done.
+      const claimed = await this.jobClaims.claimLease(
```
kenlasko, line 156: **Blocking — no ambient identity context, so the whole cron aborts on the first due user.**

`JobClaimService` documents (line 49) that it "requires an ambient identity context (`withUserContext` / `withSystemContext`) like any other caller", and `claimLease` goes straight to `withScopedDb`, which throws `MISSING_CONTEXT_MESSAGE` without one.

The `withSystemContext` wrapper in this handler covers only the `manualBills` fan-out read at the top. Inside this per-user loop, the `UserPreference` and `User` reads are correctly wrapped in `withUserContext(userId, ...)` — but `claimLease` here, `wasDelivered` just below, `markDelivered` after the send, and `releaseClaim` are not.

Worse, this call sits *outside* the `try`, so the throw propagates out of `sendBillReminders` entirely: the first user with a due bill kills the run for every other user.

Wrapping the loop body in `withUserContext(userId, ...)` would cover all four call sites at once.

---

backend/src/accounts/mortgage-reminder.service.ts
```diff
@@ -85,6 +127,44 @@ export class MortgageReminderService {
...
+      // A **lease**, not a permanent claim, plus a durable delivery record.
+      //
+      // `claimOnce` was taken before the send and was the only record that the
+      // send was owed, so a replica killed in between left a permanent row that
+      // every later run read as "already handled" -- the mortgage reminder was never
+      // sent and nothing could notice, because the row could not tell "I sent
+      // this" from "I intended to" (audit RV4-006). The lease bounds *doing* the
+      // work; `delivered_at`, written after the send and re-read here, is what
+      // says it was done.
+      const claimed = await this.jobClaims.claimLease(
```
kenlasko, line 143: **Blocking — same missing-context defect as `bill-reminder.service.ts`.**

`claimLease` (and the `wasDelivered` / `markDelivered` / `releaseClaim` calls around it) run with no ambient identity context, so `withScopedDb` inside `JobClaimService` throws `MISSING_CONTEXT_MESSAGE`. It's outside the `try`, so no mortgage renewal notice is ever sent — the cron dies on the first user with a renewal in window.

Same fix: wrap the per-user loop body in `withUserContext(userId, ...)`.

Both this spec and the bill-reminder spec pass because `JobClaimService` is mocked, so the context requirement is never exercised. Worth a shared guard: the claim service is the kind of dependency where a mock hides the entire failure mode.

---

backend/src/backup/auto-backup.service.ts
```diff
@@ -714,6 +802,35 @@ export class AutoBackupService {
...
+      );
+
+      // Never back up a dataset that is mid-replacement. A `.mny` import with
+      // "start fresh" commits its wipe and then writes rows for minutes, so an
+      // hourly backup landing in that window would export the empty dataset,
+      // save it as today's file, and enforce retention -- rotating the last good
+      // backup out to make room for one containing nothing. Skipping without
+      // claiming leaves `next_backup_at` in the past, so the next hour retries
+      // (audit DR-04-02).
+      if (await this.maintenance.isUnderMaintenance(settings.userId)) {
```
kenlasko, line 819: **Blocking — `isUnderMaintenance` runs with no ambient identity context.**

`UserMaintenanceService.isUnderMaintenance` reaches `withScopedDb`, which throws `MISSING_CONTEXT_MESSAGE` here. The `withSystemContext` above covers only the `dueSettings` query; this loop body is outside it.

It's also outside the `try`, so the first due user aborts the hourly backup cron for everyone else.

This is inconsistent within the same loop: `claimDueBackup` and `recordBackupOutcome` a few lines below *are* wrapped. Wrapping the whole loop body would make it uniform.

Separately, and as the PR body already acknowledges: this check remains a TOCTOU pre-check. `isUnderMaintenance` is documented as a hint, and maintenance can begin between this read and the export snapshot. The comment above claims the window is closed ("Never back up a dataset that is mid-replacement"), which is stronger than what the code guarantees — worth softening until the shared admission protocol lands.

---

backend/src/budgets/budget-period.service.ts
```diff
@@ -68,29 +69,40 @@ export class BudgetPeriodService {
...
+    // never recomputed. The period row is locked first so a second close cannot
+    // interleave -- the monthly cron fires on every replica, and both replicas
+    // find the same OPEN period.
+    return withScopedDb(this.dataSource, async (manager) => {
+      const openPeriod = await manager.getRepository(BudgetPeriod).findOne({
         where: { budgetId, status: PeriodStatus.OPEN },
         relations: ["periodCategories"],
-      }),
-    );
+        lock: { mode: "pessimistic_write" },
```
kenlasko, line 87: **Blocking — `closePeriod` always fails: `FOR UPDATE` over a LEFT JOIN.**

`lock: { mode: "pessimistic_write" }` combined with `relations: ["periodCategories"]` makes TypeORM emit `FOR UPDATE` on a query with a LEFT JOIN. PostgreSQL rejects that outright:

```
FOR UPDATE cannot be applied to the nullable side of an outer join
```

So every close — cron and manual — raises. This is the only `pessimistic_write` site in the repo that also loads joined relations, which is why the pattern reads as fine by analogy with the others.

Fix is to split it: lock the `BudgetPeriod` row by id with no relations, then load `periodCategories` in a second query inside the same transaction. That still serializes the two cron replicas, which is the point of the change.

The spec presumably mocks the repository, so the SQL is never generated — an integration case against a real database would be the durable guard here.

---

backend/src/users/users.service.ts
```diff
@@ -538,11 +551,34 @@ export class UsersService {
...
+   *   takes the maintenance lease. `"mny-import"` is the importer's "start
+   *   fresh" wipe, which already holds the user's single import slot -- taking
+   *   the lease there would have it refuse itself, and consulting it would
+   *   refuse on the importer's own in-flight job (audit DR-04-02).
    */
   async deleteData(
     userId: string,
     dto: DeleteDataDto,
     reauthPurpose: OidcReauthPurpose = "delete-data",
+    initiator: "user-request" | "mny-import" = "user-request",
```
kenlasko, line 566: **Blocking — the new `initiator` param defaults to `"user-request"`, and the one caller that needs `"mny-import"` doesn't pass it.**

`MnyImportService.start()` (`backend/src/import/mny/mny-import.service.ts:158`) calls:

```ts
await this.usersService.deleteData(
  userId,
  { password: ..., oidcIdToken: ..., deleteAccounts: true, ... },
  "import-wipe",
);
```

Three arguments. `"import-wipe"` correctly lands on `reauthPurpose` — but `initiator` falls through to its `"user-request"` default, which takes `withMaintenanceLease`, whose `hasActiveImport(userId)` check sees the `import_jobs` row `this.jobs.create(...)` inserted eleven lines earlier. The wipe 409s and the caller discards the pending job, so `wipeExistingData` can never start.

`mny-import.service.ts` has no diff in this PR, so the new default silently changed that caller's behaviour — this is the failure mode the PR body files as the MEDIUM blocker.

Note the fix isn't just passing `"mny-import"`: as the PR body says, that trades the visible 409 for the cross-table race, since the import slot and the maintenance lease still have no shared serialization point. Both need to land together.

A defaulted parameter that flips a caller into taking a lock is worth avoiding as a shape — making `initiator` required would have turned this into a compile error.

---

backend/src/scheduled-transactions/scheduled-transactions.service.ts
```diff
@@ -1618,66 +1630,126 @@ export class ScheduledTransactionsService {
...
+          payeeId: scheduled.payeeId || undefined,
+          payeeName: scheduled.payeeName || undefined,
+          categoryId: transferCategoryId || undefined,
+          tagIds:
+            scheduled.tagIds && scheduled.tagIds.length > 0
+              ? scheduled.tagIds
+              : undefined,
+        });
+      } else {
+        await this.transactionsService.create(userId, transactionPayload);
```
kenlasko, line 1728: `transactionsService.create` fires an unawaited `ActionHistoryService.record`, and now that this call sits inside the posting transaction, that record joins it — it lacks the `runOutsideActiveScopedManager` guard `NetWorthService` uses for the same shape.

Two outcomes, both bad:

- the recorder deliberately swallows its own failures, so a failed insert aborts the ambient transaction and the *posting* fails with `25P02` (`current transaction is aborted`), attributed to nothing; or
- it races the commit and the undo entry is silently dropped.

This is exactly the pattern `derived-state-writers.guard.spec.ts` adds a check for in this same PR (`ActionHistoryService.record` calls inside an active transaction). The scan is per-file, so it doesn't see that a call two service hops away is now transactionally nested — worth noting in the guard's own comment, since the guard reads as stronger than it is.

---

database/migrations/142_job_claim_lease_ownership_enforcement.sql
```diff
@@ -0,0 +1,80 @@
...
+    END IF;
+    IF TG_OP = 'DELETE' THEN
+        RETURN OLD;
+    END IF;
+    RETURN NEW;
+END;
+$$;
+
+DROP TRIGGER IF EXISTS trg_job_claims_guard_delete ON job_claims;
+CREATE TRIGGER trg_job_claims_guard_delete
```
kenlasko, line 61: This BEFORE DELETE guard fires for any live tokenized lease, including the `ON DELETE CASCADE` from `users` — and a cascade delete carries no `app.job_claim_lease_token`, so it raises.

That means `deleteAccount` fails whenever the user holds a live lease. The windows are not small: AI insight generation is 15 minutes, `USER_MAINTENANCE_LEASE_MS` is 60, the demo reset is 30. Account deletion is also exactly the operation likely to run near a maintenance lease.

Options: exempt the cascade (check `pg_trigger_depth()` or a `app.cascade_delete` GUC set by the deletion path), or scope the trigger's `WHEN` to exclude it. Worth an integration case that deletes a user holding a live lease — the existing `job-claim-lease-rollout.integration.spec.ts` is the right home.

Also, the header comment on this file says "141" while the file is `142_...` — same off-by-one in `140_...sql` (says 139) and `141_...sql` (says 140), left over from the renumber described in the commit message.

---

database/migrations/139_concurrency_integrity_guards.sql
```diff
@@ -0,0 +1,348 @@
...
+    ON CONFLICT (storage_provider, storage_key) DO NOTHING;
+    RETURN OLD;
+END;
+$$;
+
+-- Only for providers whose bytes live outside PostgreSQL. The database provider
+-- keeps them in `attachment_blobs`, whose own foreign key cascades -- a
+-- tombstone there would describe work already done.
+DROP TRIGGER IF EXISTS trg_attachment_blob_tombstone ON transaction_attachments;
+CREATE TRIGGER trg_attachment_blob_tombstone
```
kenlasko, line 344: This trigger ships with no consumer. Nothing in the branch reads `attachment_blob_tombstones` — there's no sweeper service, and no reference to the table outside this migration and `schema.sql`. Meanwhile `schema.sql`'s comment on the table asserts a sweeper exists, which is a claim about source that isn't true (the `CLAUDE.md` rule about docs naming identifiers applies to schema comments too).

Two consequences:

1. The table grows without bound for every external-provider attachment ever deleted.
2. There's a latent hazard once a sweeper does land. `ON CONFLICT (storage_provider, storage_key) DO NOTHING` means a tombstone can outlive a *re-created* attachment pointing at the same key — a later sweeper reading this table would delete bytes that are live again. Whatever lands needs to re-check `transaction_attachments` before deleting, and that constraint belongs in a comment here now, not in the sweeper's PR.

Either ship the consumer in this PR or drop the trigger and table until it's ready. Half of a reclamation protocol is worse than neither half, because it looks done.

---

database/migrations/139_concurrency_integrity_guards.sql
```diff
@@ -0,0 +1,348 @@
...
+                          id DESC
+               ) AS rn
+          FROM import_jobs
+         WHERE status IN ('pending', 'running')
+      ) ranked
+     WHERE rn > 1
+)
+UPDATE import_jobs j
+   SET status = 'failed',
+       error_key = CASE WHEN j.data_committed THEN 'mnyJobStalledAfterCommit'
```
kenlasko, line 163: This preflight branches on `data_committed` and writes `error_key = 'mnyJobStalledAfterCommit'`, but neither exists yet:

- no backend constant or frontend i18n key for `mnyJobStalledAfterCommit`, so the wizard renders a missing-key placeholder for exactly the users whose data was half-written;
- nothing in the branch ever sets `import_jobs.data_committed` to true — the column is added by this migration and has no writer.

Since the column is uniformly `false`, `retryable = (j.data_committed = false)` marks every superseded job retryable, including ones whose rows are already in the ledger. Retrying those double-imports.

The `CASE` is written as if the checkpoint were already being maintained. Either add the writer (and the i18n key) in this PR, or drop the branch and retire superseded jobs as non-retryable — the conservative answer while the checkpoint doesn't exist.

---

kenlasko (changes requested):
Reviewed head `1d11b183bed7a0a168254c7993e80e1b1092d37a` (single commit on `3e8fde8`; 59 files, +5735/-857).

## Overall

The design is sound. Separating a permanent claim from an expiring lease, a per-attempt token, and a delivery record is the right decomposition — the `claimOnce`/`claimLease`/`markDelivered` split is doing real work rather than putting one generic mutex around every cron, and the reasoning for why the demo reset and the intraday generator get *different* contracts is the kind of thing that usually gets flattened. Pushing lease ownership into a PostgreSQL trigger so a previous-release pod can't mutate a retaken lease is a genuinely good call for rolling deploys, and `job-claim-lease-rollout.integration.spec.ts` exercising the literal old statement shapes against a migrated database is the right way to prove it.

But the branch does not currently work. Six paths are broken by the coordination changes themselves, and the specs don't catch any of them because `JobClaimService` is mocked and the repository-level locks are never exercised against a real database.

## Blocking

1. **Every scheduled transfer post throws.** `createTransfer` moved inside the posting `withScopedDb`; its authorization read is `withSystemContext(withScopedDb(...))`, which can't join a user-identity transaction — `IDENTITY_MISMATCH_MESSAGE`. Cron and manual both.
2. **`bill-reminder.service.ts`** — `claimLease`/`wasDelivered`/`markDelivered`/`releaseClaim` called with no ambient identity context; `MISSING_CONTEXT_MESSAGE`, thrown outside the `try`, so the first due user aborts the run for everyone.
3. **`mortgage-reminder.service.ts`** — identical defect; no renewal notice is ever sent.
4. **`auto-backup.service.ts`** — `isUnderMaintenance` has the same problem, and is inconsistent with `claimDueBackup`/`recordBackupOutcome` in the same loop, which *are* wrapped.
5. **`closePeriod` always fails** — `pessimistic_write` plus `relations: ["periodCategories"]` emits `FOR UPDATE` over a LEFT JOIN, which PostgreSQL rejects.
6. **Wipe-first `.mny` import refuses itself** — the new `initiator` param defaults to `"user-request"`, and the one caller that needs `"mny-import"` wasn't updated (`mny-import.service.ts` has no diff in this PR).

Details inline. Items 2–4 are one mistake in three places — the fix is the same in each.

## Also flagged inline

Unawaited `ActionHistoryService.record` now transactionally nested in the posting path (the exact shape this PR's own new guard forbids — the guard's per-file scan can't see it two service hops away); the BEFORE DELETE lease guard raising on the `users` cascade, so `deleteAccount` fails while any live lease exists; the `attachment_blob_tombstones` trigger shipping with no consumer while `schema.sql` claims one exists; and the migration-138 preflight branching on a `data_committed` column nothing writes and an `error_key` nothing translates.

## Testing

Four of the six blocking defects are things a mock cannot see. Given that this PR's whole subject is coordination, the specs need to reach a real database for: the claim service's context requirement, `closePeriod`'s generated SQL, the cascade-vs-lease-guard interaction, and account deletion holding a live lease.

More broadly — the guard spec added here checks source patterns per file, which catches a re-introduced `Set`-as-cron-exclusion but not a call that became transactionally nested through a service hop. Worth saying so in the guard's own comment, since it currently reads as stronger than it is.

## Minor

Migration files are numbered 139–142 but the headers of the last three say 139, 140, 141 (off by one after the renumber). `docs/cron-jobs.md` lists `0 04 * * *` for a `0 4 * * *` handler. `sweepOldClaims` open-codes `result[1]` instead of the `affectedRowCount` helper this PR introduces. `updatePreferences` re-inlines what `patchUserPreferences` already does — looks like a stale-branch artifact.

## On the PR's own blockers

The three HIGH items in the description are still accurate on this head: the import/maintenance cross-table TOCTOU, the backup admission TOCTOU, and the uncoordinated emergency-access delivery. The MEDIUM `.mny` self-refusal is blocking item 6 above. The restack onto current `main` (streaming export from #1109) is still outstanding, so the backup section needs re-reviewing against that implementation regardless.

---

kenlasko:
The findings from my review above have been applied on a branch, if it's useful to pull from: **`claude/review-pr-1111-jyju6s`**, opened as #1117.

It's this branch's commit `1d11b18` restacked onto current `main` plus the fixes. Worth noting the restack was cheaper than this PR's description expects: `3e8fde8` is already in `main`, so #1109's streaming-export work is in the base and the backup section needed no reconciliation. One conflict, in `backup-restore.service.ts`, where `main` had since added `unusableAiProviderKeys` alongside `skippedAttachments`.

Two things there that go beyond what I filed inline:

- **The transfer fix is larger than the comment on line 1711 suggests.** There is a second `withSystemContext` at `transaction-transfer.service.ts:292`, for the cross-owner leg write, which also cannot nest — so hoisting only the authorization read would not have been enough. `createTransfer` is now `prepareTransfer` → `writeTransferLegs(manager?)` → `completeTransfer`: authorization decided above the posting transaction, only the leg writes joining it, and the post-commit half (action history, tags, net worth) after it commits. The occurrence claim stays atomic with the money.
- **`LockScope.UserImport` was already declared in this branch and never used.** It's now taken by both `MnyImportJobService.create` and `UserMaintenanceService.withMaintenanceLease`, which gives the two protocols one object to contend on — the HIGH cross-table TOCTOU in the description. The `.mny` initiator fix rides on that rather than bypassing the exclusion.

Also: migrations renumbered 140–143, because the restack collided with a `139` that landed in `main` in the meantime.

Two of my minor findings did not survive contact and are not worth acting on: `0 04 * * *` is literally what `CronExpression.EVERY_DAY_AT_4AM` evaluates to, so `docs/cron-jobs.md` was correct; and `updatePreferences` needs its read *between* the row's materialization and the write, which `patchUserPreferences` cannot express — not duplication.

The three HIGH items in this PR's own description: the import/maintenance TOCTOU is addressed as above, and the backup-admission TOCTOU and the uncoordinated emergency-access delivery are **not** — they're still open. The backup pre-check now says in its comment that it narrows the window rather than closing it, instead of claiming otherwise.

---

PR #1142, opened 2026-08-13
06: Mask user-authored currency metadata in support backups (v1.15.0)

WMP:
## Summary

Mask the free-text name and symbol of user-created currencies in de-identified support backups while keeping standard currency reference data usable.

The support-backup UI promised that names were masked, but `currencies.name` and `currencies.symbol` were always kept verbatim. A custom currency such as `Alice Family Credits` / `AFam` could therefore disclose household, employer, or project information in an artifact intended to leave the user's data boundary.

## What changes

- Extend support-backup rules with a row-conditional `maskWhen` rule.
- Add a `userCreatedCurrency` predicate based on a non-null `created_by_user_id`.
- Mask `name` and `symbol` only for user-created currencies.
- Keep standard ISO currency names and symbols readable as public reference data.
- Keep the currency `code` unchanged so references from other tables continue to resolve.
- Update support-backup explanatory copy to disclose that custom currency codes remain readable while their names and symbols are masked.
- Localize the updated copy across supported locales.

## User impact

User-authored currency labels no longer leak verbatim into support artifacts, while support engineers can still interpret currency-denominated data and restore references remain coherent.

## Validation

The branch adds/updates coverage for:

- masking a custom currency's name and symbol;
- preserving standard currency metadata;
- preserving the custom currency code for referential integrity;
- gunzipping the generated support artifact and asserting the original custom name and symbol do not appear anywhere;
- keeping the frontend disclosure aligned with the de-identification rules.

## Audit context

Closes the support-backup de-identification gap identified during Phase 6 remediation.


WMP:
This change is already on `main`, implemented more comprehensively — so this PR can be closed as superseded.

The custom-currency masking landed via **#1077** (`WMP/pr/03-03-backup-core`, merged Aug 6 2026 as `8ebca94`), in these commits:

- **`74fc879`** — *"Harden manual backup and restore: integrity, self-contained artifacts, bounded memory"* — adds `pseudonymiseCustomCurrencies` + `allocatePseudonymCode` in `support-backup.service.ts`, plus the `SupportBackupService custom currencies` test suite.
- **`2a322cb`** — *"Address PR 1077 maintainer review"* — adds `backend/src/currencies/currency-reference-columns.ts`, the named list used to rewrite every currency reference.

How `main` covers each part of this PR:

| This PR (`maskWhen`) | `main` (`pseudonymiseCustomCurrencies`) |
|---|---|
| Mask user-created currency `name` | Same — `maskText(row.name)` |
| Mask `symbol` | Stronger — replaced with the generic sign `¤`, so length isn't leaked |
| Keep `code` verbatim (for referential integrity) | Pseudonymises `code` (random, unique, not an ISO code) **and** rewrites every reference via `CURRENCY_REFERENCE_COLUMNS`, so the artifact still restores while two exports can't be correlated by the code |
| Leave standard ISO currencies readable | Same — rows with `created_by_user_id IS NULL` are untouched |
| `created_by_user_id: keep` | Fixed here — `keep` leaks the *creating* user's real UUID (a custom currency from another user of the same instance survives id-remapping); `main` overwrites it with the exporting user's id first |

The one part not on `main` is the modal note change (*"currency codes stay readable... including the code of any currency you created yourself"*). That statement is false against `main`'s behaviour, since `main` pseudonymises the code — so it shouldn't be ported.

Verified locally against `main` (`883f8b0`): `support-backup` + `currency-references` suites green (76/76).

---

PR #1145, opened 2026-08-13
06: Keep unknown financial values distinct from zero (v1.15.0)

WMP:
## Summary

Keep missing rates, failed valuations, and unpriced holdings visibly distinct from measured zero values.

Several financial surfaces previously converted an unknown state into a numeric result. Missing FX rates could be rendered under the target currency symbol without conversion, failed brokerage summaries could remove holdings from totals as if their value were zero, and Monte Carlo could start from zero when the portfolio could not be valued.

### Missing FX rates

- Return `null` from conversion helpers when a required rate is unavailable instead of returning the unconverted amount.
- Propagate partial totals together with the currencies that could not be converted.
- Render partial/unknown markers for aggregates and single values.
- Use gaps (`null`) in chart series rather than fabricated numeric points.
- Withhold cumulative forecasts when a missing conversion invalidates all later points.
- Add frontend FX-rate rounding at storage precision instead of money precision.

### Failed and partial portfolio valuation

- Distinguish a failed portfolio-summary request from a successful summary in which a brokerage account genuinely has zero holdings.
- Stop folding unpriced holdings into allocations as zero-valued positions.
- Mark portfolio totals as partial and identify unpriced symbols where the payload exposes them.

### Monte Carlo

- Return `null` when the current portfolio value cannot be computed.
- Refuse saved and ad-hoc simulations when the starting value is unknown.
- Keep a genuine zero balance and an empty selection as valid, known states.
- Make `HistoricalStats.currentBalance` nullable end to end so the report can disable Run and explain why.

### Localization

- Localize the new unknown/partial-state messages across supported locales.

## Financial impact

The UI no longer presents an unavailable conversion or valuation as a measured amount. This prevents plausible but false totals, allocations, forecasts, and retirement projections from being shown as complete results.

## Validation

The branch adds/updates regression coverage for:

- missing FX rates returning unknown instead of a 1:1 value;
- partial aggregate propagation;
- Accounts-page behavior when portfolio summary fails;
- unpriced holdings in portfolio summaries and allocations;
- forecast gaps and invalid cumulative series;
- Monte Carlo refusing an unknown starting value while still accepting real zero and empty-selection cases.

## Audit context

Addresses Phase 6 findings **P6-006**, **P6-007**, and **P6-008**.

## Checklist


WMP:
@kenlasko : this will need your tests on your big data, this pr make changes (unknown isn't zero) everywhere...

WMP:
## Independent read-only regression validation (1.14.0 → this PR)

I ran the branch's own `e2e-financial-regression` harness against a **real, existing
database** (not an empty CI one) to check that this PR's "unknown is not zero" work
doesn't silently move any financial figure that a user with **complete** data can see.

Sharing the setup, the result, and a few notes that might be useful for the harness itself.

### Setup

- Isolated test instance in Kubernetes (own namespace), pointed at a **disposable CNPG
  copy** of a real database (~9.4k transactions, back to 2019, multi-currency: PLN default,
  EUR/USD holdings). The live instance was never touched.
- Walked images from `1.14.0` forward through several intermediate release PRs up to this
  PR's build, capturing the financial signal set on each version and diffing consecutive
  captures. Everything stayed **read-only** (the route guard blocked/aborted any write).
- The comparison that matters for *this* PR is the clean **1.14.0 → PR build** diff, since
  that isolates the full delta this branch carries.

### Result on complete data

The "unknown is not zero" contract produced **no regression** on a complete-data account:

| transition | count |
|---|---|
| `value → unknown` | 0 |
| `unknown → value` | 0 |
| `value → missing` | 0 |
| `missing → value` | 0 |

No previously-known figure became unknown, missing, or silently changed status.

What the API layer *did* show, all consistent with the contract:

- The new `account.extraPaymentAmount` field (migration `154_account_extra_payment_amount`)
  renders as **`null` / unknown**, not `0`, for accounts that have no extra payment set —
  i.e. a genuinely-absent value is reported as unknown, exactly as intended.
- The new `holdingsByAccount[*].unpricedHoldingsCount` completeness telemetry is present
  (0 here, since every holding is priced).

Because the account's data is complete, the contract has **no visible effect on the existing
figures** — which is the expected outcome: it only changes the incomplete-data cases, and
this account has none in the captured surface. The substantive value changes seen versus
`1.14.0` come from *other* v1.15.0 work carried along (credit-netting for spending totals,
adjusted-close daily valuations), not from this contract.

> Caveat I want to be explicit about: with a single complete-data account, the harness does
> **not** exercise the incomplete-data paths this PR is really about (missing price / missing
> FX → "unknown" rendering). Those are better covered by this branch's frontend unit tests
> and by the `exceptional-missing-fx` suite. This run validates the *"complete data is
> untouched"* half of the contract, not the *"incomplete data stops lying"* half.

### Notes that might help the harness

Running the harness against a **deployed** instance (vs the CI compose) needed a handful of
adaptations that may be worth folding back in:

1. **Login selector is ambiguous when OIDC/SSO is enabled.** `getByRole('button',
   {name: /sign in/i})` matches both "Sign in" and "Sign in with SSO", so it fails strict
   mode. Exact match (`{name: 'Sign in', exact: true}`) fixes it.
2. **`page.goto(..., {waitUntil: 'domcontentloaded'})` hangs** under the route interception on
   the Next.js App Router (streaming RSC) — DOMContentLoaded doesn't reliably fire, so every
   screen timed out. Switching to `waitUntil: 'commit'` + a passive settle (no
   `waitForLoadState('networkidle')` / `locator.waitFor`, which stalled the client render)
   made it reliable.
3. **`devices['Desktop Chrome']`'s pinned userAgent** prevented the client from rendering at
   all under the runner (blank screens, every DOM signal `missing`); a plain desktop viewport
   with the bundled Chromium UA rendered fine.
4. **Two app-initiated POSTs fire on navigation** and trip the read-only guard:
   `POST /budgets/category-budget-status` (a read that uses a body — safe to allow) and
   `POST /securities/prices/refresh/selected` (a genuine write — correctly blocked; worth
   suppressing rather than failing the run, since the app auto-fires it on the Investments
   screen).
5. **Positional array signals cause false diffs.** `spending-by-category` is captured as
   `data[N].total`, but the array reorders/shrinks when a total changes, so consecutive diffs
   flag index shifts as "changes". Keying array leaves by a stable identity (category id/name)
   instead of position would remove that noise. (Same shape applies to any ordered report
   array.)

### Other caveats

- Price-derived values (portfolio / holdings / net-worth) drift between runs from the
  scheduled FX/price refresh, independent of code — filtered as noise below a small relative
  threshold.
- Per-PR branch images have different bases, so a diff between two consecutive PR builds mixes
  the PR's own change with what its branch happened to include; the clean signal is a PR
  against its own parent.

Happy to share the harness adaptations as a diff if useful.

PR #1155, opened 2026-08-14
fix(securities): persist website and IR website on edit (issue #1122)

### LINKED ISSUE #1122: Editting/Adding website to a Security
For all my ETFs, they list a greyed out apple.com and investor.apple.com when I look at the security details.

Image

I paste a valid link to the providers webpage for the ETF but it does not save the link after hitting save. It does this for Website and Investor relations website.

Should I be able to include my own custom urls for these fields?

WMP:
## Linked discussion / issue

Closes #1122
Approved in: #1122

## Summary

The `SecuritiesService.update()` method was omitting `website` and `irWebsite` fields from its explicit property-mapping block. Because the service normalizes these fields (converting bare domains to HTTPS URLs and empty strings to null), the normalized values were computed but never copied onto the entity before save. This caused edits to these fields to silently fail to persist.

The fix adds explicit assignments for both fields in the update method, matching the pattern used for all other scalar properties. Two regression tests verify that:
1. Website and IR website URLs are normalized and persisted correctly
2. Empty submissions clear these fields (set to null)

## Checklist

- [x] This PR addresses a **single concern** (field persistence bug).
- [x] No user-facing strings added (bug fix only).
- [x] No shared/core areas refactored.
- [ ] AI assistance disclosure: N/A

https://claude.ai/code/session_011bVQLCsojE4rtguxA9sGeX

### DIFF
--- a/backend/src/securities/securities.service.spec.ts
+++ b/backend/src/securities/securities.service.spec.ts
@@ -715,6 +715,37 @@ describe("SecuritiesService", () => {
       expect(saved.isFavourite).toBe(true);
     });
 
+    // Regression for issue #1122: the explicit property-mapping block omitted
+    // both address fields, so an edit normalised them and then dropped them --
+    // the save persisted the unchanged security and the link never stuck.
+    it("persists and normalizes website and irWebsite updates (issue #1122)", async () => {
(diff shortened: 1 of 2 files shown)


WMP:
Tested, ready to merge

Issue #1154, opened 2026-08-14
Scheduled investment: editing away from BUY/SELL leaves a stale investmentFundingAccountId, misrouting the posted cash


WMP:
## Severity
Medium — real financial-correctness defect (cash lands in the wrong account), but only on the *edit* path and only for cash-bearing non-funding actions (DIVIDEND / INTEREST / CAPITAL_GAIN).

## Summary
When an existing scheduled investment is edited to change its action **away from** `BUY`/`SELL` (the only actions that use a funding account), the previously-stored `investmentFundingAccountId` is **not cleared**. It persists on the row, and when the occurrence posts, the cash side of the transaction is routed to that stale account instead of the brokerage's linked cash sleeve.

This is a pre-existing defect: the lines involved are not touched by the current stored-price-discipline PR. It was surfaced while reviewing that PR.

## Root cause
Two halves that only misbehave together:

1. **Frontend omits instead of clearing.** `frontend/src/components/scheduled-transactions/ScheduledTransactionForm.tsx` (submit payload) sends:
   ```ts
   investmentFundingAccountId:
     FUNDING_ACCOUNT_ACTIONS.includes(investmentAction) && investmentFundingAccountId
       ? investmentFundingAccountId
       : undefined,
   ```
   On the non-funding branch this yields `undefined`, which is dropped from the JSON body rather than sent as an explicit clear.

2. **Backend update treats "absent" as "leave unchanged."** `backend/src/scheduled-transactions/scheduled-transactions.service.ts` (update path) writes the column only when the key is present:
   ```ts
   if (updateData.investmentFundingAccountId !== undefined) { /* write column */ }
   ```
   Because the key is absent, the previously-stored value is preserved.

3. **Post routing is unconditional.** `postInvestment` passes `fundingAccountId: scheduled.investmentFundingAccountId || undefined` regardless of the action, so a `DIVIDEND`/`INTEREST`/`CAPITAL_GAIN` occurrence sends its cash to the stale account.

The **create** path is clean: `undefined` → backend defaults the column to `null`.

## Reproduction
1. Create a scheduled `BUY` with a funding account set to "Checking".
2. Edit the schedule; change the action to `DIVIDEND` and set a total amount; save.
3. Inspect the row: `investment_funding_account_id` is still "Checking".
4. When the occurrence posts, the dividend cash transaction lands in "Checking" instead of the brokerage's linked cash account.

[...]

PR #1168, opened 2026-08-16
Fix scheduled-investment state cleanup, posting consistency, and concurrency (#1154)

WMP:
## Summary

This PR fixes issue #1154 and then hardens the same scheduled-investment posting path through a multi-round adversarial review.

The original defect was straightforward but financially dangerous: a scheduled investment edited away from `BUY` / `SELL` could retain a stale `investmentFundingAccountId`. When a later `DIVIDEND`, `INTEREST`, or `CAPITAL_GAIN` occurrence posted, that stale account could be forwarded into the investment posting path and route cash to the wrong account instead of the brokerage's linked cash account.

The fix is intentionally defensive:

- only `BUY` and `SELL` are allowed to use an explicit funding account;
- stale funding accounts are cleared on create/update and ignored at post time;
- MNY import follows the same invariant;
- action-specific stale investment fields are cleared;
- post-time overrides are revalidated after precedence resolution;
- posting refuses stale pre-lock schedule, override, FX-tuple, and split generations;
- cron rechecks `isActive` / `autoPost` under the row lock;
- loan split recalculation now serializes on the same parent schedule lock;
- `LINE_OF_CREDIT` uses the same loan-payment recalculation path as `LOAN` and `MORTGAGE`.

The branch was repeatedly reviewed after the initial fix. The review cycle found several adjacent edge cases that were real enough to fix rather than leave as unrelated follow-up debt.

**Base:** `09d7396d90aec8f7a865b802b7d50b58115ac735`  
**Current head:** `0e0beca1484b522cfa77db32a282246c3e778ec0`  
**Branch delta:** 12 commits

---

## AI review / implementation trail

Two AI systems had distinct roles in this branch:

- **Claude Code / Claude Opus 4.8** implemented the branch and performed the initial/internal review passes. Those passes found several write-side consistency gaps before the external re-review cycle, including the MNY writer bypass, stale action-specific numeric fields, an insufficient regression assertion, and inaccurate comments about FX/settlement behavior.
- **ChatGPT** performed the external adversarial re-review rounds in this review thread. Those rounds found the broader API validation, stale-snapshot, split-generation, cron, loan-recalculation, and `LINE_OF_CREDIT` defects that were subsequently implemented by Claude Code.

The table below records the discovery source separately from the implementation commit so the review provenance is explicit.

---

## Edge cases closed

| # | Edge case / invariant | Failure before the fix | Found by | Fixed / documented in |
|---|---|---|---|---|
| 1 | Stale funding account after changing away from `BUY` / `SELL` | A later dividend/interest/capital-gain occurrence could route cash through the old funding account instead of brokerage linked cash | Original issue #1154 | `28a5d49` |
| 2 | Non-funding action created/updated with a funding account | Invalid state could remain persisted even when the action would never use that account | Original #1154 implementation review | `28a5d49` |
| 3 | Frontend omitted the clear on action change | Editing `BUY -> DIVIDEND` could leave the backend column unchanged | Original #1154 implementation review | `28a5d49` |
| 4 | MNY bill import bypassed the funding-account invariant | Imported non-`BUY`/`SELL` investments could persist the same stale column even though normal create/update cleaned it | Claude Code internal multi-pass review | `7ce0b1e` |
| 5 | Funding-action membership duplicated across writers | Scheduled service and import could drift on which actions may use funding accounts | Claude Code internal multi-pass review | `7ce0b1e` |
| 6 | `REINVEST` legacy row carried a stale funding account | Share-moving but cashless action could still contain an irrelevant funding account; post must ignore it | Claude Code internal review | `7ce0b1e` |
| 7 | Stored FX rate could outlive its settlement basis | A stored scalar rate could be reused after action/account/security settlement basis changed | Claude Code review; later expanded by ChatGPT | documented `fe2b9be`, behavioral partial fix `b2f7c1d`, durable follow-up `95eead2` |
| 8 | Action switch retained stale numeric fields | `BUY -> DIVIDEND`, `DIVIDEND -> BUY`, or `BUY -> ADD_SHARES` left irrelevant quantity/price/commission/total values on the row | Claude Code review | `5cba485` |
| 9 | BUY “keep funding account” test was too weak | A future accidental empty-string/other write could pass while still corrupting the stored funding account | Claude Code adversarial test review | `31386b7` |
| 10 | Comment claimed all external FX work was hoisted before the DB transaction | Investment settlement FX can still call the quote provider while the schedule lock is held | Claude Code end-to-end review | `ffc05b8` |
| 11 | Post-time overrides bypassed create/update validation | A negative dividend total or zero-quantity BUY supplied at posting time could create money despite valid stored configuration | ChatGPT external adversarial re-review #1 | `b2f7c1d` |
| 12 | Explicit `null` was collapsed into “use stored value” during update validation | `{ investmentTotalAmount: null }` could validate against the old value and then persist `null` | ChatGPT re-review #1 | `b2f7c1d` |
| 13 | Zero/non-positive investment FX rate accepted | Invalid persisted rate could reach settlement code | ChatGPT re-review #1 | `b2f7c1d` |
| 14 | Settlement-basis edit retained an old stored FX rate | A rate for the old settlement tuple could be applied after account/security/action basis changed | ChatGPT re-review #1 | `b2f7c1d` |
| 15 | `INTEREST` could retain a hidden security | UI/action transition could leave security-derived settlement state that no longer belonged to the effective action | ChatGPT re-review #1 | `b2f7c1d` |
| 16 | Funding account validation ran for actions that do not use it | A request could be rejected for an irrelevant field that would be dropped at persistence time | ChatGPT re-review #1 | `b2f7c1d` |
| 17 | Investment posting used a pre-lock row snapshot | Concurrent `BUY -> DIVIDEND` / security / funding edit could post the old action to the old account | ChatGPT re-review #1 | `b2f7c1d` |
| 18 | Post dialog preview could still show the stale funding account | UI projected the wrong account even though backend posting no longer used it | ChatGPT re-review #1 | `b2f7c1d` |
| 19 | Funding-action documentation overclaimed non-BUY/SELL behavior | `REINVEST` and share-only actions have no external cash leg; they do not all “settle to linked cash” | ChatGPT re-review #1 follow-up | `9339e14` |
| 20 | Locked-row protection covered only part of the posting shape | Concurrent plain/transfer/investment mode, account, amount, category, or investment-scalar changes could leave a prebuilt stale payload | ChatGPT external adversarial re-review #2 | `a50fc77` |
| 21 | Occurrence override was read before the transaction and then consumed/deleted | Concurrently edited override could be posted stale or deleted without posting its new state | ChatGPT re-review #2 | `a50fc77` |
| 22 | `update()` could apply action-dependent clears from a stale snapshot | One edit could overwrite a concurrent action change | ChatGPT re-review #2 | `a50fc77` |
| 23 | Frontend `INTEREST` fix over-cleared a legitimate security | A presentation-only edit could clear security state that was intentionally stored for a security-specific interest entry | ChatGPT re-review #2 | `a50fc77` |
| 24 | Base scheduled split set was outside the parent mutation guard | Concurrent transfer/investment split retarget could post money to the old account even though parent scalars were unchanged | ChatGPT external adversarial re-review #3 | `e709eb3` |
| 25 | Foreign-currency tuple omitted from mutation basis | `originalAmount`, `originalCurrencyCode`, or `exchangeRate` could change while account-currency `amount` stayed the same, allowing a stale converted post | ChatGPT re-review #3 | `e709eb3` |
| 26 | Cron did not re-check live `isActive` / `autoPost` after candidate selection | User could disable auto-post after cron selection and still get an automatic post | ChatGPT re-review #3 | `e709eb3` |
| 27 | Override guard depended too heavily on `updatedAt` | Same-millisecond changes could escape timestamp-only detection | ChatGPT re-review #3 | `e709eb3` |
| 28 | Loan split recalculation mutated child splits without the parent schedule lock | Principal/interest could change after the posting split guard but before the financial write, causing a stale occurrence | ChatGPT external adversarial re-review #4 | `95eead2` |
| 29 | Loan recalculation accepted a loan ID derived from a stale pre-lock split snapshot | Concurrent retarget `Loan A -> Loan B` could recalculate current splits using the wrong loan account | ChatGPT re-review #4 | `95eead2` |
| 30 | FX currency-pair residual was only partially described | Durable rates exist on parent schedules, scheduled investment splits, and override JSON; all three can become stale when referenced currency changes with stable IDs | ChatGPT re-review #4 | `95eead2` tracker |
| 31 | `LINE_OF_CREDIT` excluded from loan split recalculation | LOC scheduled payments kept reusing the first principal/interest allocation rather than advancing after each payment | ChatGPT external adversarial re-review #5 | `0e0beca` |
| 32 | Split-guard comment claimed an unlocked category writer was “still caught” | A category reassignment can commit after the point-in-time split comparison; the guarantee was stronger than the mechanism | ChatGPT re-review #5 | `0e0beca` comment corrected; residual explicitly accepted |

---

### 1. Funding-account routing is now action-authoritative

The central invariant is:

> An explicit investment funding account is meaningful only for `BUY` and `SELL`.

`BUY` draws purchase cash from the funding account. `SELL` deposits proceeds there. Cash-bearing `DIVIDEND`, `INTEREST`, and `CAPITAL_GAIN` settle through the brokerage's linked cash account. `REINVEST` and share-only actions do not have an external funding-account cash leg.

The branch enforces this at multiple boundaries:

- `postInvestment()` only forwards `investmentFundingAccountId` for `BUY` / `SELL`;
- create/update clear the field for every other effective action;
- the frontend sends the required clear when the user changes actions;
- MNY import uses the same shared `FUNDING_ACCOUNT_ACTIONS` authority;
- regression tests cover both the “drop” and “keep” branches.

The post-time gate is the authoritative safety net: legacy rows already carrying stale data cannot misroute cash even before they are edited again.

### 2. Action changes clean their obsolete state

Changing a scheduled investment action now removes fields that are not part of the new action's model.

Examples:

- `BUY -> DIVIDEND`: clear quantity, price, commission, and funding account; keep/write the dividend total.
- `DIVIDEND -> BUY`: clear the old total; keep/write BUY quantity and price.
- `BUY -> ADD_SHARES`: keep quantity, clear price, commission, total, and funding account.
- switching to a non-security action clears a security that no longer belongs to the effective action.

This prevents old action state from silently surviving behind hidden form controls.

### 3. Validation matches the value that will actually be written

Updates now distinguish:

- **omitted key** → preserve stored value;
- **explicit `null`** → clear the value.

The previous nullish merge could validate the stored value while persisting the explicit `null`.

The branch introduces explicit “supplied or stored” semantics for validation and also validates resolved post-time overrides, because manual/cron/internal posting callers can bypass the controller DTO.

Consequences:

- zero BUY quantity is rejected;
- negative/zero amount-only investment totals are rejected;
- a provided FX rate must be positive;
- invalid post-time overrides are rejected before any financial write.

### 4. Stored FX is invalidated when the settlement basis changes

A stored `investmentExchangeRate` is preserved for presentation-only edits but cleared when the schedule's settlement basis changes and no replacement rate is supplied.

The relevant basis includes the action/account/security/effective funding-account structure that determines what currencies are being settled.

This closes the immediate “same stored rate, different scheduled settlement tuple” defect without blindly clearing legitimate rates on unrelated edits.

A broader residual remains when a **referenced security or account changes its own `currencyCode` while keeping the same ID**. See “Known follow-up: currency-pair provenance” below.

### 5. Posting now validates the generation it is about to commit

Several payloads are necessarily prepared before the schedule row is locked. The branch therefore verifies that the locked/current generation still describes the same financial operation before committing money.

The mutation basis covers:

- plain / transfer / investment shape;
- source/destination accounts;
- amount and currency;
- category where it shapes the posting;
- `originalAmount`, `originalCurrencyCode`, and `exchangeRate`;
- investment action, security, funding account, quantity, price, commission, total, and investment FX.

If those values changed between the pre-lock read and the row lock, posting returns a conflict instead of committing the stale payload.

### 6. Base split generations are checked separately

Parent scalar equality is not enough for a split scheduled transaction.

A child split can independently change:

- transfer destination;
- category;
- amount;
- embedded investment action/security/quantity/price/commission/FX;
- memo/tags relevant to that split generation.

When posting actually uses the base scheduled split set (rather than inline or occurrence-override splits), the service re-reads that set while holding the parent schedule lock and compares its semantic basis with the pre-lock generation.

A concurrent retarget such as:

```text
transfer split: Account A -> Account B
```

therefore causes a conflict rather than posting to Account A from stale data.

### 7. Occurrence overrides are re-read under the lock

The occurrence override is a mutable record of its own.

Posting now:

1. reads the candidate override before the transaction as needed for preparation;
2. locks/re-reads the override while the parent schedule lock is held;
3. compares semantic state plus `updatedAt`;
4. refuses on any meaningful change;
5. posts/removes the locked generation.

This prevents both stale posting and deleting a concurrent override edit that was never actually posted.

### 8. Cron eligibility is checked twice

Cron discovery still selects active auto-post schedules, but that selection is only a candidate snapshot.

For automatic posting, `post()` now receives `requireActiveAutoPost` and re-checks the locked row:

- `isActive === true`
- `autoPost === true`

If the user disabled the schedule after cron selected it, the automatic post is refused.

Manual posting intentionally does not use this restriction.

### 9. Loan split recalculation shares the parent serialization lock

Scheduled loan payments recalculate principal/interest children after posting. That writer previously modified child rows without locking the parent scheduled transaction.

This created a real interleaving:

```text
poster reads principal 700 / interest 300
poster locks parent
poster verifies child set
loan recalculator writes principal 720 / interest 280
poster posts stale 700 / 300
```

Because the parent total remains 1,000 in both versions, parent scalar checks alone cannot detect that reallocation.

`recalculateLoanPaymentSplits()` now:

1. takes `pessimistic_write` on the parent scheduled transaction;
2. re-reads the current child split set under that lock;
3. derives the current loan account from those rows;
4. recalculates and writes the children while still serialized by the parent.

Callers no longer pass a loan ID derived from an older snapshot.

### 10. `LINE_OF_CREDIT` follows the loan-payment recalculation contract

`LoanPaymentSetupService` supports:

- `LOAN`
- `MORTGAGE`
- `LINE_OF_CREDIT`

but the recalculator historically recognized only the first two.

The branch centralizes the discriminator in `LOAN_LIKE_ACCOUNT_TYPES` and uses it in both loan-account derivation sites.

Regression example:

```text
LOC balance:           910.00
annual rate:            12%
monthly payment:       100.00
previous principal:     90.00
previous interest:      10.00

periodic rate = 0.12 / 12 = 0.01

next interest  = 10.00 - 90.00 * 0.01 =  9.10
next principal = 100.00 - 9.10        = 90.90
```

The regression test asserts the next template is `90.90 / 9.10`, not the stale first-installment `90.00 / 10.00`.

---

## Commit-by-commit review history

| Commit | Review stage | Main purpose |
|---|---|---|
| `28a5d49` | Initial #1154 fix | Make funding-account routing action-authoritative across post/create/update/frontend |
| `7ce0b1e` | Claude Code internal review | Gate MNY import and centralize `FUNDING_ACCOUNT_ACTIONS`; add REINVEST coverage |
| `fe2b9be` | Claude Code internal review | Record the non-trivial stored-FX settlement-basis concern |
| `5cba485` | Claude Code internal review | Clear stale action-specific numeric fields |
| `31386b7` | Claude Code adversarial test review | Strengthen BUY keep-path regression assertion |
| `ffc05b8` | Claude Code end-to-end review | Correct the FX-hoisting / external-call documentation |
| `b2f7c1d` | ChatGPT re-review #1 | Post-time validation, explicit-null semantics, FX validity/invalidation, hidden security cleanup, locked investment row, frontend preview |
| `9339e14` | ChatGPT re-review #1 follow-up | Correct over-broad description of non-BUY/SELL settlement behavior |
| `a50fc77` | ChatGPT re-review #2 | Full schedule mutation guard, locked override generation, update concurrency, INTEREST frontend regression |
| `e709eb3` | ChatGPT re-review #3 | Child split generation guard, foreign FX tuple, live cron eligibility, semantic override comparison |
| `95eead2` | ChatGPT re-review #4 | Parent-lock loan recalculation, current-loan derivation, durable three-surface FX tracker |
| `0e0beca` | ChatGPT re-review #5 | Include `LINE_OF_CREDIT`; correct the accepted category-race documentation |

---

## Regression coverage added or strengthened

The branch adds targeted coverage for, among others:

- stale `DIVIDEND` funding account ignored at post;
- `SELL` still forwards the valid funding account;
- `REINVEST` ignores a stale funding account;
- create/update drop funding accounts for non-funding actions;
- MNY import does the same;
- BUY edits that do not touch funding do not write the column;
- action-specific numeric cleanup;
- explicit-null update semantics;
- positive investment FX;
- post-time negative/zero override rejection;
- locked-row investment posting;
- settlement-basis FX invalidation and presentation-only non-invalidation;
- INTEREST security cleanup and preservation where legitimate;
- schedule mutation conflicts;
- occurrence-override mutation conflicts;
- base split retarget conflicts;
- foreign-currency tuple conflicts;
- cron disabled-after-selection behavior;
- parent-lock loan recalculation;
- `LINE_OF_CREDIT` 90.90 / 9.10 recalculation.

The LOC test was also mutation-checked during review: removing `LINE_OF_CREDIT` from the discriminator makes the targeted test fail.

---

### Category reassignment can race after the point-in-time split comparison

`CategoriesService.reassignTransactions()` can update `ScheduledTransactionSplit.categoryId` without taking every affected scheduled parent lock.

Therefore a category-only write can still commit after the posting path's split comparison.

This residual is explicitly accepted because it affects categorization/reporting metadata, not the amount or account routing of the occurrence.

The code comment now describes this as a point-in-time guarantee rather than claiming that every unlocked writer is automatically caught.

### Presentation-only parent fields are not part of the financial mutation basis

Description/payee/tag changes can race with posting and one occurrence may carry the pre-lock presentation text.

Those fields are deliberately excluded from the financial mutation basis because they do not change account routing or monetary shape.

### Two-connection integration test remains desirable

Unit tests verify that the parent lock is requested and that current child state is used.

The strongest regression for the loan race would still be a real two-connection PostgreSQL test proving:

```text
A: post acquires parent lock
B: recalculation blocks on same parent
A: verifies/posts/commits
B: proceeds and reads the committed generation
```

That test is desirable but was treated as non-blocking rather than faking database serialization with mocks.

---

## Known follow-up: currency-pair provenance for persisted scheduled-investment FX

This PR fixes stale FX when the **scheduled settlement basis itself** changes.

A separate financial-correctness residual remains when a referenced security or account changes `currencyCode` while retaining the same ID.

A persisted rate is currently a scalar without durable `fromCurrency` / `toCurrency` provenance. The problem exists on three surfaces:

1. `scheduled_transactions.investment_exchange_rate`
2. `scheduled_transaction_splits.investment_exchange_rate`
3. occurrence override `splits[*].investment.exchangeRate`

Example:

```text
stored pair: EUR -> CAD
stored rate: 1.50

security is later changed to USD
current USD -> CAD rate: 1.35

quantity = 10
price    = 100
```

Reusing the stale scalar yields:

```text
10 * 100 * 1.50 = 1,500 CAD
```

while the current pair yields:

```text
10 * 100 * 1.35 = 1,350 CAD
```

Difference: **150 CAD / 11.11%**.

The preferred future fix is to persist the currency pair alongside every durable rate and treat a pair mismatch exactly like a missing rate. The narrower alternative is comprehensive invalidation of all three surfaces whenever a referenced security/account currency changes.

This residual is tracked separately and does not block #1154.

---

## Why this is safe to merge

The original #1154 money-misrouting scenario is closed at the authoritative posting boundary and at every normal write boundary.

The adversarial review then exercised adjacent failure classes:

- stale state left by action transitions;
- direct/internal callers bypassing DTO validation;
- pre-lock vs locked generation mismatch;
- mutable child split generations;
- mutable occurrence overrides;
- cron selection vs live state;
- post-commit loan-template mutation;
- loan-like account discriminator drift.

After five external re-review rounds, no unresolved `BLOCKER`, `HIGH`, or `MEDIUM` defect remains in the focused #1154 scope.

The remaining known items are either:

- explicitly accepted LOW content/categorization consistency trade-offs;
- stronger integration-test coverage;
- or the separately tracked FX currency-pair provenance problem.

---

## Verification

During the review cycle the branch was repeatedly checked with targeted scheduled-transaction/loan tests, TypeScript typechecking, ESLint, i18n parity/checks, and documentation-path validation.

The final review round reported the relevant scheduled-transaction suites and new LOC regression passing after the mutation check.

No frontend behavior change was made in the final LOC-only round; the frontend changes belong to the earlier funding-state, validation, and INTEREST-regression fixes.

---

## AI assistance disclosure

This PR was implemented with **Claude Code / Claude Opus 4.8**.

The implementation was then subjected to repeated external adversarial review by **ChatGPT**. Findings that survived source-level verification were implemented in subsequent Claude Code commits and accompanied by regression coverage.

The AI split is intentionally recorded in the tables above:

- Claude Code: implementation plus initial/internal review findings.
- ChatGPT: external re-review findings.
- Repository maintainer: owns the resulting behavior, review decisions, and merge.

### DIFF
--- a/backend/src/i18n/locales/de/errors.json
+++ b/backend/src/i18n/locales/de/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Terminierte Investmentbuchungen erfordern ein Maklerkonto",
     "notFound": "Terminierte Buchung mit ID {{ id }} nicht gefunden",
     "occurrenceAlreadyPosted": "Dieser Vorgang wurde bereits gebucht.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "Eine abweichende Währung ist nur bei einer einfachen terminierten Buchung möglich, nicht bei einer Aufteilung, Überweisung oder Investition",
     "fxRateUnavailable": "Für {{ from }} zu {{ to }} ist am {{ date }} kein Wechselkurs verfügbar. Geben Sie den Betrag in {{ to }} ein, um ihn zu buchen.",
     "investmentActionRequired": "Investmentaktion ist für terminierte Investmentbuchungen erforderlich",
--- a/backend/src/i18n/locales/en/errors.json
+++ b/backend/src/i18n/locales/en/errors.json
@@ -372,6 +372,7 @@
     "requiresBrokerageAccount": "Scheduled investment transactions require a brokerage account",
     "notFound": "Scheduled transaction with ID {{ id }} not found",
     "occurrenceAlreadyPosted": "This occurrence has already been posted.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "A different currency can only be used on a plain scheduled transaction, not a split, transfer or investment",
     "fxRateUnavailable": "No exchange rate is available for {{ from }} to {{ to }} on {{ date }}. Enter the amount in {{ to }} to post it.",
     "investmentActionRequired": "Investment action is required for scheduled investment transactions",
--- a/backend/src/i18n/locales/es/errors.json
+++ b/backend/src/i18n/locales/es/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Las transacciones de inversión planificadas requieren una cuenta de corretaje",
     "notFound": "Transacción planificada con ID {{ id }} no encontrada",
     "occurrenceAlreadyPosted": "Esta ocurrencia ya se ha registrado.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "Solo se puede usar otra moneda en una transacción planificada simple, no en un desglose, una transferencia ni una inversión",
     "fxRateUnavailable": "No hay un tipo de cambio disponible de {{ from }} a {{ to }} el {{ date }}. Introduce el importe en {{ to }} para registrarlo.",
     "investmentActionRequired": "Se requiere acción de inversión para las transacciones de inversión planificadas",
--- a/backend/src/i18n/locales/fr/errors.json
+++ b/backend/src/i18n/locales/fr/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Les transactions d'investissement récurrentes nécessitent un compte de courtage",
     "notFound": "Transaction récurrente avec l'ID {{ id }} introuvable",
     "occurrenceAlreadyPosted": "Cette occurrence a déjà été enregistrée.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "Une devise différente n'est possible que sur une transaction récurrente simple, pas sur une transaction fractionnée, un virement ou un investissement",
     "fxRateUnavailable": "Aucun taux de change n'est disponible de {{ from }} vers {{ to }} le {{ date }}. Saisissez le montant en {{ to }} pour l'enregistrer.",
     "investmentActionRequired": "Une action d'investissement est requise pour les transactions d'investissement récurrentes",
--- a/backend/src/i18n/locales/hi/errors.json
+++ b/backend/src/i18n/locales/hi/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "अनुसूचित निवेश लेनदेन के लिए ब्रोकरेज खाता आवश्यक है",
     "notFound": "ID {{ id }} वाला अनुसूचित लेनदेन नहीं मिला",
     "occurrenceAlreadyPosted": "यह घटना पहले ही दर्ज की जा चुकी है।",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "अलग मुद्रा केवल सामान्य अनुसूचित लेनदेन पर उपयोग की जा सकती है, विभाजन, स्थानांतरण या निवेश पर नहीं",
     "fxRateUnavailable": "{{ date }} को {{ from }} से {{ to }} के लिए कोई विनिमय दर उपलब्ध नहीं है। इसे पोस्ट करने के लिए राशि {{ to }} में दर्ज करें।",
     "investmentActionRequired": "अनुसूचित निवेश लेनदेन के लिए निवेश क्रिया आवश्यक है",
--- a/backend/src/i18n/locales/id/errors.json
+++ b/backend/src/i18n/locales/id/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Transaksi investasi terjadwal memerlukan akun pialang",
     "notFound": "Transaksi terjadwal dengan ID {{ id }} tidak ditemukan",
     "occurrenceAlreadyPosted": "Kejadian ini sudah dicatat.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "Mata uang lain hanya dapat dipakai pada transaksi terjadwal biasa, bukan pada transaksi terbagi, transfer, atau investasi",
     "fxRateUnavailable": "Tidak ada nilai tukar yang tersedia untuk {{ from }} ke {{ to }} pada {{ date }}. Masukkan jumlah dalam {{ to }} untuk mempostingnya.",
     "investmentActionRequired": "Tindakan investasi diperlukan untuk transaksi investasi terjadwal",
--- a/backend/src/i18n/locales/it/errors.json
+++ b/backend/src/i18n/locales/it/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Le transazioni di investimento pianificate richiedono un conto di intermediazione",
     "notFound": "Transazione pianificata con ID {{ id }} non trovata",
     "occurrenceAlreadyPosted": "Questa occorrenza è già stata registrata.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "Una valuta diversa può essere usata solo su una transazione pianificata semplice, non su una suddivisione, un trasferimento o un investimento",
     "fxRateUnavailable": "Non è disponibile alcun tasso di cambio da {{ from }} a {{ to }} il {{ date }}. Inserisci l'importo in {{ to }} per registrarla.",
     "investmentActionRequired": "L'azione di investimento è obbligatoria per le transazioni di investimento pianificate",
--- a/backend/src/i18n/locales/ja/errors.json
+++ b/backend/src/i18n/locales/ja/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "スケジュール投資取引には証券口座が必要です",
     "notFound": "ID {{ id }} のスケジュール取引が見つかりません",
     "occurrenceAlreadyPosted": "この回はすでに記帳されています。",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "別の通貨を使用できるのは通常のスケジュール取引のみで、分割、振替、投資では使用できません",
     "fxRateUnavailable": "{{ date }} の {{ from }} から {{ to }} への為替レートが利用できません。投稿するには金額を {{ to }} で入力してください。",
     "investmentActionRequired": "スケジュール投資取引には投資アクションが必要です",
--- a/backend/src/i18n/locales/ko/errors.json
+++ b/backend/src/i18n/locales/ko/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "예약 투자 거래에는 중개 계정이 필요합니다",
     "notFound": "ID가 {{ id }}인 예약 거래를 찾을 수 없습니다",
     "occurrenceAlreadyPosted": "이 발생 건은 이미 등록되었습니다.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
     "fxPlainOnly": "다른 통화는 일반 예약 거래에서만 사용할 수 있으며 분할, 이체, 투자에서는 사용할 수 없습니다",
     "fxRateUnavailable": "{{ date }}에 {{ from }}에서 {{ to }}로의 환율을 사용할 수 없습니다. 게시하려면 금액을 {{ to }}로 입력하세요.",
     "investmentActionRequired": "예약 투자 거래에는 투자 작업이 필요합니다",
--- a/backend/src/i18n/locales/nl/errors.json
+++ b/backend/src/i18n/locales/nl/errors.json
@@ -363,6 +363,7 @@
     "requiresBrokerageAccount": "Geplande beleggingstransacties vereisen een brokerage-rekening",
     "notFound": "Geplande transactie met ID {{ id }} niet gevonden",
     "occurrenceAlreadyPosted": "Deze keer is al geboekt.",
+    "changedConcurrently": "The scheduled transaction changed during the operation. Reload it and retry.",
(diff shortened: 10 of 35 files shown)


WMP:
@kenlasko : I thinking maybe we should make next .md based on findings in this pr description

PR #1182, opened 2026-08-17
Extend repository path guard to TypeScript and SQL comments

### LINKED ISSUE #1093: Extend the repository path guard to source and migration comments
## Problem

`backend/src/common/doc-paths.spec.ts` validates backticked paths in `CLAUDE.md` files and
top-level `docs/*.md`, but PR #1077 introduced four stale identifier/path claims in TypeScript and
SQL comments (wrong migration numbers, a renamed spec file, a moved grant location). The existing
guard cannot see them, and all four were caught only by a manual review pass.

## Scope

Evaluate scanning comments in:

```text
backend/src/**/*.ts
database/migrations/*.sql
```

Do not scan string literals or SQL/code syntax indiscriminately. Extract comments first, then apply
the existing rooted-versus-relative path rules. Migration-number references such as `migration 136`
need a separate check against filenames because they are not paths.

## Acceptance criteria

- the four corrected PR #1077 claims fail when reverted;
- legitimate code strings/templates/globs do not become false positives;
- exclusions carry reasons rather than path-specific allowlist entries;
- future-plan and historical-release exceptions retain their current semantics;
- test runtime remains acceptable.

WMP:
## Linked discussion / issue

Closes #1093
Approved in: https://github.com/kenlasko/monize/issues/1093

## Summary

Extends the existing repository path guard to validate path and migration references made in TypeScript and SQL source comments.

The guard now scans comments in:

- `backend/src/**/*.ts`
- `database/migrations/*.sql`

It reuses the repository-path grammar already used by the documentation guard and adds comment-aware parsing so strings, templates, regexes, SQL string literals, dollar-quoted bodies, and similar code constructs do not become false positives.

Migration references such as `migration 136`, `migrations 136 and 137`, and `sibling in 136` are checked against actual `NNN_*.sql` migration filenames.

### TypeScript comments

The TypeScript scanner:

- extracts line and block comments;
- ignores quoted strings;
- ignores raw template-literal text;
- ignores regex literal contents;
- recursively scans `${...}` interpolation bodies, including nested object expressions.

This means a real comment inside a template interpolation is checked, while comment-looking text in the template itself is not.

### SQL comments

The SQL scanner:

- extracts `--` and block comments;
- ignores ordinary single-quoted strings;
- handles doubled quote escaping;
- ignores PostgreSQL dollar-quoted and named dollar-quoted bodies;
- handles nested block comments;
- handles PostgreSQL `E'...'` strings, including backslash-escaped quotes.

This prevents `--` or path-looking text inside legitimate SQL literals from being interpreted as source comments.

### Compiled `.js` references

Source comments sometimes describe compiled runtime files such as `main.js` or `db-migrate.js` while the repository contains their TypeScript sources.

For source-comment validation only, `.js` references may therefore resolve to a corresponding `.ts` or `.tsx` source.

This behavior is opt-in and does **not** apply to documentation. The existing documentation guard retains exact-path semantics.

### PR #1077 regression coverage and scope boundary

The change was motivated by four stale comment references introduced in PR #1077.

The guard catches the renamed `currency-reference-columns.spec.ts` reference directly against the live tree.

The other three historical mistakes expose a deliberate limitation of an existence-based path guard:

- migrations `133` and `134` still exist, but now refer to unrelated migrations;
- `backend/src/db-init.ts` still exists, although the runtime grant referenced by the old comment moved elsewhere.

A literal revert of those comments therefore points at real-but-wrong targets. Detecting that would require semantic/content-level verification rather than repository-path existence checking.

The test suite now records this boundary explicitly against the live repository instead of hiding it behind synthetic fixtures.

This PR intentionally does not introduce a general semantic comment analyzer. If issue #1093 is closed with this implementation, its first acceptance criterion should be interpreted or updated to reflect this existence-vs-semantics boundary.

### Review follow-up

The implementation also incorporates all findings from an independent review of the initial version:

- PostgreSQL `E'...'` parsing was fixed;
- comments inside `${...}` interpolations are now scanned;
- `.js -> .ts/.tsx` fallback was restricted to source-comment validation;
- the existence-vs-semantics boundary is explicitly tested against the live tree.

No repository-specific path allowlist was introduced.

### Verification

Locally verified:

- lint;
- typecheck;
- both repository-path guard specs;
- 35 relevant tests passing.

The guard scans approximately 1,200 TypeScript files and 148 SQL migrations.

The full existing test suite has not been independently verified for this SHA, so that checklist item remains open until the complete required suite passes.

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity). No user-facing strings are introduced by this PR.

## AI assistance disclosure

This PR was developed with Claude Code assistance and independently reviewed with ChatGPT.

I reviewed and own the resulting implementation, tests, scope decisions, and documented limitations.

### DIFF
--- a/backend/src/backup/support-backup/support-backup-integrity.ts
+++ b/backend/src/backup/support-backup/support-backup-integrity.ts
@@ -532,7 +532,7 @@ function withSuffix(base: string, n: number, maxLen?: number): string {
  * Written as an escape, not as the byte itself: the literal control character
  * used to sit in this file, which made `grep`, `file` and every other text tool
  * classify the source as binary and skip it -- so a source-scanning guard test
- * silently stopped covering it. `src/test/source-bytes.spec.ts` now fails on any
+ * silently stopped covering it. `source-bytes.spec.ts` now fails on any
  * source file carrying a raw control byte.
  */
 const GROUP_SEPARATOR = "\u0000";
--- a/backend/src/common/db/runtime-role-check.spec.ts
+++ b/backend/src/common/db/runtime-role-check.spec.ts
@@ -303,7 +303,7 @@ describe("runtimeRoleViolations -- exempt role membership (DR-V2)", () => {
     //
     // A source scan, because the fix is the SHAPE of the predicate: the ownership
     // question has to be asked with the candidate as the subject, not the runtime
-    // role. `rls-elevation-and-role.integration.spec.ts` proves the behaviour.
+    // role. `runtime-role.integration.spec.ts` proves the behaviour.
     const reachableArm = RUNTIME_ROLE_FACTS_SQL.slice(
       RUNTIME_ROLE_FACTS_SQL.indexOf("AND pg_has_role(r.oid, h.oid, 'SET')"),
       RUNTIME_ROLE_FACTS_SQL.indexOf("AS exempt_reachable_contexts"),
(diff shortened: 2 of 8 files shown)


WMP:
Other issues i will do in next version, we are tired ;)

Issue #1190, opened 2026-08-17
Adding a cash transaction to investment account doesn't update holdings view immediately


bcorob:
From /accounts view, navigate to an investment account details page
Observe the Holdings by Account section and make note of any cash balance displayed
In Cash transactions, add a transaction, for example, a simple cash deposit
Observe the Holdings by Account section

Actual result:
Cash row is not updated/refreshed immediately.
Upon refreshing the page, the new balance is correctly displayed.

Expected result:
Upon making a change, any displayed balances or totals are immediately updated.

PR #1210, opened 2026-08-18
Investments: make the two registers, and the two pages that draw them, behave as one

kenlasko:
## Linked discussion / issue

Follow-up to #1190 — that issue was closed by #1196, which wired the account detail page's *create* path; this branch starts from the delete path it left behind. Everything after the first commit was requested directly by @kenlasko in the Claude Code session linked at the bottom, rather than in a discussion.

Closes: nothing outright (#1190 is already closed).
Approved in: direct request from the repository owner — no discussion thread exists, so the first checklist box below is left for @kenlasko to tick.

## Summary

**The two investment registers, and the two pages that draw them, made to behave as one thing.**

An investment account is two ledgers — trades on one, cash on the other — behind a toggle, and the same pair is drawn on the Investments page and on the account detail page. Each of the four combinations had drifted, so this branch works through them:

- **Every write refreshes the whole page.** The #1190 fix covered the detail page's create path; on the Investments page, adding a cash row refreshed the summary, the allocation, Holdings by Account and the brokerage register, while deleting one reloaded the cash rows alone — the row vanished and every figure above it kept its pre-delete value until Refresh was pressed. All write paths (create, edit, both deletes, status change, undo, redo, AI action) now go through one `refreshAfterWrite`, which also drives the value chart via `refreshKey`.
- **Both registers page from the strip above their rows** (`ListTopToolbar`, composed by `TransactionList` and `InvestmentTransactionList` instead of rebuilt). A pager below the table is one the reader meets only after scrolling past everything it could have helped them skip. It keeps the "Showing 1-7 of 7" line on a single page, which is exactly where a filter lands.
- **Both registers filter, each in its own terms** — a trade by symbol and action, a cash row by payee and category — and the options are the values those accounts' rows actually use, not the whole ledger's. Two new endpoints answer that; the cash one reads split lines as well as parent rows and carries the ancestors of each used category, because a tree picker drops a child whose parent is missing.
- **Both registers look the same**: one heading, a `+` on both new-row buttons, the same gap above the table.
- **A trade's cash leg is edited as the trade** on both, and a transfer is fetched in full first — the detail page had neither.

Two defects found along the way are worth calling out, because neither was visible from the tests:

- **The option pickers never received their options in dev.** Both hooks latched a ref *and* cancelled in the effect cleanup; under React's StrictMode the effect runs, is cleaned up, and runs again, so the first pass fired the only request and the cleanup discarded its answer. That is what "No options found" and an Action picker offering all twenty actions were. Testing Library does not double-invoke effects, so both hooks now carry a test that renders them inside `<StrictMode>`.
- **A filter scrolled the detail page to the top**, because the register answered the change with skeletons: the table collapsed, the page shortened, and the browser clamped the scroll. The skeleton now gates on there being nothing to show yet.

Also in the branch, outside the register theme (see the single-concern box below): the reconciliation reminder is unmounted from the header, and a nav tab is lit by the section a page belongs to (`/accounts/<id>` lights Accounts) rather than by an exact path match.

### Endpoints added

- `GET /transactions/filter-options` — the payees and categories used by the rows in the given accounts, with the ancestors of every used category. Delegates see only their readable accounts; an empty readable set is an empty picker, never an unfiltered query.
- `GET /investment-transactions/filter-options` — the actions (in the enum's order) and symbols (alphabetical) those rows use. Symbols come from the rows rather than current holdings, so a position sold in full is still offered — its trades are what a symbol filter is usually after. Both queries read VOID rows deliberately: the register lists them struck through, so a voided row must stay findable.

## Checklist

- [ ] An approved discussion or issue exists and is linked above. — *no discussion thread; the work was requested directly by the repository owner, who should confirm.*
- [ ] This PR addresses a **single concern** (large work is split into a series). — *no, and stated plainly: the register work is one theme, but the header badge removal and the nav-highlight change are not part of it. Happy to split those into their own PRs if preferred.*
- [ ] No shared/core areas were refactored without prior agreement. — *shared areas were touched: `TransactionList`'s toolbar was extracted into `ListTopToolbar` (six mount sites), `AppHeader`/`MobileNavDrawer` now share one nav-section predicate, and two backend endpoints were added. Details above; call it back if any of it should be separate.*
- [x] The branch is rebased on the latest `main` (2bab40d0).

### Verification

- Frontend: 700 files / 13,712 tests green (`npx vitest run`).
- Backend unit: 486 suites / 13,077 tests green (`npm run test:unit`), lint and `tsc --noEmit` clean on both layers.
- Backend **integration** suites need a live database and were not run in this environment — CI is the first place they execute.
- i18n parity green and the pseudo-locale regenerated; two now-unread keys removed from every locale, three added and translated in all eighteen full catalogues.
- Where a fix changed behaviour, the accompanying test was checked against the *previous* code and confirmed to fail — including the pager placement, the StrictMode option load, the symbol source, and both cash-row edit paths.

## AI assistance disclosure

Written by Claude Code (Opus 5) across one session, driven interactively by @kenlasko: each defect above was reported by hand after using the running app, and the branch was rebased on `main` mid-session at their request. Conventions and the reasoning behind each fix are recorded in `frontend/CLAUDE.md` alongside the code. The author remains responsible for correctness; the session transcript is linked below.

https://claude.ai/code/session_016khg9C8a3vMxe992pEQCSk


kenlasko:
**E2E shards 1, 3 and 4 failed on infrastructure, not on the branch.**

All three died in the same place, before any test body ran:

```
##[error]The action 'Install Playwright system dependencies' has timed out after 10 minutes.
```

`npm exec playwright install-deps chromium firefox` was still running when the runner killed the job (it shows up in each job's orphan-process cleanup), because apt could not reach the Azure Ubuntu mirror — the log is a wall of `Ign: http://azure.archive.ubuntu.com/ubuntu noble InRelease` before it fell back to `archive.ubuntu.com` and ran out of time. No spec executed and no report was produced: `No files were found with the provided path: e2e/playwright-report/`. The diagnostics comments above say the same thing from the other side — all three containers healthy, the backend fully booted, and the frontend serving nothing but `/login`.

A re-run is the fix here rather than a code change. `rerun_failed_jobs` on run 32165794140 currently returns `403 This workflow is already running` (shard 2 is still going); it will be retried once the run is idle, and I will report back if the shards fail a second time on anything other than that apt step.

---

PR #1225, opened 2026-08-19
Add "Accrued Interest" field to "Redeem CD/Bond" investment activity

### LINKED ISSUE #1217: Redeem CD/Bond investment activity has no way to specify accrued interest as part of the investment transaction
Import a Money file with CD or bond transactions covering a full lifecycle: purchase, pay out or reinvest interest, then sell/redeem, possibly with accrued interest.
Examine the Monize transaction for the redemption

Actual results:
Money "redeem CD/bond" is imported as a sell transaction, showing only the proceeds of the sale.  Observe that the reconciliation status of this investment transaction cannot be changed, as there is a note indicating this investment transaction is actually part of a split.
Upon examining the cash ledger, the sell transaction is revealed to be one leg of a split in the cash account, where the accrued interest is a separate split.  (Great, we didn't lose transaction fidelity!)
Note that even if changing the investment activity to "Redeem CD/Bond", there is nowhere in the investment edit form to specify the amount of accrued interest to be recorded.

Expected results:
A redemption activity should be a first-class investment activity, with accrued interest allowed to be specified directly in the investment activity similar to the commission.
The total value of the redemption, including accrued interest and minus any commission, should be reflected in the investment ledger.
There should be no requirement to have a parent split transaction appear in the cash ledger to contain both the accrued interest and the redeemed principal.  Similar to commission, accrued interest is just a part of this single atomic investment transaction.

All the above notwithstanding, accrued interest should be recorded and reportable like any other interest action.  Year-end tax reporting in particular needs to properly reflect this interest income as a discrete cash transaction.

(The more I dive into this, the more I recognize what an excellent job Money did at exposing the right UI to the user while still recording transactions in the proper way, abstracting away the underlying complexity to make it simple enough to use but preserving all the data in the correct manner for reports or other consumption later.  I'm not even convinced Monize's current method of storing a split cash transaction with interest as a separate leg is necessarily wrong, but the way the transaction is presented to the user in the cash and investment ledgers must be improved so that it feels and operates like a single investment transaction.)

bcorob:
## Linked discussion / issue

Closes #1217
Approved in: #1217 

## Summary

A Microsoft Money "Redeem CD/Bond" imported as a sale showing only the principal, with the accrued interest stranded in a sibling leg of a split in the cash ledger -- and no way to enter accrued interest by hand at all. A redemption is now one atomic investment activity that carries accrued interest beside commission, with no split parent required.

No schema change. The interest is persisted as a linked INTEREST companion transaction (`linked_transaction_id`, until now used only by transfer legs), which is what makes it income to every report with no special case. The companion writes no cash row of its own: the redemption's single cash transaction carries proceeds plus interest.

`total_amount` on the redemption stays proceeds. Folding the interest in would have reported it as capital gain on the tax surfaces this is meant to get right, so cost basis and realized gain are untouched -- asserted by a test that the gain with interest equals the gain without it.

That breaks the invariant every reader assumed, that a linked cash row is `|totalAmount| x rate`. `disposalCashAmount` is now the only place the two are added, and `accrued-interest.guard.spec.ts` fails a second copy.

REDEEM only. Money emits accrued interest on `act` 30 alone, so a SELL field would be manual-entry-only with no import evidence behind it. Refused, before any write, on a non-REDEEM action, a negative value, an embedded split row (whose parent leg is already the cash side), and a direct edit of the companion.

Import:
- `TRN_INV.amtInt` was parsed and dropped; it now becomes the companion, and `TRN.amt - amtInt` the proceeds.
- Money's two-leg split collapses into the single cash row the redemption adopts -- but only when the parent has exactly two legs, one of them the investment leg, the sibling equals `amtInt`, and the legs sum to the parent. Any other split keeps the shape Money wrote.
- `act` 30 already mapped to REDEEM, but coverage stopped at the code map; a writer-level regression test now asserts the persisted row.

Also carried through both assistant surfaces (AI tools and MCP) in the same change, and localized into every supported locale.

Spec: docs/specs/redemption-accrued-interest.md
Invariant: INV-REDEEM-001

## Checklist

- [X] An approved discussion or issue exists and is linked above.
- [X] This PR addresses a **single concern** (large work is split into a series).
- [X] New behavior has tests, and the existing suite passes.
- [X] All user-facing strings are translated for **every** locale (i18n parity).
- [X] No shared/core areas were refactored without prior agreement.
- [X] The branch is rebased on the latest `main`.
- [X] AI assistance is disclosed below, and I have reviewed and own the result.

## AI assistance disclosure

Co-authored with AI


backend/src/import/mny/map/map-investments.ts
```diff
@@ -235,11 +257,29 @@ function mapOne(
...
   const price = detail ? positiveOrNull(detail.price, PRICE_DECIMALS) : null;
   const commission = detail ? roundMoney(Math.abs(detail.commission)) : 0;
-  const totalAmount = totalAmountOf(row, action, quantity, price, commission);
-  const cashAmount = cashAmountOf(action, totalAmount);
-
-  return {
-    id: randomUUID(),
+  // Money records accrued interest on any investment detail row, but only a
+  // redemption pays it out; anywhere else there is no action to carry it.
+  const accruedInterest =
```
kenlasko, line 262: The companion is emitted whenever `amtInt > 0`, but the decision whether the cash split collapses happens later, in `map-transactions.ts` — and when it does *not* collapse (three legs, or a sibling whose amount differs from `amtInt`; your own `trade-cash-legs.spec.ts` cases at "keeps a split whose sibling is not the accrued interest" / "keeps a three-leg split"), both representations of the interest survive the import:

- the preserved interest split leg in the cash ledger, categorized as income there (today's behavior), **and**
- the linked INTEREST companion, which is income to every investment fold.

That breaks invariant 8 ("counted exactly once as income") on exactly the path section 5.2 promises to leave untouched.

It also creates the state invariant 6 exists to refuse: in the non-collapse shape the redemption is *embedded* (`transactionSplitId` set, asserted in the same spec), so the imported pair is an embedded REDEEM with a companion. Any later `update()` of that row that doesn't explicitly send `accruedInterest: 0` — including a description-only edit, because absent-means-unchanged re-reads the companion's amount — is refused by `assertAccruedInterestAllowed(..., isEmbedded=true)` with a message about a field the user never touched. Editing the companion instead is refused with a pointer back at the redemption. The only ways out are an explicit `accruedInterest: 0` (which the split editor never sends) or deleting the row.

Since the mapper can't see the collapse decision, this probably wants a reconciliation pass after `mapTransactions`: drop the companion (and the `linkedInvestmentId` back-link) for any trade whose parent split was kept, leaving the interest where Money wrote it. `totalAmount = TRN.amt - amtInt` can stay — proceeds are right for the gain fold either way.

---

backend/src/securities/investment-transactions.service.ts
```diff
@@ -3328,6 +3580,30 @@ export class InvestmentTransactionsService {
...
+      transaction.linkedTransactionId
+    ) {
+      const partner = await withScopedDb(this.dataSource, (m) =>
+        // includes VOID rows: records read -- the partner is loaded whatever
+        // its status, to decide whether this row is a companion at all.
+        m.getRepository(InvestmentTransaction).findOne({
+          where: { id: transaction.linkedTransactionId!, userId },
+        }),
+      );
+      if (partner && isAccruedInterestCompanion(transaction, partner)) {
```
kenlasko, line 3597: This refusal has no counterpart on the delete path. `remove()` (unchanged, below) treats any linked pair as a transfer: deleting either row removes both. Called with the *companion's* id, `linkedLeg` resolves to the redemption, so "delete this 87.50 interest row" silently deletes the whole redemption — reversing the 10,062.50 cash row and restoring 10 units of holdings — and records it in action history as "Deleted INTEREST transaction".

The register hides companions, but the id is reachable: `DELETE /investment-transactions/:id` directly, and the MCP/AI delete path, whose lists show the companion as an ordinary INTEREST row per invariant 9 (with no marker distinguishing it), so a confirmation card would read as approving an 87.50 deletion.

A direct *edit* of the companion is refused here with a pointer at the redemption; a direct *delete* deserves the same refusal (or "set the redemption's accrued interest to 0" semantics), not a cascade the caller can't see coming. Same shape as the CLAUDE.md rule: when you refuse something on one path, the bulk/AI/MCP routes to the same write need the refusal too.

---

kenlasko (commented):
Reviewed the full diff against the spec, the invariant catalog, and the surrounding services, and ran the suites locally. This is a well-built change — the two inline comments are the only defects I found, both on edge paths.

## What I verified (all green locally)

- **Backend**: the 7 touched suites plus `doc-paths`, the replay/deletion-balance guards, `transaction-split`, investment-reports, MCP, AI-action-builder, and the remaining import specs — 1,133 tests passing; `typecheck`, ESLint on changed files, `i18n:check`, and `locales.parity` all clean.
- **Frontend**: `investment-actions`, `InvestmentTransactionForm`, i18n parity and duplicate-key scans (1,559 tests), `tsc --noEmit`, ESLint, `i18n:check` — clean. No duplicate keys in any touched catalog.

## Design points that hold up

- **Proceeds stay proceeds.** `total_amount` on the REDEEM excludes the interest, and I confirmed the realized-gain fold in `investment-report-data.service.ts` measures `totalAmount` against cost basis while the companion lands in `state.income` via the ordinary INTEREST case — the "income with no special case" claim is real, not aspirational.
- **One addition site.** `disposalCashAmount` is genuinely the only place proceeds and interest are added, the guard spec scans for a second copy, and the frontend mirror (`redemptionTotalWithInterest`) is similarly single-sited.
- **The VOID boundary is closed on every door I could find.** `updateStatus` carries the companion across as a linked leg (both directions); voiding the *cash row* from the cash register is already refused by `assertVoidTransitionAllowedOnRow`; create-with-VOID writes a VOID companion and moves nothing. The status-only, edit, delete-of-redemption, cross-currency and future-dated rows of the truth table all have tests that assert balance deltas, not just calls.
- **Absent-means-unchanged** on `accruedInterest` is keyed on the value with an explicit-zero escape hatch, and the form always sends the field for REDEEM (including 0) — the "field presence is not a change" lesson applied correctly.
- **The register exclusion** removes the companion from the count as well as the page, so pagination can't split the pair; covered by a test.
- The refusal on dropping the action while interest remains (`update` → SELL with a live companion) is deliberate and tested; worth knowing the UI path there is "zero the field first, then change the action", which the error message implies but doesn't spell out.

## The two findings (inline)

1. **Import, non-collapse shape** (`map-investments.ts`): the companion is emitted before the collapse decision exists, so a kept split leaves the interest recorded twice (preserved split leg + companion) and produces an embedded-REDEEM-with-companion state that invariant 6 refuses at the API — which then blocks ordinary edits of that imported row. This is the one I'd fix before merge.
2. **Delete of the companion** cascades to the whole redemption via the transfer-pair logic in `remove()`, while edit of the same row is refused with a pointer. Lower severity (the register hides companions), but the MCP/AI delete path reaches it.

---

backend/src/import/mny/map/map-investments.ts
bcorob: This is now resolved in https://github.com/kenlasko/monize/pull/1225/commits/de43606098fa453a0690593d9c97b3151abffa92

backend/src/securities/investment-transactions.service.ts
bcorob: This is now resolved in https://github.com/kenlasko/monize/pull/1225/commits/de43606098fa453a0690593d9c97b3151abffa92

bcorob:
Updated PR according to feedback:

- Prevented duplicate interest companions for preserved Money splits.
- Refused direct companion deletion before any mutation.
- Added regression tests and updated contracts/guidance.

WMP:
My review, don't think that this is required to fix, this is only @kenlasko decision: 

# Review of PR #1225 — Accrued Interest for REDEEM Transactions

**Repository:** `kenlasko/monize`  
**Pull request:** #1225 — Add "Accrued Interest" field to "Redeem CD/Bond" investment activity  
**Reviewed head SHA:** `de43606098fa453a0690593d9c97b3151abffa92`  
**Base SHA:** `aa0e2be6a86fc1a1e5460bf989acf7c85df3ef7c`  
**Verdict:** **REQUEST CHANGES**

## Summary

The core model is sound: accrued interest is represented as a linked `INTEREST` companion while `REDEEM.total_amount` remains disposal proceeds, avoiding accidental treatment of interest income as capital gain.

However, the current implementation has several cross-layer correctness failures. The most important issues affect confirmation flows, undo/redo, concurrency, and status handling. These can cause the persisted financial state to differ from what the user approved, lose interest income after redo, duplicate cash/income under concurrent edits, or allow the hidden companion to mutate the redemption indirectly.

### Confirmed findings

| Severity | Count |
| --- | ---: |
| BLOCKER | 0 |
| HIGH | 4 |
| MEDIUM | 0 |
| LOW | 2 |

---

## HIGH-1 — AI and MCP confirmation paths drop accrued interest

**Severity:** HIGH  
**Confidence:** high

### Relevant paths

- `backend/src/ai/actions/ai-action.types.ts`
- `backend/src/ai/actions/ai-action-builder.service.ts`
- `backend/src/ai/actions/ai-actions.service.ts`
- `backend/src/ai/query/tool-executor.service.ts`
- `backend/src/mcp/tools/investments.tool.ts`

### Violated invariant

A confirmed financial action must persist the exact financial state shown in the preview.

For `INV-REDEEM-001`, the approved accrued interest must:

1. create the linked `INTEREST` companion;
2. be included in the single cash movement;
3. survive the preview -> signed descriptor -> confirmation -> DTO -> persistence path unchanged.

### Problem

The new tool schemas and preparation logic accept `accruedInterest`, but several confirmation paths fail to carry it through.

Examples include:

- singular create/update descriptors that do not contain `accruedInterest`;
- create confirmation DTO construction that omits it;
- singular/individual AI update preparation that does not pass it into `previewUpdateInvestmentTransaction`;
- direct MCP create/update writes that omit it;
- MCP `commitCard()` create/update paths that omit it;
- bulk create descriptors that do not preserve it.

This means the preview may show the correct combined cash amount while the confirmed write persists a redemption without accrued interest.

### Reproduction

Input:

- action: `REDEEM`
- quantity: `10`
- price: `1000.00`
- commission: `25.00`
- accrued interest: `87.50`

Expected:

- redemption proceeds: `10 * 1000 - 25 = 9,975.00`
- interest companion: `87.50`
- cash movement: `9,975.00 + 87.50 = 10,062.50`

Affected confirmation path can produce:

- redemption proceeds: `9,975.00`
- interest companion: missing
- cash movement: `9,975.00`

**Difference:**

- cash: `-87.50`
- interest income: `-87.50`

For an update where the only requested change is accrued interest, the AI singular path can instead treat the request as having no supported update field.

### Impact

The user can approve one financial result and receive another.

This is a financial-correctness and confirmation-integrity failure.

### Recommended correction

Carry `accruedInterest` through every layer:

- create/update preview;
- singular descriptor;
- bulk row descriptor;
- signed payload;
- AI confirmation executor;
- MCP relay path;
- MCP direct-confirm path;
- `commitCard()`;
- final DTO passed into `InvestmentTransactionsService`.

The descriptor should contain the resolved value that the user actually approved.

### Recommended regression tests

Use a **non-zero** accrued interest value such as `87.50`.

Cover at least:

- AI singular create;
- AI singular update;
- AI individual create/update;
- AI bulk create/update;
- MCP relay create/update;
- MCP direct create/update;
- MCP individual-card commit.

For each path assert:

1. preview cash amount is `10,062.50`;
2. signed descriptor contains `accruedInterest: 87.50`;
3. confirmation DTO contains `accruedInterest: 87.50`;
4. persisted companion exists exactly once;
5. final cash row is exactly `10,062.50`.

Tests using only `accruedInterest: 0` are insufficient because field loss remains invisible.

### Verification method

Trace one non-zero value through:

`tool input -> preview -> descriptor -> signed payload -> confirm executor -> DTO -> service -> DB rows`

and compare the final cash row with the approved preview.

---

## HIGH-2 — Undo/redo loses the accrued-interest companion

**Severity:** HIGH  
**Confidence:** high

### Relevant paths

- `backend/src/securities/investment-transactions.service.ts`
- `backend/src/action-history/action-history.service.ts`

### Violated invariant

A redemption with accrued interest is one durable financial event represented by:

- one `REDEEM` row;
- one linked `INTEREST` companion;
- one cash row.

Undo followed by redo must restore the same event.

### Problem

`InvestmentTransactionsService.create()` records the redemption in action history and captures the linked cash transaction for redo support.

It does **not** capture the accrued-interest companion.

The undo path can remove the linked companion because it follows `linkedTransactionId`.

The redo path restores a secondary investment row only when action history contains the transfer-specific `linkedTransferLeg` structure.

Therefore:

- initial create is correct;
- undo removes the complete event;
- redo restores the redemption and cash row;
- redo does not restore the accrued-interest companion.

### Reproduction

Initial event:

- redemption proceeds: `9,975.00`
- accrued interest: `87.50`
- cash row: `10,062.50`

After create:

- `REDEEM`: present
- `INTEREST` companion: present
- cash row: present

After undo:

- all removed

After redo:

- `REDEEM`: restored
- cash row: restored at `10,062.50`
- `INTEREST` companion: **missing**

Result:

- interest income: `0.00` instead of `87.50`
- cash ledger: `10,062.50`
- redemption read model can fall back to `accruedInterest = 0`

### Impact

Redo produces a state that did not exist before the undo.

The cash ledger and investment/income model disagree by `87.50`.

Any income report relying on the companion under-reports interest.

### Recommended correction

Action history must capture the full redemption pair.

Prefer a generic paired-investment representation rather than reusing a field named `linkedTransferLeg`, because `linked_transaction_id` now has two semantic uses.

Redo should:

1. restore cash transaction;
2. restore redemption;
3. restore companion;
4. recreate the mutual link;
5. rebuild affected holdings;
6. preserve status, exchange rate, date, amount, security and account.

### Recommended regression test

Create:

- `REDEEM`
- proceeds `9,975.00`
- accrued interest `87.50`

Then:

1. assert two investment rows + one cash row;
2. undo;
3. assert all three are absent;
4. redo;
5. assert exactly two investment rows + one cash row;
6. assert mutual links;
7. assert cash amount `10,062.50`;
8. assert derived `accruedInterest = 87.50`;
9. assert interest income includes exactly `87.50`.

### Verification method

Inspect DB rows and run the relevant register/income read paths after redo.

---

## HIGH-3 — Concurrent redemption edits can duplicate interest and cash effects

**Severity:** HIGH  
**Confidence:** high

### Relevant paths

- `backend/src/securities/investment-transactions.service.ts`
- `backend/src/common/db/locks.ts`
- `database/migrations/078_investment_transaction_transfer_link.sql`

### Violated invariant

`INV-REDEEM-001` claims one redemption has:

- one companion;
- one cash row;
- one financial effect.

The documented concurrency scope is per redemption pair.

### Problem

The generic investment `update()` path reads the transaction before entering the write transaction.

Inside the transaction it resolves the existing companion without first locking and reloading the redemption row.

There is also no database uniqueness constraint making `linked_transaction_id` one-to-one.

The repository already contains `lockInvestmentTransactionRow(... FOR UPDATE)` specifically to prevent financial mutations derived from stale transaction snapshots, but this path does not use it.

Two concurrent requests can therefore derive their writes from the same stale state.

### Reproduction scenario

Start with:

- `REDEEM.total_amount = 9,975.00`
- no companion
- cash row = `9,975.00`

Two concurrent requests:

- request A sets accrued interest to `50.00`
- request B sets accrued interest to `75.00`

Both read the original redemption before either commit.

Both can observe:

- no existing companion;
- the same original cash transaction ID.

Possible interleaving:

1. A removes the original cash side.
2. A creates:
   - companion `50.00`;
   - cash row `10,025.00`;
   - updates redemption links.
3. B proceeds from its stale in-memory redemption.
4. B clears/reverses using the old transaction state.
5. The old original cash row is already gone, so the stale reversal does not remove A's newly created cash row.
6. B creates:
   - another companion `75.00`;
   - another cash row `10,050.00`;
   - replaces the redemption link with B's companion.

Potential result:

- active cash rows: `10,025.00 + 10,050.00 = 20,075.00`
- expected cash: `10,050.00`
- **cash overstatement: +10,025.00**

Interest rows:

- `50.00 + 75.00 = 125.00`
- expected: `75.00`
- **interest overstatement: +50.00**

The first companion can also become orphaned from the redemption.

### Impact

A concurrent edit can corrupt both cash and income state.

This is particularly serious because the PR marks the invariant as enforced while no pair-level serialization mechanism exists.

### Recommended correction

Inside the same `withScopedDb` transaction:

1. lock the redemption row with `lockInvestmentTransactionRow`;
2. reload the authoritative mutable state after the lock;
3. derive `transactionId`, `linkedTransactionId`, status, date and amount from the locked row;
4. lock the existing companion when present;
5. only then reverse and reapply effects.

A uniqueness constraint preventing multiple companions from pointing to the same redemption can be added as defense in depth, but it does not replace locking because it does not prevent stale cash reversal.

### Recommended regression test

Use a PostgreSQL integration/concurrency test with a barrier:

1. create one redemption without interest;
2. start two PATCH operations simultaneously;
3. force both to overlap before mutation;
4. allow both to complete;
5. assert:
   - one redemption;
   - exactly one valid companion;
   - exactly one linked cash row;
   - no orphan cash row;
   - no orphan companion;
   - cash balance equals the winning final state exactly once.

### Verification method

Run against PostgreSQL rather than repository mocks and inspect row counts plus the account balance.

---

## HIGH-4 — Status handling treats the accrued-interest link as if it were always a transfer

**Severity:** HIGH  
**Confidence:** high

### Relevant paths

- `backend/src/securities/investment-transactions.service.ts`
- `backend/src/securities/investment-transactions.controller.ts`

### Violated invariant

The specification states:

> Two rows, one event, one status.

The companion must also be independently immutable.

### Problem A — direct companion status mutation is not refused

`updateStatus()` does not first determine whether the requested row is an accrued-interest companion.

On a `VOID` boundary crossing it follows `linkedTransactionId` using logic written for linked transfer legs.

Therefore a direct status call on the hidden companion can also mutate the redemption.

### Reproduction A

Create:

- redemption proceeds: `9,975.00`
- interest companion: `87.50`
- cash movement: `10,062.50`
- quantity redeemed: `10`

Obtain the companion ID from security history.

Call:

`PATCH /investment-transactions/{companionId}/status`

with:

`VOID`

Expected:

- reject direct companion mutation;
- no financial state changes.

Actual path can:

- VOID the companion;
- follow its link;
- VOID the redemption;
- reverse the `10,062.50` cash effect;
- restore 10 shares.

A hidden implementation row becomes an indirect command surface for the full redemption.

### Problem B — non-VOID reconciliation statuses do not stay synchronized

If old and new status are both on the same side of `VOID`, the method returns early after updating only the requested row.

Example:

`UNRECONCILED -> CLEARED`

on the visible redemption:

- redemption becomes `CLEARED`;
- companion remains `UNRECONCILED`.

This directly violates the PR's stated one-status invariant.

### Impact

The hidden companion can unexpectedly control the parent event, and the pair can persist contradictory statuses.

### Recommended correction

Before generic linked-leg handling:

1. identify whether the requested row is an accrued-interest companion;
2. reject direct companion status mutation;
3. if the requested row is a `REDEEM`, resolve its companion;
4. synchronize the companion on **every** status change;
5. preserve the existing financial-effect logic for `VOID` boundary transitions;
6. keep transfer-specific behavior separate from redemption-companion behavior.

### Recommended regression tests

Cover:

1. direct companion `-> VOID` returns `400`;
2. cash and holdings stay unchanged after the refusal;
3. redemption `UNRECONCILED -> CLEARED` updates both rows;
4. redemption `CLEARED -> RECONCILED` updates both rows;
5. redemption `active -> VOID` updates both rows and reverses cash/holdings once;
6. redemption `VOID -> active` reapplies cash/holdings once.

### Verification method

Use endpoint integration tests and assert:

- both investment row statuses;
- cash transaction status/effect;
- holdings quantity;
- final account balance.

---

## LOW-1 — Register filter options expose a hidden companion action

**Severity:** LOW  
**Confidence:** high

### Relevant path

`backend/src/securities/investment-transactions.service.ts`

### Violated behavior

The filter-options method states that the available filter values should represent the rows the register can actually display.

`findAll()` excludes accrued-interest companions.

`getRegisterFilterOptions()` does not apply the same exclusion.

### Reproduction

Data:

- one `REDEEM` with accrued-interest companion;
- no standalone `INTEREST` investment transaction.

Result:

- filter options include `INTEREST`;
- selecting `INTEREST` returns an empty register because the only such row is the hidden companion.

### Impact

Misleading UI state.

No financial corruption.

### Recommended correction

Share the companion-exclusion predicate between:

- register query;
- filter-options query.

### Recommended regression test

1. companion-only interest must not introduce the `INTEREST` filter;
2. adding one genuine standalone `INTEREST` transaction must introduce it.

### Verification method

Run both endpoints against the same integration fixture.

---

## LOW-2 — Security history exposes an Edit action that the backend always rejects

**Severity:** LOW  
**Confidence:** high

### Relevant paths

- `frontend/src/components/securities/SecurityTransactionHistory.tsx`
- `backend/src/securities/investment-transactions.service.ts`

### Violated behavior

The new model says the accrued-interest companion is not independently mutable.

Security history still renders it as an ordinary `INTEREST` transaction and unconditionally shows an Edit button.

The backend then rejects the edit.

### Reproduction

1. Open the security history for a redeemed CD/bond.
2. Locate the linked accrued-interest `INTEREST` row.
3. Click **Edit**.
4. Change any field and submit.

The edit form is presented normally, but the backend rejects the request with the companion-locked validation error.

### Impact

The UI advertises an action that cannot succeed.

### Recommended correction

Expose enough metadata in the security-history read model to identify an accrued-interest companion.

Then either:

- hide/disable Edit for that row; or
- route Edit to the parent redemption.

### Recommended regression test

A companion row must not expose an independent edit action.

### Verification method

Component test with a companion fixture plus backend integration coverage.

---

## Previously Reported Issues Rechecked

Two issues that appeared in earlier PR review comments were rechecked against the current head and are **not** retained as open findings:

1. **Money import double-counting when the redemption remains embedded in a split**  
   The later reconciliation logic now removes the generated companion/back-link when the preserved sibling split already represents the interest.

2. **Direct `remove(companionId)` deleting the pair**  
   The current `remove()` implementation identifies an accrued-interest companion and rejects direct deletion before financial mutation.

---

## Recommended Merge Gate

Before merging, at minimum:

1. Fix every preview -> descriptor -> confirmation propagation path for `accruedInterest`.
2. Add create/undo/redo coverage for a redemption with non-zero accrued interest.
3. Serialize redemption edits using a locked authoritative row inside the transaction.
4. Separate transfer-link status behavior from redemption-companion status behavior.
5. Add integration tests using a non-zero accrued-interest value.
6. Add one real PostgreSQL concurrency test for simultaneous accrued-interest edits.

The implementation should not mark `INV-REDEEM-001` as fully enforced until the pair-level concurrency behavior is mechanically enforced and regression tested.

---

## Review Limitations

- The review was read-only.
- No GitHub review, comment, commit, branch, or repository file was modified.
- The inspected PR head was `de43606098fa453a0690593d9c97b3151abffa92`.
- GitHub CI for the reviewed head was reported green during review.
- The findings above were established by code-path inspection.
- No separate local PostgreSQL instance was executed during this review.
- The concurrency finding should be closed only after a production-boundary PostgreSQL test demonstrates pair-level serialization.

PR #1226, opened 2026-08-20
Document and enforce critical system invariants

WMP:
## Linked discussion / issue

Closes #
Approved in: #1227

## Summary

This PR turns the system-invariant audit into explicit, enforceable repository contracts: the critical invariants are documented, the verification requirements are defined, concrete gaps found during the audit are fixed, and load-bearing regression tests guard the mechanisms that make those invariants true.

The main changes are:

- document the repository's core system invariants in `docs/system-invariants.md`, including their source of truth, enforcement mechanism, concurrency/retry semantics, failure behavior, required verification, and current status;
- align `docs/verification-contract.md` with the implemented test suite and define the load-bearing verification kind for each invariant;
- extend the strict reconciled-transaction lock to mutation paths that could previously bypass it through action-history undo/redo and embedded investment updates;
- enforce creator-only deletion of shared custom currency definitions while still allowing other users to remove their own currency preference;
- add a deterministic two-connection balance test proving that an absolute recompute cannot overwrite a concurrent atomic balance delta;
- add a deterministic two-connection holding test that sends concurrent BUYs through the real `InvestmentTransactionsService.create` path and verifies the stored holding against a replay of persisted `investment_transactions`;
- add a PostgreSQL failpoint test for logout revocation, proving that a database failure propagates instead of being presented as a successful logout;
- extend source-scan guards for reconciled-lock coverage and add an invariant-catalog parity guard that rejects missing, extra, or duplicate invariant IDs between the catalog and verification matrix;
- correct stale verification documentation, including the removed `--passWithNoTests` behavior and the now-present load-bearing source scans;
- record previously known-wrong tests as corrected where the current suites now enforce the intended behavior.

Verification performed for the final changes:

- backend typecheck passes;
- backend lint passes;
- the targeted PostgreSQL integration suites for balance recompute, concurrent holding updates, and logout revocation failpoints pass;
- invariant/source-scan guards and invariant-catalog parity tests pass;
- the new load-bearing tests were also validated negatively by removing or bypassing the protected mechanism and confirming that the corresponding test fails for the intended reason.

The branch has been rebased onto the current `main`.

## Checklist

- [x] All user-facing strings are translated for **every** locale (i18n parity). No new user-facing strings are introduced by this PR.

## AI assistance disclosure

Claude Code was used to perform the invariant audit, implement the fixes, and author or refine the associated tests and documentation. ChatGPT was used for an independent review of the branch, follow-up verification of review fixes, and drafting this pull request description. I reviewed the resulting changes and take responsibility for their correctness and scope.

### DIFF
--- a/backend/src/action-history/action-history.service.ts
+++ b/backend/src/action-history/action-history.service.ts
@@ -10,6 +10,7 @@ import { Cron } from "@nestjs/schedule";
 import { ActionHistory } from "./entities/action-history.entity";
 import { Transaction } from "../transactions/entities/transaction.entity";
 import { TransactionSplit } from "../transactions/entities/transaction-split.entity";
+import { assertReconciledRowsMutable } from "../transactions/reconciled-lock.util";
 import { Account } from "../accounts/entities/account.entity";
 import { Category } from "../categories/entities/category.entity";
 import { Payee } from "../payees/entities/payee.entity";
@@ -546,6 +547,7 @@ export class ActionHistoryService {
 
     const transaction = await manager.findOne(Transaction, {
       where: { id: action.entityId, userId: action.userId },
+      lock: { mode: "pessimistic_write" },
     });
     if (!transaction) {
       throw new ConflictException(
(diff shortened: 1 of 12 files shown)


kenlasko:
There's a merge conflict I can't resolve. Can you rebase on main again?

PR #1230, opened 2026-08-20
Fixed the 400 error in RecurringChargesPanel.tsx:115.

### LINKED ISSUE #1229: Console error: [RecurringChargesPanel] "Failed to load recurring charges:" AxiosError: Request failed with status code 400
I see this console error constantly show up when navigating amongst account pages in an account "full" of legacy transactions.
Investigating now.

Image

bcorob:
## Linked discussion / issue

Closes #1229
Approved in: Not approved yet

## Summary

Cause: the panel requested 500 transactions, but the backend limit is 200. It now uses the existing paginated getAllPages() helper.

## Checklist

- [X] This PR addresses a **single concern** (large work is split into a series).
- [X] All user-facing strings are translated for **every** locale (i18n parity).
- [X] No shared/core areas were refactored without prior agreement.
- [X] The branch is rebased on the latest `main`.
- [X] AI assistance is disclosed below, and I have reviewed and own the result.

## AI assistance disclosure

Co-authored with AI

### DIFF
--- a/frontend/src/app/accounts/[id]/page.test.tsx
+++ b/frontend/src/app/accounts/[id]/page.test.tsx
@@ -93,6 +93,7 @@ vi.mock('@/components/accounts/shared/ForeignCurrencyFeesSection', () => ({
 }));
 
 const mockGetAllTransactions = vi.fn();
+const mockGetAllPages = vi.fn();
 const mockGetSummary = vi.fn();
 const mockGetMonthlyTotals = vi.fn();
 const mockGetGroupedTotals = vi.fn();
@@ -100,6 +101,7 @@ const mockGetRecurringCharges = vi.fn();
 vi.mock('@/lib/transactions', () => ({
   transactionsApi: {
     getAll: (...args: unknown[]) => mockGetAllTransactions(...args),
+    getAllPages: (...args: unknown[]) => mockGetAllPages(...args),
     getSummary: (...args: unknown[]) => mockGetSummary(...args),
     getMonthlyTotals: (...args: unknown[]) => mockGetMonthlyTotals(...args),
     getGroupedTotals: (...args: unknown[]) => mockGetGroupedTotals(...args),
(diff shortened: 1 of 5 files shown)


WMP:
Finding for the maintainer
MEDIUM — confidence: medium — Unbounded payee-ID query can make recurring-charge detection fail again on high-cardinality accounts
Location: frontend/src/components/accounts/shared/RecurringChargesPanel.tsx:115-124. Supporting path: frontend/src/lib/transactions.ts, transactionsApi.getRecurringCharges.
Type: confirmed unbounded request construction; deployment-specific failure threshold.
Violated invariant / expected behavior: Opening the recurring-charges panel must remain functional as an account's transaction and payee cardinality grows. The fix removes the invalid 500-row request, but it replaces the bounded first-page input with an unbounded set of payee IDs.
The new implementation loads every transaction in the 12-month window, collects every distinct payeeId, then passes the entire set to getRecurringCharges().  getRecurringCharges() serializes that array into a single comma-separated GET query parameter, while the backend parseUuids() has no cardinality bound.
Realistic reproduction: Populate a credit-card account with roughly 250 distinct linked payees during the last 12 months. UUIDs alone require about 9.2 KB before the rest of the URL and encoding overhead. getAllPages() succeeds, but the subsequent /transactions/recurring-charges?payeeIds=... request can exceed the request-line/URI limit of a proxy or HTTP server. The component catches that failure and converts it to [], so the user sees no detected recurring charges.
Impact: The PR fixes #1229 for accounts with many transactions but can reintroduce effectively the same user-visible failure for accounts with many distinct payees. This is particularly relevant because #1229 was reported on an account containing substantial legacy history.
Recommended correction: Do not enumerate transaction/payee IDs in the browser. Extend the recurring-charges API to accept the accountId and date range, authorize that account using the same own/joint/delegated-account semantics as the register, and apply the account predicate inside TransactionAnalyticsService.getRecurringCharges(). That produces one bounded semantic request and avoids downloading all transaction objects solely to derive IDs. Chunking payeeIds or moving them to a POST body would avoid the URL limit, but server-side account filtering is the cleaner fix.
Recommended regression test: Cover an account with more than one transaction page and high payee cardinality. Assert recurring-charge detection succeeds without sending an ID list whose request size grows with transaction/payee count. Also retain the existing getAllPages() unit coverage if that helper remains used elsewhere.
Verification: Inspect the resulting network request and confirm its size is constant with respect to the number of distinct payees; then run the frontend unit suite plus an API/E2E case containing >200 transactions.

PR #1232, opened 2026-08-20
Issue #1167 - Scheduled Investment FX Provenance Review Summary

### LINKED ISSUE #1167: Scheduled investment FX rates can become stale when the referenced currency pair changes
## Summary

Scheduled investment flows can persist an explicit FX rate without persisting the currency pair that rate belongs to. If the referenced security or settlement account later changes currency while keeping the same ID, the stored rate can silently become a rate for the wrong pair.

The problem applies to all three persistent scheduled-investment rate surfaces:

1. `scheduled_transactions.investment_exchange_rate`
2. `scheduled_transaction_splits.investment_exchange_rate`
3. occurrence override embedded investment rates in `scheduled_transaction_overrides.splits[*].investment.exchangeRate`

At posting time, a positive stored rate is treated as authoritative before the current security/cash currency pair is resolved. As a result, a later scheduled occurrence can post the wrong cash amount even though the referenced entity IDs are unchanged.

This is a deferred follow-up to issue #1154. The #1154 remediation correctly clears a stored rate when the schedule's own settlement structure changes, but it cannot detect a currency change made directly on the referenced `Security` or `Account` row.

## Severity

**MEDIUM**

The trigger is uncommon because it requires a currency change after scheduling, but once triggered it can produce an incorrect future cash posting and therefore an incorrect account balance.

## Affected areas

- `backend/src/scheduled-transactions/scheduled-transactions.service.ts`
- `backend/src/scheduled-transactions/entities/scheduled-transaction.entity.ts`
- `backend/src/scheduled-transactions/entities/scheduled-transaction-split.entity.ts`
- `backend/src/scheduled-transactions/entities/scheduled-transaction-override.entity.ts`
- `backend/src/securities/investment-transactions.service.ts`
- `backend/src/securities/securities.service.ts`
- `backend/src/accounts/accounts.service.ts`
- `docs/future-plans/scheduled-investment-fx-currency-pair.md`

## Expected invariant

A persisted FX rate may only be reused when it is known to belong to the **current** settlement pair:

`source currency -> cash/settlement currency`

Changing either side of that pair must make the old rate unavailable for future posting unless the rate is explicitly re-entered for the new pair.

### Example: scheduled embedded investment split

Initial state:

- security ID: `SEC-1`
- security currency: `EUR`
- settlement cash currency: `CAD`
- quantity: `10`
- price: `100 EUR`
- stored scheduled split FX rate: `1.50 EUR/CAD`

Expected cash impact at this point:

`10 x 100 x 1.50 = 1,500 CAD`

Now change the currency of the same security ID:

- `SEC-1.currencyCode`: `EUR -> USD`

Assume the correct current `USD/CAD` rate is `1.35`.

The schedule still contains the old scalar rate `1.50`. Because the security ID did not change, no existing schedule-level settlement-basis invalidation fires.

On the next posting:

- **Implementation result:** `10 x 100 x 1.50 = 1,500 CAD`
- **Expected result:** `10 x 100 x 1.35 = 1,350 CAD`
- **Difference:** `150 CAD`
- **Percentage error:** `11.11%`

The same defect shape applies when the settlement account currency changes while its account ID remains unchanged.

## Why the current behavior occurs

`InvestmentTransactionsService.resolveCashExchangeRate()` accepts a positive supplied rate as authoritative before resolving the current source and settlement currencies.

That behavior is valid for an explicit rate entered for the current transaction, but a persisted scheduled rate has no stored `fromCurrency` / `toCurrency` provenance. After a referenced currency changes, the service cannot distinguish:

- a valid user-supplied rate for the current pair; from
- a stale rate that originally belonged to a different pair.

The #1154 update path already invalidates rates when schedule-owned settlement identifiers change. The missing case is a currency mutation on an already-referenced entity with the same ID.

## Required scope

Any fix must cover all three persistent rate surfaces together.

### 1. Parent scheduled investment

`scheduled_transactions.investment_exchange_rate`

### 2. Embedded scheduled investment split

`scheduled_transaction_splits.investment_exchange_rate`

### 3. Occurrence override investment split

`scheduled_transaction_overrides.splits[*].investment.exchangeRate`

A fix that handles only the parent schedule still leaves the same corruption path available through base splits or occurrence overrides.

### Preferred: persist FX pair provenance

Store the rate together with the pair it belongs to, for example:

- `investment_exchange_rate_from_currency`
- `investment_exchange_rate_to_currency`

or equivalent structured provenance for the JSON override payload.

At posting time:

1. resolve the current source currency from the current security / investment account;
2. resolve the current settlement currency from the effective cash/funding account;
3. compare the current pair with the pair stored alongside the scheduled rate;
4. reuse the stored rate only when the pair still matches;
5. otherwise treat the stored rate as absent and resolve a fresh rate for the current pair.

This makes the rate self-describing and prevents every security/account write path from having to know how to invalidate scheduled rates.

### Acceptable narrower fix: invalidate on currency mutation

When a referenced security or account changes `currencyCode`, atomically clear affected persisted scheduled rates across all three surfaces.

At minimum:

- `SecuritiesService.update()` must invalidate rates for schedules/splits/overrides that reference the changed security;
- `AccountsService.update()` must invalidate rates for schedules/splits/overrides whose settlement basis includes the changed account, including linked brokerage/cash pairs and explicit funding accounts.

This is simpler but more fragile because every future write path that can change settlement currencies must preserve the invalidation invariant.

## Acceptance criteria

- [ ] A stored scheduled rate is never reused for a different current currency pair.
- [ ] Parent scheduled investment rates are covered.
- [ ] Embedded scheduled investment split rates are covered.
- [ ] Occurrence override embedded investment rates are covered.
- [ ] Changing a security currency with the same security ID invalidates or supersedes the old rate.
- [ ] Changing a settlement account currency with the same account ID invalidates or supersedes the old rate.
- [ ] Linked investment brokerage/cash account currency changes are covered.
- [ ] Explicit funding-account currency changes are covered.
- [ ] A presentation-only edit that does not change the effective pair preserves a valid stored rate.
- [ ] Same-currency settlement continues to resolve to `1` and does not accidentally reuse an unrelated stored cross-currency rate.
- [ ] Posting after a pair change either resolves a valid current rate or fails explicitly if no rate is available; it must never silently use the stale scalar.

## Regression tests

Add coverage for each persisted surface.

### Parent schedule

1. Create a scheduled investment with security `EUR`, cash `CAD`, stored rate `1.50`.
2. Change the security currency to `USD` without changing the security ID.
3. Make the resolver return `1.35 USD/CAD`.
4. Post the occurrence.
5. Assert cash impact uses `1.35`, not `1.50`.

### Scheduled investment split

Repeat the same scenario with the rate stored in `scheduled_transaction_splits.investment_exchange_rate`.

### Occurrence override split

Repeat the same scenario with the rate stored in `scheduled_transaction_overrides.splits[*].investment.exchangeRate`.

### Settlement-account currency change

For each applicable surface:

1. keep the referenced security unchanged;
2. change the effective cash/funding account currency while keeping the account ID unchanged;
3. post the occurrence;
4. assert the old pair's rate is not reused.

### Preservation test

Edit only a non-financial field such as the schedule name/description and confirm that a still-valid stored rate for the same pair is preserved.

## Verification

The fix should be verified by tracing the complete posting path:

`scheduled transaction / split / override -> effective investment payload -> InvestmentTransactionsService.resolveCashExchangeRate() -> linked cash transaction -> account balance`

The decisive assertion is the currency and numeric amount of the linked cash transaction after the referenced currency pair changes.

## Notes

This issue is intentionally separate from #1154. The #1154 remediation fixed stale funding/account routing and schedule-owned settlement-basis changes. This follow-up covers the remaining case where the referenced entity keeps the same ID but its currency changes, making an otherwise unchanged stored FX scalar semantically stale.

WMP:
**Issue:** `kenlasko/monize#1167` - *Scheduled investment FX rates can become stale when the referenced currency pair changes*  
**Reviewed branch:** `claude/issue-1167-implementation-073j4q`  
**Final reviewed head:** `862eeefc83c983bc49363ce65584116e2772822f`  
**Main at final review:** `876ea0510247fd2a12a3e13bc285db3eb0f353dc`  
**Branch state at final review:** 19 commits ahead, 0 behind `main`  
**Final review disposition:** **APPROVE**  
**Remaining confirmed findings:** 0 BLOCKER / 0 HIGH / 0 MEDIUM

## 1. Review scope

This review followed issue #1167 through repeated implementation and re-review cycles. The core invariant was:

> A persisted scheduled-investment FX rate may be reused only when it is proven to belong to the current source-currency -> settlement-currency pair. A same-currency settlement always resolves to `1`.

The review covered all three persistent scheduled-investment FX surfaces:

1. `scheduled_transactions.investment_exchange_rate`;
2. `scheduled_transaction_splits.investment_exchange_rate`;
3. `scheduled_transaction_overrides.splits[*].investment.exchangeRate`.

The review also traced the important downstream path where applicable:

`scheduled transaction / override / split -> effective posting payload -> InvestmentTransactionsService -> linked cash transaction / embedded investment -> account cash impact`

The branch was reviewed read-only. No repository files, branches, pull requests, reviews, comments, labels, or settings were modified by the reviewer.

## 2. Final behavior verified

The final implementation establishes the following behavior:

- Persisted investment FX rates carry `fromCurrency` / `toCurrency` provenance.
- A stored rate with a pair that no longer matches the current settlement pair is never reused.
- A legacy stored scalar without provenance remains unknown and is re-resolved rather than re-blessed on a cosmetic edit.
- A presentation-only edit preserves a still-valid pinned rate.
- Parent schedules, embedded scheduled investment splits, occurrence override splits, and inline Post-dialog splits use the same effective-rate semantics.
- Split identity is correlated by stable server identity rather than by `securityId + rate` value matching.
- Override JSON split rows have stable server UUIDs; synthetic React keys are not accepted as server identity.
- Explicit user rate intent is represented separately from a passive full-object round-trip, including the same-value re-entry case.
- A parent scheduled investment has an update-only `investmentExchangeRateExplicit` intent marker for the same reason.
- Forecast values use the same effective rate that posting will use, including valid pinned rates and per-occurrence investment overrides.
- Unknown cross-currency FX is represented as unknown/unavailable rather than silently converted at `1`.
- Same-currency settlement always resolves to `1`, even if a persisted or explicitly supplied non-1 scalar exists.

## 3. Review progression

| Review stage | Material concern found | Resolution verified |
|---|---|---|
| Early implementation rounds | Stored scalar rates lacked durable pair provenance across parent, split, and override surfaces. | Pair provenance was persisted and checked against the current settlement pair before reuse. |
| Early implementation rounds | Forecast and posting could diverge; stale split/override amounts could survive after a pair change. | Effective-rate and effective-cash resolvers were shared across forecast/posting paths; unknown FX withholds forecast and refuses posting. |
| Round 5 | Manual Post inline splits could bypass the effective resolver; occurrence override investment forecasts could use stale amounts; same-security splits could collide in provenance lookup. | Inline posting was routed through effective resolution; override forecast gained server-resolved effective totals; provenance handling was improved. |
| Round 6 | Parent forecast ignored a still-valid pinned rate; user-entered Post-dialog rate could be ignored; top-level investment overrides were missing from forecast; `securityId + rate` remained an ambiguous identity. | Forecast now uses the effective posting rate; explicit rate edits are honored; top-level override effective amounts are forecast; stable source identity was introduced. |
| Round 7 | `sourceSplitId` leaked into the ordinary transaction serializer and could fail DTO validation; legacy no-provenance rates could be re-blessed during cosmetic edits. | Scheduled-only serialization became opt-in; source identity became independent of provenance and null-pair legacy values remain unprovenanced. |
| Rounds 8-9 | Synthetic override IDs, absent-ID ambiguity, category-to-investment conversion, same-value rate re-entry, and security changes could still confuse provenance intent. | Server UUID identity, `rateExplicit`, full source-kind identity, and security-change rate re-resolution were added. |
| Round 10 | `null` FX could become synthetic `1`; manual Post could ignore `rateExplicit` when a source ID existed; moving an override date via delete+create lost split identity/provenance. | Null remains unknown; inline Post honors explicit intent regardless of source ID; override date changes are update-in-place. |
| Round 11 | Parent scheduled-investment update could not distinguish an explicit same-value re-entry from a passive resend. | `investmentExchangeRateExplicit` was added to the parent update contract; explicit same-value re-entry stamps the current pair, passive resend preserves the old pair. |
| Round 12 | A non-1 scalar stamped against an `X -> X` pair could be treated as current and reused. Example: `10 x 100 x 1.50 = 1,500 CAD` instead of `1,000 CAD`. | Both consumption chokepoints enforce same-currency = `1`: canonical investment FX resolution checks the pair before honoring an explicit scalar, and scheduled stored-rate reuse refuses same-currency reuse so it re-resolves to `1`. |
| Final review | Rechecked final head `862eeefc` against issue acceptance criteria and current `main`. | **No remaining material finding. APPROVE.** |

### Stale cross-currency rate

Initial state:

- security currency: EUR;
- settlement currency: CAD;
- quantity: 10;
- price: 100 EUR;
- stored rate: 1.50 EUR/CAD.

Stored cash impact: `10 x 100 x 1.50 = 1,500 CAD`.

After the same security ID changes to USD and the current USD/CAD rate is 1.35:

- stale implementation result: `1,500 CAD`;
- required result: `10 x 100 x 1.35 = 1,350 CAD`;
- difference: `150 CAD` (`11.11%` relative to the correct result denominator used in the original issue example).

The final implementation rejects reuse of the EUR/CAD scalar because its provenance no longer matches USD/CAD.

### Same-currency non-1 scalar

For a scheduled BUY of 10 x 100 where the current pair is CAD/CAD but a persisted or explicitly re-entered scalar is 1.50:

- incorrect result: `10 x 100 x 1.50 = 1,500 CAD`;
- correct result: `10 x 100 x 1.00 = 1,000 CAD`;
- difference: `500 CAD`, a 50% overstatement relative to the correct cash amount.

The final implementation resolves CAD/CAD to `1` before any explicit scalar can be used and prevents scheduled stored-rate reuse for a same-currency pair.

## 5. Test execution history

The table below distinguishes **implementation-reported local runs** from what the reviewer could verify through GitHub. GitHub exposed no commit statuses and no PR-triggered workflow runs for the final reviewed head, so local pass counts were not independently reproduced by the reviewer.

| Stage / head | Test or check | Reported result | Primary coverage | Reviewer verification status |
|---|---|---:|---|---|
| Post-F1-F4 implementation | Backend unit suites covering scheduled / securities / transactions | **2,888 passed** | Parent/split/override FX provenance, posting, financial mutation | Reported by implementer; GitHub CI unavailable |
| Post-F1-F4 implementation | Frontend forecast / serialization / SplitEditor / dialogs / TransactionForm suites | **459+ passed** | Client round-trip of provenance and split identity | Reported by implementer; GitHub CI unavailable |
| Post-F1-F4 implementation | Local PostgreSQL cross-boundary integration | **2/2 passed** | `create -> createEmbeddedForSplit`; stale amount rejection | Reported by implementer; integration harness starts downstream of scheduled posting |
| Merge-main gate before F2/F4 completion | #1167 + investment + cash-impact suites | **536 passed** | Compatibility with newly merged `main`, including accrued-interest cash-impact changes | Reported by implementer; code overlap reviewed manually |
| Round 10 | Backend scheduled + override + investment + cash-impact + raw-SQL guard | **561 passed** | R10 null-FX, explicit Post intent, override date move, prior regressions | Reported by implementer |
| Round 10 | Frontend scheduled / transaction directories | **1,543 passed** | Editor, serialization, posting UI, override behavior | Reported by implementer |
| Round 10 | `OverrideEditorDialog` suite | **71 passed** | Update-in-place date move; no delete+create | Reported by implementer |
| Round 10 | TypeScript typecheck | **passed** | API/type contract consistency | Reported by implementer |
| Round 10 | Lint | **passed** | Static/style gate | Reported by implementer |
| Round 10 | Break-on-purpose mutation checks | **failed when fixes were deliberately reverted** | Demonstrates R10 guards are non-vacuous | Reported by implementer |
| Round 11 | Parent explicit-re-entry update tests | **passed** | Same-value explicit `1.50` stamps current pair; passive resend preserves old pair; changed `1.37` stamps current | Test code independently inspected; execution not independently reproduced |
| Round 11 | Downstream parent posting tests | **passed** | Current-pair 1.50 reaches `InvestmentTransactionsService.create`; stale-pair 1.50 is dropped | Test code independently inspected; execution not independently reproduced |
| Final head `862eeefc` | `scheduled-transactions.service.spec.ts` + `investment-transactions.service.spec.ts` | **506/506 passed** | R12 same-currency resolver, parent forecast/post, embedded split amount | Reported by implementer; corresponding tests and implementation independently inspected |
| Final head `862eeefc` | Typecheck | **passed** | Final TypeScript consistency | Reported by implementer |
| Final head `862eeefc` | Lint on touched files | **passed** | Final lint gate including existing Prettier reflow | Reported by implementer |
| Final head `862eeefc` | R12 break-on-purpose checks | **guards failed when either same-currency chokepoint was removed/reordered** | Proves both canonical resolver ordering and scheduled reuse guard are required | Reported by implementer |
| Final head `862eeefc` | GitHub combined commit status | **0 statuses returned** | Hosted CI visibility | Independently checked by reviewer |
| Final head `862eeefc` | PR-triggered GitHub workflow runs | **0 runs returned** | Hosted CI visibility | Independently checked by reviewer |

## 6. Important regression obligations now covered

The branch's specification and regression matrix cover, among others:

1. stale parent scheduled rate re-resolution;
2. stale embedded scheduled investment split re-resolution;
3. stale occurrence override split re-resolution;
4. valid pinned-rate preservation;
5. legacy null-provenance behavior;
6. missing FX as unavailable rather than `1`;
7. inline Post-dialog split resolution;
8. stable split identity and same-security collision cases;
9. explicit rate edits, including same-value re-entry;
10. category-to-investment conversion;
11. security changes that previously carried a stale same-currency `1` into a cross-currency pair;
12. top-level investment occurrence overrides in forecast;
13. occurrence override split forecast parity;
14. override date move preserving split identity/provenance;
15. parent forecast/post parity for valid pinned rates;
16. same-currency domination over persisted or explicit non-1 scalars.

## 7. Review limitations

- No GitHub-hosted CI status or PR-triggered workflow run was available for the final head. The reviewer therefore did not independently reproduce the reported local pass counts.
- The local cross-boundary PostgreSQL integration harness starts at `TransactionsService.create` and stubs `ScheduledTransactionsService` because of the module cycle. The scheduled posting decision itself is therefore covered by unit/component tests and the downstream integration closes the financial construction path rather than providing a single full HTTP end-to-end test.
- This was a focused review of issue #1167 and regressions discovered while following its cross-layer paths, not a new full-repository audit.
- FX and other financial behavior should receive normal maintainer/human review before production deployment.

## 8. Suggested PR title

`fix(scheduled): bind persisted investment FX rates to currency-pair provenance`

## 9. PR description - repository template

The block below follows `.github/pull_request_template.md` and is intended to be copy-pasted into the PR description.

---

## Linked discussion / issue

Closes #1167  
Approved in: https://github.com/kenlasko/monize/issues/1167

## Summary

Fix scheduled investment FX-rate staleness by binding persisted rates to their source/settlement currency pair and validating that provenance at consumption time across parent scheduled investments, embedded scheduled investment splits, and occurrence override investment splits.

The implementation also closes the review-discovered edge cases required to preserve that invariant end to end: stable server split identity, legacy no-provenance rows, explicit user rate intent (including same-value re-entry), forecast/post parity, override date moves, unknown FX semantics, and same-currency settlement always resolving to `1`.

Final reviewed head: `862eeefc83c983bc49363ce65584116e2772822f`.

## Checklist

- [x] New behavior has tests, and the existing targeted suites pass locally. **PR CI is still required before merge.**
- [x] All user-facing strings are translated for **every** locale (i18n parity). No new user-facing copy is introduced by the final #1167 fixes.
- [x] No shared/core areas were refactored without prior agreement. The `cash-impact.util.ts` overlap from `main` was merged while preserving both accrued-interest and scheduled-investment behavior.
- [x] The branch is rebased on the latest `main`. **The branch is 0 behind current `main`, but it was updated by merging `main` rather than rebasing. Rebase/squash policy should be applied by the maintainer before merge if required.**

## AI assistance disclosure

Implementation and iterative review used AI assistance, including Claude for implementation work and ChatGPT/OpenAI for repeated adversarial code review. AI-generated output was treated as supporting material rather than proof: findings were re-checked against repository code, regressions were added for confirmed failure scenarios, and the final financial/FX behavior still requires normal human maintainer review before production merge.

Local tests reported by the implementation workflow are summarized in the review notes. At final review, GitHub exposed no commit-status checks and no PR-triggered workflow runs for `862eeefc`; hosted CI should therefore be required on the PR before merge.

---

## 10. Final reviewer conclusion

**APPROVE** at `862eeefc83c983bc49363ce65584116e2772822f`.

After the final R12 same-currency correction, no remaining BLOCKER, HIGH, or MEDIUM finding was confirmed within the #1167 review scope. The final implementation satisfies the issue's material acceptance criteria across parent, embedded split, and occurrence override surfaces, including the financial rule that a same-currency settlement always resolves to `1`.


frontend/src/components/transactions/InvestmentSplitFields.tsx
```diff
@@ -151,7 +188,10 @@ export function InvestmentSplitFields({
     // investment form, which already converts at 4dp.
     const rate = Number(next.exchangeRate);
     const amount = rate > 0 ? roundToDecimals(cashImpact * rate, 4) : 0;
-    onChange(next, amount);
+    onChange(next, amount, {
+      rateEdited: field === 'exchangeRate',
```
kenlasko, line 192: **HIGH (correctness — reopens the #1167 bug class through the UI).** This `rateEdited` emission fires for a blur re-report the suppression guard above does not cover.

The guard at line ~282 only suppresses the no-op blur re-report when `statedRate === undefined` ("only relevant while the rate is market-derived"). But a *stored* split rate is written at up to 10dp (`NUMERIC(20,10)`, `roundFxRate`), and `NumericInput.handleBlur` fires `onChange` whenever the 6dp-parsed display differs from the raw `value` prop. So for any pinned rate with more than 6dp of precision, a focus+blur with **no typing** calls `updateField('exchangeRate', <6dp truncation>)`, which emits `meta.rateEdited: true` here; `SplitEditor` latches it (`prevEdited || meta?.rateEdited === true`) and sends `rateExplicit: true`; `decideSplitProvenance`'s explicit branch then stamps the **current** settlement pair onto that scalar.

Failure scenario: a split holds a pinned 10dp EUR→CAD rate `1.3456789012`; the security's currency later changes to USD. The user opens the split editor to fix a memo and tabs through the rate field. The save sends `rateExplicit: true` with `1.345679`; the server stamps USD→CAD onto a stale EUR→CAD figure, and posting reuses it — exactly the staleness this PR exists to prevent, re-blessed by a cosmetic edit (plus a silent 10dp→6dp truncation).

Suggested fix: extend the suppression to the `statedRate !== undefined` case — ignore an incoming value that equals the current `statedRate` at display precision, rather than gating the whole guard on `statedRate === undefined`.

---

frontend/src/lib/forecast.ts
```diff
@@ -245,9 +349,21 @@ function normalizeInvestmentForForecast(
...
-  const rate = transaction.investmentExchangeRate != null
-    ? Number(transaction.investmentExchangeRate)
-    : 1;
+  // Issue #1167: the forecast converts the security-currency amount into the
+  // cash account's currency with the server-resolved *forecast* rate, never the
+  // persisted `investmentExchangeRate` (which may be stale for the current
+  // settlement pair). The backend sends `1` for a same-currency pair, a resolved
+  // rate for a cross-currency one, and `null` when the current rate is unknown.
+  const rate = transaction.investmentForecastExchangeRate;
+  if (rate == null) {
```
kenlasko: **HIGH (correctness).** This branch's contract — "the forecast builders detect the null rate and withhold the projection" — is honored by `buildForecast` and `buildMultiAccountForecast` (both check `occ.unknown`), but **not by the third consumer, `getProjectedBalanceAtDate`** (line ~804), which also runs every transaction through `normalizeInvestmentForForecast` and then sums `occ.amount` with no `unknown` check.

Result: an investment schedule with unresolvable FX contributes its **raw security-currency amount** (base occurrence) or **0** (unknown override occurrence) to the projected balance shown in `PostTransactionDialog`, with no signal. Scheduled USD BUY of −1000 on a CAD account with no USD/CAD rate: the dialog's projected balance subtracts 1000 CAD-labelled units for a USD amount — the user approves a posting against a silently wrong balance. This is the "amount passed through unchanged" fallback the repo's rate-1 rule forbids, and it's introduced by this PR: pre-PR, normalize converted with the persisted scalar; post-PR it leaves the amount unconverted and relies on a check this caller doesn't make.

`getProjectedBalanceAtDate` needs to either skip unknown occurrences and surface that to the dialog (e.g. return `null`/a completeness flag so the dialog can withhold or caveat the figure), or the dialog must not display a balance built from unknown components — per the "a subtotal is not a total" rule.

---

backend/src/scheduled-transactions/scheduled-transactions.service.ts
```diff
@@ -2502,15 +3586,83 @@ export class ScheduledTransactionsService {
...
+  ): Promise<Map<number, { from: string | null; to: string | null }>> {
+    const provenance = new Map<
+      number,
+      { from: string | null; to: string | null }
+    >();
+    for (const [index, s] of (splits ?? []).entries()) {
+      const inv = s.investment;
+      if (inv?.exchangeRate === undefined || inv?.exchangeRate === null)
+        continue;
+      if (!inv.securityId) continue;
```
kenlasko: **MEDIUM (correctness/hardening).** This `continue` leaves a hole in the "the server decides the pair" invariant. When an override investment split carries a rate but no `securityId`, no provenance entry is created for its index — and `buildOverrideSplits` then falls into its `: s.investment` branch, persisting the client's echoed `exchangeRateFromCurrency`/`exchangeRateToCurrency` **verbatim** into the override jsonb (`InvestmentSplitDto` accepts both fields, so the whitelist does not strip them).

That makes this the one write shape where the pair is client-asserted rather than server-derived, contradicting the type comment that "the client never invents these". A security-less investment line (action where `securityId` is optional) sent via API with `exchangeRate: 1.5` and from/to set to the account's current pair persists an unverified pair; `storedInvestmentRateIsCurrent` later finds it matching and posting reuses a scalar the server never proved belonged to that pair — bypassing `decideSplitProvenance` entirely.

Suggested fix: for skipped indices (no securityId, or null rate), still overwrite the echoed pair to undefined in `buildOverrideSplits` (i.e. record an explicit "no provenance decided" for every index rather than only for decided ones), so posting re-resolves instead of trusting the echo.

---

frontend/src/lib/forecast.ts
```diff
@@ -269,6 +385,26 @@ function normalizeInvestmentForForecast(
...
+ * An investment schedule whose current settlement FX rate could not be resolved
+ * server-side (issue #1167). Its projected cash impact is unknown, so -- like a
+ * missing display-currency rate -- it withholds the whole cumulative projection
+ * rather than contributing a stale or 1:1 figure. The security's currency (the
+ * `from` side of the unresolved pair) names it in `missingCurrencies`, falling
+ * back to the settlement account's currency.
+ */
+function hasUnknownForecastRate(transaction: ScheduledTransaction): boolean {
+  if (transaction.isInvestment) {
+    return transaction.investmentForecastExchangeRate == null;
```
kenlasko: **MEDIUM (correctness — transient, rolling deploy).** `== null` matches `undefined`, so an **absent** `investmentForecastExchangeRate` / `investmentForecastAmount` — a response from a backend that predates this field (rolling deploy, or a cached scheduled-transactions payload) — is read as "unknown FX" and withholds the entire forecast series.

CLAUDE.md's defensive-read rule for newly added response fields says absent means "no information", not "incomplete": here, every investment schedule reads as unknown, `missingRateCurrencies` fills, and `buildForecast` returns `points: []` — the cash-flow chart goes blank for every user holding any investment schedule until the backend catches up, indistinguishable from a genuine missing-rate condition.

Since the new backend always includes the field (a number when known, explicit `null` when unknown), `=== null` is the defensive read that distinguishes "backend said unknown" from "backend never said".

---

backend/src/scheduled-transactions/scheduled-transactions.service.ts
```diff
@@ -1068,6 +1825,65 @@ export class ScheduledTransactionsService {
...
+    // same-currency rows short-circuit to 1 with no rate lookup at all.
+    const asOf = new Date();
+    const result: (ScheduledTransaction & {
+      overrideCount?: number;
+      nextOverride?: ScheduledTransactionOverride | null;
+      futureOverrides?: ScheduledTransactionOverride[];
+      investmentForecastExchangeRate?: number | null;
+      investmentForecastAmount?: number | null;
+    })[] = [];
+    for (const row of rows) {
```
kenlasko: **MEDIUM (efficiency — hot list endpoint).** This loop performs sequential per-row FX resolution on every `findAll` — each investment schedule, embedded split, and future override re-derives the settlement pair with its own account/security reads (`resolveInvestmentForecastRate`, `resolveInvestmentForecastSplitAmount`, `resolveOverrideInvestmentForecastAmount` each derive it again), and for any pair with **no stored rate**, `resolveCashExchangeRateOrNull` → `getRateForDate` goes out to Yahoo synchronously.

The "first resolution persists it" dedup in the comment only helps for *fetchable* pairs. An unfetchable pair (provider can't serve it) persists nothing, so it is retried on **every** list read: a user with ~20 scheduled investments referencing unserved pairs pays 20 sequential external fetch attempts plus the redundant per-row SELECTs on every GET of the schedules/bills list, indefinitely — a multi-second list read with no cap.

Suggested: derive the pair once per (account, security) within a list read, and add some form of negative caching / short-circuit for pairs that just failed to fetch (even an in-memory per-request memo of failed pairs would remove the N-fetches-per-request behavior).

---

backend/src/scheduled-transactions/scheduled-transactions.service.ts
```diff
@@ -1068,6 +1825,65 @@ export class ScheduledTransactionsService {
...
+      ) => {
+        if (!override) return override ?? null;
+        const overrideForecastAmount =
+          await this.resolveOverrideInvestmentForecastAmount(
+            userId,
+            row,
+            override,
+            asOf,
+          );
+        return Object.assign(override, {
```
kenlasko: **LOW (convention).** `Object.assign(override, {...})` mutates the loaded entity in place — the repo rule is "Immutability always — never mutate objects or arrays", and the surrounding code spreads rows immutably (`{ ...row, ... }` two lines later), so this is also inconsistent with its own function. `return { ...override, investmentForecastAmount: overrideForecastAmount }` keeps the same behavior without aliasing risk.

---

kenlasko (commented):
Review of head `4eadc733` (the rebase of the PR body's reviewed head `862eeefc` onto current `main` — the migration renumber to 162/163 accounts for the SHA difference).

The core provenance design held up under tracing: the pair is stamped with fresh rates, preserved on passive round-trips, checked at the consumption chokepoints, and same-currency settlement dominates any stored or explicit scalar. I verified the create/update/override/post/forecast paths agree on effective-rate semantics, that `findCashAccount` falls back rather than throws, and that amount-only actions always carry a total.

Six findings survived verification, ranked in the inline comments:

1. **HIGH** — `InvestmentSplitFields.tsx:192`: a no-typing focus+blur on a >6dp stored rate latches `rateExplicit`, and the server stamps the current pair onto a truncated, possibly stale-pair scalar — the #1167 bug class reopened through a cosmetic edit.
2. **HIGH** — `forecast.ts:358`: `getProjectedBalanceAtDate` (used by `PostTransactionDialog`) never checks the new `unknown` occurrence flag, so unknown-FX schedules feed a raw security-currency amount (or 0) into the projected balance the user approves against.
3. **MEDIUM** — `scheduled-transactions.service.ts:3622`: an override investment split with a rate but no `securityId` persists the client's echoed currency pair verbatim — the one write shape where the pair is client-asserted.
4. **MEDIUM** — `forecast.ts:398`: `== null` treats an absent forecast-rate field as "unknown FX", blanking the forecast during a rolling deploy, against the repo's defensive-read rule.
5. **MEDIUM** — `scheduled-transactions.service.ts:1843`: `findAll` retries a synchronous external FX fetch per unfetchable pair on every list read, with no negative caching.
6. **LOW** — `scheduled-transactions.service.ts:1868`: in-place `Object.assign` on loaded entities against the immutability rule.

Findings 1 and 2 are in the direct scope of the invariant this PR establishes and are worth fixing before merge; 3–5 are hardening/robustness within the same feature; 6 is a convention nit.

Also noting for the record: GitHub exposes no CI statuses or workflow runs for this head (consistent with the PR body's own disclosure), so hosted CI should run before merge.

---

WMP:
### Follow-up review: MEDIUM #2 / #3 / #4 verified fixed

I re-reviewed the three follow-up commits on the current PR head:

`8d63279abf0427cd6ebaf4696be46431e9f1a575`

The fixes for review findings #2, #3 and #4 are independently verified.

#### #2 — Scheduled FX read-model invalidation after exchange-rate refresh

Fixed by `acd3fc73`.

The new `invalidateScheduledFxReadModel()` helper correctly centralizes invalidation of the `scheduled:` cache family when an upstream FX dependency changes.

Verified paths now include:

- successful exchange-rate refresh;
- account update;
- security update.

The refresh invalidation happens **after** the refresh POST succeeds, which is important: the generation-aware cache primitive can therefore also obsolete a scheduled request that was already in flight during the refresh.

The regression test verifies actual behavior rather than only spying on the invalidator:

`1.50 cached -> refresh -> fresh scheduled GET -> 1.35`

This finding is **FIXED**.

#### #3 — Repeated settlement-pair derivation inside `findAll()`

Fixed by `872799b6`, with one non-blocking residual noted below.

`resolveForecastSettlementPair()` now provides a request-local Promise memo keyed by the complete settlement tuple:

`(accountId, fundingAccountId, securityId)`

The memo is threaded through the forecast resolution chain, so repeated parent/split/override calculations for the same tuple join the same pair derivation instead of repeatedly performing the account/security DB reads.

The implementation correctly keeps the two cache layers separate:

- settlement tuple -> currency pair;
- directional currency pair -> FX-rate lookup.

The added tests prove both behaviors:

- same tuple across rows -> one settlement-pair derivation;
- two distinct securities resolving to the same currency pair -> two tuple derivations but one provider-capable rate fetch.

This closes the previously reported O(N) repeated-derivation amplification.

One non-blocking residual remains: the first actual `resolveCashExchangeRateOrNull()` call for a pair still performs its own internal `resolveSettlementCurrencyPair()` call after the pair was already derived for the pair-cache key. This is no longer the O(N) performance defect that was reported; it is bounded to the first rate fetch for a unique pair. Removing it cleanly would require a pair-based canonical FX resolver in `InvestmentTransactionsService`, so I would not block this PR on it.

This finding is **FIXED_WITH_NONBLOCKING_RESIDUAL**.

#### #4 — Sequential forecast enrichment

Fixed by `8d63279a`.

`findAll()` now uses the repository's existing `mapWithConcurrency()` helper with a bound of 6.

The shared request-local caches store Promises and perform their `get -> set` without an intervening await, so concurrent rows resolving the same semantic work still join one in-flight operation rather than duplicating pair derivations or provider fetches.

The implementation preserves input/output ordering.

The deferred-Promise regression test also verifies the important behavioral property directly: two independent FX pairs are both in flight before either one is released. Setting the concurrency bound back to 1 makes that regression fail.

Existing pair-rate and settlement-pair deduplication tests remain compatible with the concurrent execution model.

This finding is **FIXED**.

### Follow-up disposition

For the three fixes above:

**APPROVE — no new material regression found in this delta.**

One previously confirmed MEDIUM remains intentionally out of scope for this follow-up: secondary scheduled-transaction consumers can still use the persisted `amount` even when #1167 has produced a different current/effective amount or marked it unavailable.

That issue affects AI/MCP and several secondary UI/reporting surfaces and requires a broader shared `effectiveAmount` / completeness contract. It should be tracked and fixed in a separate PR rather than expanding #1232 further.

Accordingly, the PR description should not currently claim that there are zero known MEDIUM findings unless the deferred issue is explicitly linked and described as out-of-scope follow-up work.

The PR branch also still needs to be synchronized with current `main`, and the remaining hosted CI jobs should complete successfully before merge.

PR #1249, opened 2026-08-22
Claude/portfolio valuation manual prices

### LINKED ISSUE #1242: Daily portfolio chart and others ignore imported or manual prices (AKA: Who robbed my 401K?)
For holdings without public quote support, the Portfolio Value Over Time chart can differ substantially from the Current Portfolio Breakdown.  This is a critical logic flaw :(

This occurs when:

A security has automatic price updates disabled.
Valid imported or manually entered prices exist.
A short chart range, such as 1 week, falls back to daily snapshots because intraday quotes are unavailable.

**Actual behavior**
The current breakdown uses the latest stored imported or manual price.

The daily chart ignores those stored prices and instead carries forward the price from the most recent qualifying investment transaction, such as a purchase or sale. If that transaction is old, the chart can materially understate or overstate the portfolio.

The yellow intraday warning explains the lower chart resolution, but not that a different and potentially stale price source is used.

**Expected behavior**
Daily historical values should use the latest accepted stored price on or before each date, regardless of whether it came from:

A quote provider
An import
A manual entry
An investment transaction
Disabling automatic price updates should prevent external quote fetching only. It should not cause valid stored prices to be ignored.

**Technical cause**
The daily valuation path treats skipPriceUpdates as “ignore stored security prices” and derives prices exclusively from selected investment transactions. The current portfolio summary correctly reads the latest entry from security_prices.

These two valuation paths therefore disagree for private funds, employer-plan holdings, and other securities without public quotes.

**Impact**
The chart may show a plausible but materially incorrect portfolio value without indicating that it is using an older transaction price. The error can persist indefinitely until another qualifying transaction occurs.

Daily portfolio totals, including 1W fallback and 3M/YTD/1Y charts: net-worth.service.ts:1367-1377
“By security” daily and monthly charts: net-worth.service.ts:1883-1910
Monthly investment snapshot recalculation: net-worth.service.ts:2362-2381
The monthly recalculation writes the stale transaction-price valuation into monthly_account_balances. I confirmed the current-month snapshot already contains that incorrect value.

**Indirectly affected**

Because those snapshots are shared, this can also distort:

Longer-range Portfolio Value charts
Dashboard Portfolio Value widget
Net Worth report and dashboard chart
Latest assets, liabilities, and net-worth summaries
Monthly Comparison investment performance and net worth
AI Assistant/MCP account summaries that use latest net worth
Period change and return calculations using the daily series
The tests currently assert this transaction-price behavior for skipPriceUpdates securities (net-worth.service.spec.ts:3171-3224), so the defect is effectively encoded as expected behavior.

**Not affected**

Current Portfolio Breakdown
Current holdings values
Account Balances “as of” report
Cash/register balances
Those use accepted stored prices correctly. The eventual fix should establish one shared rule: skipPriceUpdates controls external fetching only; every valuation path should still use accepted security_prices rows, including imported and manual prices.

WMP:
## Linked discussion / issue

Closes #1242
Approved in: https://github.com/kenlasko/monize/issues/1242

## Summary

Fix historical portfolio valuation so `skipPriceUpdates` controls external quote fetching only and no longer changes which accepted prices are used to value a holding.

Previously, daily portfolio history, per-security history, and monthly investment snapshots could ignore valid manual or imported `security_prices` rows for securities with automatic price updates disabled. Those paths could instead carry forward an older investment-transaction price indefinitely.

The issue reproduction is a 120-unit holding whose last transaction price is $61 but whose accepted manual price is $120:

- incorrect historical value: `120 * $61 = $7,320`
- correct accepted-price value: `120 * $120 = $14,400`
- difference: `$7,080` (49.17% understatement)

This PR establishes one valuation rule across the affected paths:

- use the latest accepted stored close on or before the valuation date;
- treat provider, imported, manual, and transaction-derived stored prices consistently;
- use transaction history only as a compatibility fallback for legacy data that is missing its derived `security_prices` rows;
- merge stored and legacy observations chronologically, with the accepted stored row winning on the same date;
- never allow a future price to value an earlier date;
- use raw close prices for point-in-time position valuation.

It also fixes the derived snapshot lifecycle around the valuation change:

- manual price create/update/delete now marks affected holding accounts dirty in the same database transaction as the price mutation;
- the stale-snapshot sweep can recover a recomputation lost to a process crash by measuring the grace period from the age of the account mutation;
- migration 162 invalidates existing investment-account `monthly_account_balances` so previously persisted incorrect valuations are rebuilt;
- backup restore no longer treats `monthly_account_balances` as authoritative source data and lets the corrected net-worth path rebuild them after restore.

The legacy transaction-price fallback was also tightened while this path was being corrected:

- short report ranges no longer aggregate the full lifetime transaction history;
- same-day transaction prices use the same six-decimal averaging semantics as the canonical transaction-price writer;
- zero-price disposal observations are preserved consistently with the canonical writer.

Regression coverage includes:

- mixed legacy transaction and stored-price timelines;
- same-day stored-price precedence;
- future stored prices not leaking backward;
- manual/imported price valuation with `skipPriceUpdates`;
- the `$7,320 -> $14,400` issue reproduction;
- bounded legacy transaction fallback and six-decimal rounding;
- zero-price disposal handling;
- real PostgreSQL crash recovery through `SecurityPriceService.createManualPrice`;
- a PostgreSQL failpoint proving the manual price and durable dirty marker roll back atomically;
- backup -> restore -> lazy snapshot rebuild from `$7,320` to `$14,400`.

Local verification completed:

- `TZ=UTC npm run test:unit` — 13,376 passing
- `npm run typecheck`
- `npm run build`
- migration lint
- one-door guard
- doc-path guard

The new PostgreSQL integration tests compile but were not run locally because the development environment used for the implementation did not have Docker/PostgreSQL available. `test:integration` must therefore be verified by CI before merge.

Independent review of commit `1c2138737b4086027fa239fc5303e9c4a1549f63` found no remaining BLOCKER, HIGH, MEDIUM, or LOW implementation findings for #1242.

## Checklist

- [ ] New behavior has tests, and the existing suite passes.

## AI assistance disclosure

Claude was used for implementation, test authoring, and iterative remediation of review findings.

ChatGPT was used for independent adversarial review of the implementation and regression coverage, including repeated re-review of the exact branch SHA.

All generated changes and review findings were manually reviewed. The final implementation remains the contributor's responsibility.


kenlasko:
## Review summary (practical findings only)

Reviewed the full diff (`ff758e55...da4420e6`, 13 files). **Verdict: mergeable, with one small hardening worth doing first and one comment to make accurate.**

The core #1242 valuation fix checks out clean: the boundary-CTE windowing (including pre-window carry-forward), the 6dp rounding matched to `upsertTransactionPrice`, migration 162's tracking and rebuild path, and the restore-side skip of `monthly_account_balances` with its lazy repopulation all hold up against their stated invariants.

### 1. Lock ordering in `markHoldingAccountsDirty` (fix before merge — one line)

`UPDATE accounts SET updated_at = now() WHERE id = ANY($1)` acquires row locks in scan order, while every other account writer (`lockAccountsForBalanceWrite`, per `common/db/locks.ts`) locks in ascending-id order. The concurrency source is real, not hypothetical: the debounced recompute runs in the background, so saving a second manual price while a prior save's recompute is in flight can lock the same holding accounts in opposite orders — a `40P01` deadlock surfacing as a generic 500. Same class of bug as audit RV4-005. Fix: call `lockAccountsForBalanceWrite(m, accountIds, userId)` before the UPDATE.

### 2. The durability comment overstates the mechanism (comment fix, or `clock_timestamp()`)

The marker's `now()` (and the `accounts` `updated_at` trigger) stamp **transaction-start** time. A recompute that starts after the price transaction begins, but commits snapshots first, leaves `computed_at` newer than the marker — so if the post-commit debounce timer is then lost to a crash, `sweepStaleSnapshots`' `updated_at > computed_at` predicate misses it. The joint probability is near zero and the snapshot self-heals on the next touch of the account, so this is **not** a practical defect — but the comment claims unconditional crash-durability the mechanism doesn't have. Either stamp with `clock_timestamp()` or soften the comment to name the residual window.

### Noted, no action needed

- `updatePrice`/`deletePrice` split the precondition read and the write across two transactions — only exploitable by the same user racing themselves from two tabs.
- Changing a manual price's `priceDate` doesn't backfill a transaction-derived row for the vacated date the way `deletePrice` does — needs a very specific sequence, and valuation already compensates via the tx-price fallback.
- The `monthly_account_balances` restore skip is string-matched in the restore loop rather than declared on the `RESTORE_PLAN` entry where `restore-plan.spec.ts` could verify it — maintainability preference, not a runtime issue.

---

PR #1254, opened 2026-08-23
Pub mortgageamortization

### LINKED ISSUE #1253: Scheduled loan payment interest drifts from amortization report
## Summary

After a scheduled loan or mortgage payment posts, Monize can calculate the next bill's interest a few cents differently from the projected row in the loan amortization report.

The issue affects standard US mortgages and other amortizing loans. It is not specific to Canadian mortgage compounding.

## Affected behavior

- The loan amortization report calculates each projected period's interest from the outstanding balance and periodic interest rate.
- Initial bill setup can also calculate interest from the current balance when the detected historical split option is disabled.
- Subsequent bill recalculation instead advances the previous stored principal/interest split with an amortization recurrence.
- Stored split amounts have already been rounded to money precision, so the recurrence carries that rounding difference into future bills.
- The stored `currentBalance` excludes future-dated transactions. Using it after a future payment posts therefore repeats the prior balance and interest, even when future regular or principal-only payments have reduced the debt.

## Reproduction

1. Configure an amortizing loan or mortgage with an annual interest rate, payment amount, payment frequency, and scheduled split payment.
2. Ensure the scheduled payment contains principal and interest splits.
3. Post one scheduled payment.
4. Compare the interest split prepared for the next bill with the first projected interest row in the loan amortization report.
5. Repeat across several payments.

## Actual result

The next bill's interest may differ from the amortization report by a small amount. Differences can accumulate because each recalculation uses previously rounded split values.

## Expected result

For the same balance, annual rate, payment frequency, and mortgage compounding mode, the next scheduled bill and the first projected amortization row calculate the same interest amount. The balance is measured from the ledger through the next due date so all already-posted principal movements on or before that date are included:

```text
debt = max(0, -(openingBalance + postedTransactionsThroughNextDueDate))
interest = roundMoney(debt * periodicRate)
```

Principal is the regular payment less interest, subject to the existing loan payment waterfall and remaining-balance caps.

## Root cause

`ScheduledTransactionLoanService.recalculateLoanPaymentSplits` used this recurrence when prior principal and interest splits were available:

```text
nextInterest = previousInterest - previousPrincipal * periodicRate
```

The recurrence is algebraically equivalent to recalculating from balance only when its inputs retain full precision. Scheduled transaction splits are stored at money precision, so fractions discarded from the previous split cannot be recovered.

Replacing the recurrence with `account.currentBalance` alone is incomplete because that column intentionally represents the balance through today and excludes future-dated transactions. A payment posted for a future date therefore does not change it.

## Proposed fix

Always calculate the next scheduled interest split from the authoritative loan ledger balance through the schedule's next due date. Include every non-void, top-level transaction through that date, including regular and principal-only payments. Continue using the existing periodic-rate rules:

- Standard US mortgage and loan: nominal annual rate divided by payments per year.
- Canadian fixed-rate mortgage: effective periodic rate derived from semiannual compounding.
- Existing variable-rate behavior remains unchanged.

Continue passing the calculated interest and principal through `allocateLoanPayment` so interest-first allocation, extra principal, and final-payment clamps remain unchanged.

## Acceptance criteria

- A standard US mortgage bill calculates interest from the post-payment balance.
- Previously rounded or inconsistent template splits do not influence the next interest calculation.
- Future-dated regular and principal-only payments posted on or before the next due date reduce the balance used for interest.
- Transactions after the next due date do not affect that installment.
- Canadian fixed-rate periodic-rate behavior remains intact.
- Loan, mortgage, and line-of-credit scheduled payments remain supported.
- Extra-principal and final-payment clamps continue to pass their existing tests.
- The scheduled-loan service test suite includes a regression where prior rounded splits disagree with the authoritative balance.

bcorob:
## Linked discussion / issue

Closes #1253 
Approved in: Not approved yet

## Summary

Work in progress

## Checklist

- [ ] This PR addresses a **single concern** (large work is split into a series).
- [ ] New behavior has tests, and the existing suite passes.
- [ ] All user-facing strings are translated for **every** locale (i18n parity).

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -189,6 +189,10 @@ A check capable of refusing a command belongs inside the transaction that perfor
 
 Give the operation the caller's precondition as a parameter -- the expected owner, scenario or revision -- and let it refuse before writing. Return the refusal distinguishably: "no such row", "not yours" and "done" are three answers, and folding two into `null` makes the caller guess. Tests assert the rejected response **and** the stored state; see `docs/financial-calculation-contract.md` section 7.
 
+## Scheduled loan interest uses a dated ledger balance
+
+`accounts.current_balance` excludes future-dated transactions, so it cannot price an installment after future payments have been posted. Recalculate scheduled loan interest from opening balance plus every non-void, top-level transaction through the schedule's next due date; this includes regular and principal-only payments, while later transactions belong to later installments.
+
 ## A category's leaf name is not its identity
 
 "Cell Phone" under **Bills** and under **Business** is an ordinary chart of accounts, so a bare leaf name identifies nothing. Both halves go through `categories/category-name.util.ts`:
--- a/backend/src/scheduled-transactions/scheduled-transaction-loan.service.spec.ts
+++ b/backend/src/scheduled-transactions/scheduled-transaction-loan.service.spec.ts
@@ -15,6 +15,7 @@ describe("ScheduledTransactionLoanService", () => {
   let scheduledTransactionsRepository: Record<string, jest.Mock>;
   let splitsRepository: Record<string, jest.Mock>;
   let accountsRepository: Record<string, jest.Mock>;
+  let manager: Record<string, jest.Mock>;
 
   const loanAccountId = "acc-loan";
   const scheduledTransactionId = "st-1";
@@ -43,6 +44,7 @@ describe("ScheduledTransactionLoanService", () => {
       name: "Loan Payment",
       amount: -500,
       frequency: "MONTHLY",
+      nextDueDate: "2026-06-01",
       isActive: true,
       splits: [
         {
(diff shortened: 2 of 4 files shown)


kenlasko:
Closing as superseded — #1278 landed the fix for #1253 (merged as `1e819ee`), and the issue is now closed.

Credit where it is due: @bcorob, this PR's approach is the one that shipped. Pricing the installment from the ledger balance through the schedule's next due date, rather than advancing the previously stored split, is exactly what `ScheduledTransactionLoanService` does on `main` now — the SQL shape, the inclusive `<= nextDueDate` boundary and the non-VOID / top-level predicates all match what you wrote here. The issue you filed alongside it did most of the diagnostic work, and its acceptance criteria are what the merged change was tested against.

#1278 went on to cover three things this branch does not, which is the only reason it landed instead:

- **The posting path.** Recalculating the template fixes the arithmetic but not its staleness — the stored P/I split is a template, and a principal movement committed between occurrences leaves it describing last period. The posting now re-resolves the allocation inside its own transaction, at the date the money actually moves.
- **The report anchor.** The amortization report still projected from today while the bill priced through its due date, so the two disagreed by construction for any future-dated or between-occurrences payment.
- **The rate.** A rate change recorded through the rate-history UI deliberately does not write `accounts.interest_rate`, so pricing at that column charges a rate nobody pays. The bill now resolves the rate from the `loan_rate_changes` timeline at the same date, against a truth table both layers assert.

Nothing here is lost by closing. `docs/bugs/scheduled-loan-interest-rounding-drift.md` was this branch's only unique file and its content is issue #1253's body; `docs/specs/scheduled-loan-installment-pricing.md` on `main` carries the same analysis plus the invariant (INV-LOAN-006), the scope limits and the test obligations.

Thanks for the report and the fix — reopen if you think something in here still isn't covered.

---

PR #1275, opened 2026-08-28
fix(backup): stop automatic backups going out in plaintext, and rename AI_ENCRYPTION_KEY to ENCRYPTION_KEY

### LINKED ISSUE #1269: Encrypted backup password setting
Hi

Trying to set up encryption password for backups. The docs [here ](https://github.com/kenlasko/monize/wiki/Backup-and-Restore#enabling-encrypted-backups-local-password-users) say that there should be a field to confirm the password. I don't see that in my UI.

Logged in as Admin with local email + password, on v1.15.0

Image

kenlasko:
## Linked discussion / issue

Closes #1269
Approved in: https://github.com/kenlasko/monize/issues/1269 (reported by @doll-house-hubgit, confirmed with evidence, and scoped by @kenlasko's "Clearly not working as designed. Will get it working.")

## Summary

Automatic backups for local-auth accounts are meant to be encrypted with the login password the server captures when the user types it, with nothing to switch on. On the reporter's deployment none of them were: the stored copy of that password was encrypted with `AI_ENCRYPTION_KEY`, which was **optional** — commented out in `.env.example`, documented as being for cloud AI providers, and passed as `${AI_ENCRYPTION_KEY:-}` by the compose files. So on any deployment that never configured an AI provider, `rememberLoginPassword` returned early at every login, `backup_encryption_enabled` stayed false, and every backup was written as `.json.gz`. Three things kept it invisible: no log line, no status field, and the Settings encryption section renders only for OIDC accounts — so a local user's page said nothing about encryption at all, which is what the issue was originally opened about.

**A feature that is on by default cannot hang off configuration a deployment is told it does not need.** Four changes, one concern:

1. **One key, one name, one service.** `ENCRYPTION_KEY` replaces `AI_ENCRYPTION_KEY` — the variable was named for the first thing that used it, and that name is why this happened. `AI_ENCRYPTION_KEY` is still read, and still **wins** where both are set: every column already encrypted under it (provider keys, emergency-access credentials, the stored backup password) is AES-GCM ciphertext only that key opens, so preferring the new name would open none of it. Existing deployments upgrade untouched. `AiEncryptionService` became `EncryptionService` under `common/encryption/` with its own small module — `BackupModule` and `EmergencyAccessModule` were importing the entire AI graph to reach one 90-line provider.

2. **Announced before it is enforced.** The server still starts without a key, because refusing to boot would turn an upgrade into an outage for exactly the deployments that have the problem. Instead it warns on every boot, in four lines, that backups are going out **unencrypted** and that a future release will refuse to start. The write paths still refuse, so no secret is ever stored in the clear — an unkeyed deployment finds out on the request that needed one. Making the requirement real later is one branch: turn the unkeyed case into a throw.

3. **The state is visible.** Every unencrypted automatic backup logs a warning; `getStatus` reports `available` ("this server cannot encrypt") separately from `enabled` ("this user has not turned it on") because they have different fixes; and the Settings page now renders all four states for local accounts, where it previously rendered nothing.

4. **A rescue for long-lived sessions.** The sign-in capture never fires for a session older than the deploy that shipped it, and such a session can outlive the backups it leaves in plaintext. `POST /backup/encryption/login-password` lets a local account confirm the password it already has — verified against the account's own hash inside the transaction that writes, so a misremembered password is refused rather than encrypting future backups under a string nobody knows. Throttled like the auth controller's other credential checks.

Also corrected along the way: `ENCRYPTION_KEY` does **not** open a backup file — the user's password does. Losing the key costs stored provider keys, emergency-access credentials, and automatic backups until each user signs in again; it strands no artifact whose password is still known. That is now stated in the backup contract and on the wiki, along with the trap that does strand backups: a file opens with the password in effect **when it was written**, so changing a password leaves older backups needing the older one.

The wiki pages (Backup and Restore, Getting Started, Emergency Access) were updated separately and are already live — they described encrypted backups as an opt-in feature with an Enable/Disable pair, which stopped being true in 1.14 and is the gap this issue was reported from.

## Checklist

**On the unticked box:** this does refactor a shared area. `AiEncryptionService` is renamed and moved to `common/encryption/`, and two modules now import `EncryptionModule` instead of `AiModule`. It is mechanical (57 references, no behaviour change) and it is what makes the env-var rename coherent — a class called `AiEncryptionService` reading `ENCRYPTION_KEY` and encrypting backup passwords is the same naming drift that caused the bug. Flagging it rather than ticking the box: say the word and I will split it out.

**Two upgrade consequences**, both deliberate and both worth a second opinion:

- A deployment with **neither** variable set keeps working and gets the warning. Nothing breaks; its backups stay unencrypted until an operator acts.
- `docker-compose.e2e.yml` deliberately set an empty key so the E2E suite exercised the plaintext path. It now gets a real one, and the export test drives the encrypted path end to end (prompt, `.mzbe`) — testing the configuration deployments should actually have.

### Testing

- `encryption-key.spec.ts` — key precedence (legacy wins), a too-short value under either name treated as absent, and the boot warning's content pinned by *what it must say* (the coming requirement, what breaks now, `openssl rand -hex 32`) rather than a call count.
- `backup-encryption.service.spec.ts` ends with a block that builds the **real** `EncryptionService`, once per live variable name. Every other test in that file supplies a double whose `isConfigured()` returns true and is therefore structurally blind to this class of bug — that is how the suite stayed green while backups went out in plaintext.
- The enable path's refusals (wrong password, OIDC account, no key) each assert **nothing was written**; the auto-backup warning and all four Settings states have tests.
- Both new backend tests were verified to fail against the pre-fix code by reverting each mechanism in turn.
- Backend `test:unit` (13,924) and the full frontend suite (14,306) pass, minus one pre-existing failure on each side — `occurrence-selection.guard.spec.ts` and its twin `scheduled-effective-amount.guard.test.ts` are red on `main` at `fed433f` in a clean worktree, unrelated to this change. Both layers lint, typecheck, and pass i18n parity and pseudo-locale checks.

### Invariants touched

INV-BACKUP-001. Its "encryption is settled" note claimed a backup whose password cannot be decrypted is refused rather than written in clear — true, but it never claimed a backup is encrypted *at all*, which is the gap this closes. `docs/system-invariants.md` and `docs/backup-restore-contract.md` §7 record the new state and are explicit that today's enforcement is visibility, not refusal.

## AI assistance disclosure

Written by Claude Code (Opus), driven and reviewed by @kenlasko. Root cause, fix design, tests, docs and the wiki edits were produced in-session; the two upgrade consequences and the unticked checklist item above are flagged for human judgement rather than assumed.

---
_Generated by [Claude Code](https://claude.ai/code/session_016hJQYVb3WXQVenoXyT4hiQ)_


kenlasko:
**"Documentation vs Manifests" is failing on `main`, not on this branch — and I am deliberately not fixing it here.**

The job's third step, `check-migration-prefixes.mjs`, reports:

```
note: compared against origin/main: 0 migration(s) added on this branch, base maximum prefix 166.

Migration prefix check failed:
  duplicate migration prefix 165: 165_clamp_reminder_days_before.sql, 165_heal_semimonthly_scheduled_frequency.sql
  duplicate migration prefix 166: 166_heal_loan_schedule_end_dates.sql, 166_override_occurrence_identity.sql
```

Both pairs are on `origin/main` at `fed433f`, and the check says so itself: **0 migrations added on this branch**. `git diff origin/main..HEAD -- database/migrations` is empty. Two separately-merged PRs each claimed 165 and 166. Re-running the job will not change the result — it is deterministic, not a flake — so I have not spent a re-run on it.

The first two steps of the job pass (`OK: all 51 referenced env vars are documented`, `Documentation/manifest check: OK`), which matters here because those are the ones this PR could plausibly have broken by adding `ENCRYPTION_KEY`.

### Why I am not renumbering the files

The obvious fix — bump one file of each pair to 167/168 — is not obviously safe, and it is a decision about somebody else's already-merged migrations rather than about this change:

`db-migrate.ts` tracks applied migrations by **filename** (`SELECT filename FROM schema_migrations` / `INSERT INTO schema_migrations (filename)`). Renaming a migration that has already run therefore makes every existing deployment run it again. For a pure DDL migration that is a no-op by design — the repo requires migrations to replay cleanly on top of `schema.sql`. But two of the four files are data *heals*:

- `165_heal_semimonthly_scheduled_frequency.sql`
- `166_heal_loan_schedule_end_dates.sql`

Re-running a heal re-applies a data correction to rows the user may have deliberately changed since. That is a real risk and a maintainer's call, not something to fold into a backup-encryption PR.

### Proposed patch, for whoever picks it up

One of these, in its own PR:

1. **Renumber the two non-heal files** (`165_clamp_reminder_days_before.sql` → `167_`, `166_override_occurrence_identity.sql` → `167_`/`168_`) *if* each is verifiably re-runnable as a no-op, and say so in the PR. This keeps the heals' filenames — and therefore their applied state — untouched.
2. **Leave the filenames alone** and teach `check-migration-prefixes.mjs` that a duplicate pair already on `main` is a recorded fact rather than a new error, with the pairs listed explicitly so a *new* duplicate still fails. This is the honest option if either non-heal file also cannot be safely re-run.

Option 2 is the one I would take: the ordering damage, if any, is already done on every deployment that has applied these four, and renaming files to satisfy a lint is how a heal gets re-run by accident.

Until one of those lands, this check stays red here through no fault of the diff. Everything else on the PR is mine to get green and I am watching it.

---

kenlasko:
**Fixed here after all** — @kenlasko asked for the duplicate prefixes to be resolved rather than recorded, so `d829dff` renumbers them. This supersedes the "not fixing it here" position in my comment above; the rest of that comment (why the check is red, that it is red on `main` too) still stands.

Which file moves was forced rather than chosen. `schema_migrations` keys on **filename** — `db-migrate.ts` reads the applied names and runs whatever is missing — so renaming a migration runs it again on every database that already applied it. Only a file whose body excludes its own result survives that:

| Pair | Moved | Why that one |
|---|---|---|
| 165 | `165_clamp_reminder_days_before.sql` → **167** | Both `UPDATE`s stop matching once clamped; the `CHECK` is dropped before it is added. A re-run does nothing. |
| 165 | `165_heal_semimonthly_scheduled_frequency.sql` — **stays** | Its heal re-dates `next_due_date` against the day it runs. A re-run is a no-op today (no `SEMI_MONTHLY` rows remain), but that is not a property to arrange deliberately. |
| 166 | `166_override_occurrence_identity.sql` → **168** | The `DELETE` keeps the newest row per occurrence slot, so a second pass matches nothing — and the new `UNIQUE` makes duplicates unreachable anyway. Every DDL statement is guarded. |
| 166 | `166_heal_loan_schedule_end_dates.sql` — **must never move** | It steps a loan's `end_date` back one interval and *cannot* exclude its own result: a second run retires the schedule a payment early. Its own header says so, and it is registered in `NON_RERUNNABLE_DATA_MIGRATIONS` under that filename. |

Each moved file carries that reasoning at the top, so the next reader does not have to re-derive why 167 and 168 sit after two lower-numbered heals.

A renumber is a rename, so the references moved with it: the two specs that read a migration **by path** (`override-identity.guard.spec.ts`, `reminder-window.spec.ts`), the `RAISE EXCEPTION` text inside the override migration, and the "migration 166" prose in `schema.sql`, `scheduled-occurrences.ts`, the override entity and service, and `docs/financial-semantics.md`. The two surviving "migration 165" mentions belong to the SEMI_MONTHLY heal and stay — and are now unambiguous, which the duplicate prefix had quietly cost them.

**Verified:** `check-migration-prefixes.mjs` (OK, 158 migrations), `migration:lint` (OK, with the one registered one-shot still correctly excused), `source-comment-paths.spec.ts` — which caught my first draft for naming the old filenames in the new headers — and the full `scheduled-transactions` and `common` suites.

**Not verified locally:** `scripts/verify-schema.sh` needs a Docker daemon this session does not have. The rename changes apply order but no SQL, and the two moved files touch `scheduled_transactions.reminder_days_before` and `scheduled_transaction_overrides` respectively — neither of which the heals keeping 165/166 goes near — so the replay is order-independent. CI's "Schema vs Migrations Drift" job is the confirmation, and it is watching.

One consequence worth stating plainly for whoever deploys this: **both renamed migrations will run once more on every existing database.** That is the cost of the rename, it is why only self-excluding files were eligible, and it is a no-op on both.

---

kenlasko:
**Backend Integration Tests was mine — fixed in `b835da1`.**

One test failed, and the rename caused it:

```
● emergency access across grant cycles › refuses to arm the feature when
  credential encryption is unavailable

  Expected pattern: /AI_ENCRYPTION_KEY/
  Received message: "Credential encryption is not configured. Emergency access
  cannot be enabled until ENCRYPTION_KEY is set, ..."
```

The assertion pinned the old variable name in a message this PR renamed. Two changes in `emergency-access-grant-cycles.integration.spec.ts`:

- It now matches `/until ENCRYPTION_KEY is set/` — the sentence rather than the bare token, because `/ENCRYPTION_KEY/` on its own would still pass if the copy regressed to `AI_ENCRYPTION_KEY`, which is exactly what this assertion exists to catch.
- While there: the suite's `EncryptionService` double keyed its test key on `AI_ENCRYPTION_KEY`, so it was proving the *legacy* path and leaving the current name untested in integration. It keys on `ENCRYPTION_KEY` now. The legacy name keeps its coverage in `encryption-key.spec.ts` and in `backup-encryption.service.spec.ts`, whose regression block runs once per live variable name.

**How this got past me, stated plainly:** the integration suite is the one gate this session cannot run — it needs a reachable PostgreSQL, and the container has neither a server nor a Docker daemon. I ran `test:unit`, both lint and typecheck passes, and the frontend suite locally; CI found the rest, which is what it is for.

**Backend Unit Tests** also failed on that same head, on `occurrence-selection.guard.spec.ts › expands a recurrence in exactly one place`. That one is not this PR's: it fails identically on `main` at `fed433f` (CI Pipeline run #2888, which failed on exactly these three jobs — Documentation vs Manifests, Backend Unit, Frontend Unit) and in a clean worktree of that commit locally, and this branch changes nothing under `scheduled-transactions/` except the migration-number references in comments. Expect Frontend Unit Tests to be red for its twin, `scheduled-effective-amount.guard.test.ts`, for the same reason. Consider this the one standing-down note covering both.

So on the current head the expected state is: the migration prefix check green (fixed in `d829dff`), Backend Integration green (this commit), and Backend/Frontend Unit still carrying `main`'s two pre-existing guard failures. Still watching.

---

kenlasko:
**Both inherited guard failures fixed here** (`ad26c7f`), at @kenlasko's direction. This supersedes the standing-down note in my earlier comment.

They were worth fixing rather than waiting out, because they gate more than their own red X:

```
e2e-browsers:  needs: [backend-unit-tests, frontend-unit-tests]
e2e-tests:     needs: [e2e-browsers]
```

Every E2E shard has been **skipped** on this PR — including `e2e/tests/backup.spec.ts`, which this branch rewrote to drive the encrypted download end to end (password prompt → `.mzbe`). The only automated coverage of the flow at the centre of this PR had never run.

Neither offender is the defect the guards were built for; both are the case their allowlists already exist for, so both get an entry naming why:

- **`accounts/payment-frequency.util.ts`** — `advancePaymentDates` takes one date and returns one date, the date N payments later, which dates a loan's payoff and a mortgage's term end. No window, no overrides, no list of occurrences, so nothing in it can price one. It walks the shared engine *deliberately*: its own comment says deriving the date arithmetically instead "would be a second calendar again" — the very thing this guard protects against, which makes the loop the remedy rather than an instance of the problem.

- **`lib/loan-overpayments.ts`** — `recurringOccurrencesDue` counts how many occurrences of a recurring *overpayment* fall due by each amortization row. The cadence is one the user typed into the loan calculator (`RecurringExtra`: an amount and a frequency), not a `scheduled_transactions` row — so there is no override to look up and no stored amount to re-price. The amount is the one on screen.

An allowlist entry weakens a guard, so each entry says exactly what was exempted and what would still be a violation. The frontend guard's companion test — *"every exemption still exists, so the list cannot go stale"* — keeps both honest if either file moves.

**Verified locally:** backend `test:unit` fully green for the first time in this branch's history (518 suites, 13,923 passing, **none failing**), frontend `src/lib` green (2,200) including both guards, and lint + typecheck clean on both layers.

Both fixes are `main`'s bugs landing here rather than in their own PR, which does widen this one past its single concern — flagging that alongside the `AiEncryptionService` move already noted in the description, rather than quietly folding it in.

With this, the PR should be green end to end, and the E2E shards should actually run for the first time.

---

kenlasko:
**E2E ran for the first time and caught a defect in this PR's own test — fixed in `e8ef780`.**

Unblocking the two guard specs opened the gate, and the first thing through it failed:

```
strict mode violation: getByText('Backups are encrypted with your login password.')
resolved to 2 elements:
  1) <p>Your backups are encrypted with your login password. Nothing else to remember...</p>
  2) <span>Backups are encrypted with your login password.</span>
```

The Settings panel's description **contains** the badge note as a substring, and `getByText` matches substrings, so the locator was ambiguous the moment both were on screen. `exact: true` matches the note's whole text and nothing else. Chromium (shard 1/4) and Firefox (shard 3/4) failed identically — 86 and 81 other tests passed in those shards, so this was the only casualty.

I checked the test's other three locators against the message catalog rather than assuming they were fine: `'Backup Encryption'`, `'Encrypt Backup'` and the `'Login password'` placeholder each occur **exactly once**, case-sensitively, across `settings.json`. The encryption *setup* modal's own placeholder ("Your login password") differs in case as well as position, so it cannot collide with the export modal's either.

**What is still unproven:** everything after that assertion — the password prompt, the download, and the `.mzbe` filename — because the failure stopped the test before reaching any of it. That is exactly the flow this PR exists to protect, so the next run is the one that matters. This session has no Docker daemon and cannot drive the stack locally, so CI is the only place it can be established.

Worth recording plainly: **the E2E suite justified itself within one run of being unblocked.** The unit and component tests for this panel are green and were never going to catch a locator that is only ambiguous once both strings render together.

---

PR #1276, opened 2026-08-28
Unblock main: two guard allowlists and two duplicate migration prefixes

WMP:
## Linked discussion / issue

**None, and deliberately so — please read this section before the checklist.**

Closes #
Approved in: *(nothing to link)*

This is not a proposal. `main` is currently red on three CI checks, and every PR opened against it inherits all three, so the propose-first workflow cannot be satisfied first: opening a Discussion and waiting would leave every open PR unable to demonstrate a green build in the meantime.

That means the **PR Checklist** job will fail on this PR too, for exactly the same reason it fails on the others — no linked issue carries `approved-to-build`. I have not tried to work around that gate, and I am not asking anyone to bypass it. If you want this under an issue, say which and I will link it; if a build-breakage fix is exempt from propose-first, that is your call to make, not mine to assume.

## Summary

**`main` is red on three checks. None of the three is a defect in the code they flag; each is a guard or a check that a later commit outgrew.**

| Failing check | Flags | Fix |
|---|---|---|
| Backend Unit Tests | `occurrence-selection.guard.spec.ts` → `accounts/payment-frequency.util.ts:272` | reasoned allowlist entry |
| Frontend Unit Tests | `scheduled-effective-amount.guard.test.ts` → `lib/loan-overpayments.ts:352` | reasoned allowlist entry |
| Documentation vs Manifests | duplicate migration prefixes `165`, `166` | renumber to `167`, `168` |

Both guards landed in `7a87dd1` (2026-08-26). Both files they flag were last touched by `627fd0a` (2026-08-27) — code added a day after the guard, without extending its allowlist. That is how `main` went red.

The diff is **5 files, +66/−13, no production code**: two spec files, one test file, and two migrations that are pure renames (`git` reports both as `100%` similarity — zero content change).

### Why an allowlist entry rather than a code change, in both cases

Each guard's purpose is to keep *scheduled-transaction occurrence selection* in one place, so amount/override/date decisions cannot fork. Neither flagged site touches occurrences, overrides or per-occurrence amounts:

- **`advancePaymentDates`** answers "what date is N payments after this one". It is date arithmetic over a **loan term** — no window, no overrides, no amount — and its only two callers (`loan-amortization.util`, `mortgage-amortization.util`) ask it for a term end date. Routing it through the occurrence expander would be nonsense, not an improvement.
- **`recurringOccurrencesDue`** counts how many **overpayment** occurrences a loan schedule row has passed. It is `INV-LOAN-001`'s own mechanism and the one place that cadence is walked; it expands an overpayment *calendar*, not a scheduled transaction, so the effective-amount contract has nothing to say about it.

Both entries carry that reasoning inline, in the style of the entries already there.

### The backend allowlist had no staleness check; it does now

Its frontend twin already asserts "every exemption still exists". The backend list had nothing, so an entry could rot silently and pre-authorise the next collision. `EXPANDERS` is hoisted out of the test and a second case holds it against the tree: a path that no longer exists, or whose code no longer calls a recurrence helper at all.

**That check deliberately does not re-run `insideLoop`,** and the reason is worth recording. Written that way first, it failed — naming `common/scheduled-occurrences.ts`, the canonical expander the whole guard exists to sanction. Its `while` sits 23 lines above its `calculateNextDueDate` and the heuristic looks 12 lines up. A staleness check built on the main scan's verdict inherits the main scan's blind spots, so this one asks the narrower question it can actually answer.

### Migrations: which file moves is a safety decision, not a numbering one

Prefixes `165` and `166` were each used twice, by four review passes landing within nine hours. `GRANDFATHERED_DUPLICATES` is **not** the fix: the script fails when a listed prefix stops being duplicated, so that list may only shrink, and adding a fresh collision to it would pre-authorise the next one.

Renumbered keeping author order:

```
165 clamp_reminder_days_before             (unchanged, earliest)
166 heal_loan_schedule_end_dates           (unchanged)
167 override_occurrence_identity           (was 166)
168 heal_semimonthly_scheduled_frequency   (was 165, latest of the four)
```

Author order was not the binding constraint. **`schema_migrations` tracks by filename, so a renamed migration re-runs on every already-migrated database** — see the operational note below. `migration:lint` reports `166_heal_loan_schedule_end_dates` as a registered one-shot data repair: *"a second pass would step it back again and retire the schedule one payment early."* Renaming **that** file would silently damage every deployed database. It keeps its number. The two that do move are re-runnable, which the same lint asserts.

The renumbering also moves `heal_semimonthly` from second to last, changing its order relative to the other two. Behaviour-preserving, and checked rather than assumed:

- `heal_loan_schedule_end_dates`'s `WHERE` ends in an **exhaustive gate** admitting only `WEEKLY`/`BIWEEKLY`/`MONTHLY`/`QUARTERLY`/`YEARLY`, so a semi-monthly schedule is skipped under either spelling and in either order.
- `override_occurrence_identity` touches only `scheduled_transaction_overrides`, with no reference to `frequency` or `next_due_date`.

`override-identity.guard.spec.ts` names its migration by filename and is updated with it.

### Operational consequence of merging this

On any database that has already applied `165_heal_semimonthly_scheduled_frequency` or `166_override_occurrence_identity`, those two **will run a second time** under their new names, because `schema_migrations` keys on the filename. Both are idempotent — that is the property `migration:lint` enforces and the reason these two were the ones chosen to move — so the second pass is a no-op. Stating it explicitly because it is the only operational risk in this PR, and it is invisible in the diff.

### Verification

| | Before (this base) | After |
|---|---|---|
| Backend | 516/517 suites, 1 failed | **517/517 suites, 13,899 passed**, 2 skipped |
| Frontend | 733/734 files, 1 failed | **734/734 files, 14,302 passed** |

Suite counts are unchanged, so nothing was skipped or disabled to reach green; the backend guard has one test *more* than before. Also passing: all three `Documentation vs Manifests` steps (`check-env-docs`, `check-docs-manifests`, `check-migration-prefixes`) and `migration:lint`.

Negative controls, each run and confirmed to fail:

- removing either allowlist entry → that guard's scan fails again;
- adding a bogus allowlist entry → the new staleness check catches it.

## Checklist

      **Cannot be checked** — see the first section. There is no issue, because this fixes a breakage in `main` rather than proposing a change. Point me at one and I will link it.
      One concern: make `main` green. Three fixes, because that is how many checks are red.
      No production behaviour changes — the diff is tests, allowlists and two file renames. The added behaviour is the backend staleness check, which is itself the new test, and it carries its own negative control.
      No user-facing strings are added or changed.
      **Flagged rather than claimed.** `database/migrations/` is a shared area and two files there are renamed. Nothing is refactored and no migration body changes, but the rename has the re-run consequence documented above, so it needs your awareness rather than my assurance.
      Branched directly from `main`; `git rev-list --count HEAD..main` is 0.

## AI assistance disclosure

Written with Claude Code (Anthropic's CLI). Each of the three failures was reproduced on a pristine `main` checkout before being attributed to `main` rather than to a branch, and each fix carries a negative control. I have reviewed and own the result.


WMP:
Possible duplicate with #1275

PR #1278, opened 2026-08-29
fix: price scheduled loan installments from the dated ledger balance (#1253)

### LINKED ISSUE #1253: Scheduled loan payment interest drifts from amortization report

kenlasko:
## Linked discussion / issue

Closes #1253
Approved in: https://github.com/kenlasko/monize/issues/1253 (scope, root cause and the two residual findings from the audit of the first proposed fix, PR #1254)

## Summary

A scheduled loan payment's interest drifted away from the amortization report beside it. Three separate causes, one invariant.

**The arithmetic.** `recalculateLoanPaymentSplits` advanced each period from the *previously stored* split (`next = prev_interest - prev_principal * rate`). That identity holds at full precision, but a stored split is money already rounded to 4dp, so every occurrence compounded the discarded fraction — the issue's 98.0101 against a true 98.0000. It also read `accounts.current_balance`, a through-today read model that excludes future-dated rows, and `accounts.interest_rate`, which a rate change recorded through the rate-history UI deliberately does not write.

**The staleness.** The stored P/I split is a *template*, not an executable instruction. A principal movement committed between occurrences left it stale with no mutation path recalculating it, so the bill posted last period's numbers.

**The anchor.** The report projected from today while the bill priced through its due date, so a future-dated or between-occurrences payment made the two disagree by construction.

Installment pricing is now one rule, stated once as INV-LOAN-006:

```
debt(d)   = max(0, -(opening_balance + SUM(amount) over non-VOID, top-level
                     transactions on the loan dated <= d))
rate(d)   = latest loan_rate_changes row effective by d, else accounts.interest_rate
interest  = roundMoney(debt(d) * periodicRate(rate(d)))
```

Both inputs are dated at the installment's own due date, and `ScheduledTransactionLoanService.resolveInstallment` is the single path serving all three consumers — the template recalculation, the posting path (which now re-resolves the allocation inside the posting transaction, at the boundary the money actually moves) and the report (which anchors on `GET /scheduled-transactions/loan-anchor/:accountId`, the same due-date-bounded debt). Each one answering differently is the reported drift.

Supporting changes, each because a defect got through the gap it closed:

- **`backend/src/common/ledger-balance.sql.ts`** — eight service files already spelled the as-of ledger predicate out by hand, and pricing needed a ninth. It is now composed from shared fragments (`LEDGER_MOVEMENT_PREDICATE`, `ACCOUNT_BALANCE_AS_OF_SQL`), with `ledger-balance-sql.spec.ts` scanning the balance-reading modules for a re-hand-rolled copy. This is the one shared-area refactor in the PR, and it was asked for explicitly rather than taken on unprompted.
- **`effectiveAnnualRateOn`** with a truth table (`loan-rate-timeline-cases.json`) asserted by both layers, since backend and frontend cannot import each other and had drifted on which rate applies.
- **Serialization.** The debt read is a read-modify-write against a ledger a concurrent `TransactionsService.create` can move, so the posting path takes the existing `lockAccountsForBalanceWrite` over the source and loan accounts before pricing (CONC-001). `pricing-lock.guard.spec.ts` fails a `resolvePostingAllocation` call that is not preceded by the lock; `scheduled-loan-pricing-concurrency.integration.spec.ts` proves the blocking with two real connections.
- **Discriminated unions** (`LoanPostingDecision`, `ResolvedInstallment`) in place of an overloaded `null`: a paid-off loan, an unmanageable template shape and an unreadable ledger are three answers, and folding them made a retired loan post an installment. An unreadable ledger is a `ServiceUnavailableException`, never a priced guess.
- **An overdue anchor is refused**, falling back to today-anchored projection: the anchor's debt is measured *through* its date, so once that date is behind us everything the ledger did after it is invisible to the projection. Reconciling missed occurrences properly is a product decision, deliberately out of scope here.

Scope limits, recorded in the spec rather than left implicit: the recalculation **declines and writes nothing** for a template shape it does not understand (escrow, tax or insurance lines), because rewriting it as principal + interest + extra would break the parent-equals-children validator and stop the bill posting entirely; and `LINE_OF_CREDIT` is exempt from paid-off deactivation, since a zero balance is a normal state for a revolving line.

`docs/specs/scheduled-loan-installment-pricing.md` is the spec; INV-LOAN-006 in `docs/system-invariants.md` and the new rows in `docs/verification-contract.md` are its enforcement record.

## Checklist

Notes on the checklist:

- **Tests.** Backend 14,206 unit + 470 integration, frontend 14,424, all green; lint and typecheck clean on both layers. New coverage: two PostgreSQL integration suites (the dated-balance predicates and the two-connection lock protocol), four source-scan guards (`pricing-lock`, `ledger-balance-sql`, `loan-projection-anchor`, `loan-rate-timeline` contract) and unit matrices over the pricing path. Every new invariant was broken on purpose once to confirm its test fails.
- **i18n.** Two new backend keys (`accounts.loanLedgerUnreadable`, `scheduledTransactions.loanLedgerUnreadable`) translated across every locale, pseudo-locale regenerated. No new frontend strings.
- **Shared areas.** The ledger-SQL extraction described above is the only one, and it was explicitly requested rather than taken on unprompted. Flag it if you would rather it landed as its own PR.

## AI assistance disclosure

Written with Claude Code. The implementation, tests, spec and documentation in this PR were produced by it under direction, and it was also used to run adversarial review passes over its own diff — the findings from those passes, plus a separate human audit of PR #1278, are folded into the commits on this branch.


WMP:
# Monize PR #1278 Adversarial Audit

**Repository:** `kenlasko/monize`  
**Pull request:** #1278 — `fix: price scheduled loan installments from the dated ledger balance (#1253)`  
**Review verdict:** **REQUEST CHANGES**  
**Review snapshot:** 2026-08-29  
**Base branch:** `main`  
**BASE_SHA / current main / merge base:** `6ef5d8de900fdab9fa6307b2eb5d3fcdfef453f0`  
**PR_REVIEW_SHA:** `501224c47083c39a502c73237255a12f05257ac4`  
**PR state at close:** open, non-draft, mergeable, 4 commits ahead / 0 behind, 58 changed files, +3525 / -386.

## Executive summary

PR #1278 substantially improves the original #1253 defect. The dated ledger balance is now centralized, the backend and frontend share the rate-timeline truth table, posting re-prices managed loan templates at the consumption boundary, the Loan Amortization Report and Loan Detail schedule use a server-provided next-installment anchor, and a real-PostgreSQL integration test covers the inclusive due-date, future-row, VOID, and split-child predicates.

The core idea is sound, but three material correctness gaps remain:

| ID | Severity | Confidence | Summary |
| --- | --- | --- | --- |
| PR-1278-01 | **HIGH** | high | A fully paid-off loan can still post the entire stale scheduled installment because `paid-off` is collapsed to `null`, and `post()` interprets `null` as “use persisted template.” |
| PR-1278-02 | **MEDIUM** | high | A past-due projection anchor ignores real ledger activity after the overdue due date, so the report can append projected past rows after real history and overstate later interest/payoff. |
| PR-1278-03 | **MEDIUM** | high | The dated debt used for a financial posting is not protected by the serialization mechanism required by `CONC-001`; the schedule row lock does not serialize independent loan-ledger writers. |

No BLOCKER findings were confirmed. No security/RLS defect was confirmed in the new anchor endpoint.

The merge gate is **REQUEST CHANGES** because PR-1278-01 produces a persistent incorrect financial transaction after a real payoff, and PR-1278-02/03 leave the new financial invariant incomplete in realistic states the PR explicitly claims to cover.

---

## 1. Scope and method

This review was performed as a read-only adversarial PR audit. No repository files, branches, commits, reviews, comments, labels, or GitHub settings were modified.

The review pinned all implementation evidence to:

- base/main/merge base: `6ef5d8de900fdab9fa6307b2eb5d3fcdfef453f0`
- PR head: `501224c47083c39a502c73237255a12f05257ac4`

`main` was re-checked at audit close and still pointed to the base SHA, so the review did not silently mix a newer base revision into the PR.

The review followed the repository's `/audit` protocol and traced the material behavior across:

`scheduled occurrence -> posting controller/service -> transaction boundary -> loan installment resolver -> dated ledger SQL -> account/rate data -> frontend anchor -> projection engine -> report/detail UI -> tests/CI`

The review also challenged the implementation with payoff, overdue, post-anchor activity, rate-timeline, explicit override, joint-account, RLS, and concurrency counterexamples.

### Important execution limitation

A local checkout could not be obtained in this environment because direct GitHub DNS resolution failed (`Could not resolve host: github.com`). Therefore:

- no local grep was claimed;
- no local unit/integration test was claimed;
- no synthetic mutation was executed locally;
- no local two-connection PostgreSQL concurrency harness was executed.

Repository reads were instead performed through the authenticated GitHub connector at the pinned SHA, and hosted GitHub Actions results were inspected. The limitation matters most for PR-1278-03: the repository contract requires a real two-connection PostgreSQL test for a concurrency guarantee, and this PR does not contain one.

---

## 2. Repository instruction hierarchy reviewed

No `AGENTS.md` was found in the repository tree at the audited PR SHA.

Applicable instructions and contracts reviewed included:

- `CLAUDE.md`
- `backend/CLAUDE.md`
- `frontend/CLAUDE.md`
- `database/CLAUDE.md`
- `backend/src/mcp/CLAUDE.md`
- `README.md`
- `CONTRIBUTING.md`
- `.claude/commands/audit.md`
- `docs/concurrency-and-idempotency.md`
- `docs/financial-calculation-contract.md`
- `docs/financial-semantics.md`
- `docs/testing-contract.md`
- `docs/time-series-contract.md`
- `docs/row-level-security-contract.md`
- `docs/system-invariants.md`
- `docs/verification-contract.md`
- `docs/specs/scheduled-loan-installment-pricing.md`

The most relevant normative requirements are:

1. `CONC-001`: a value used for a destructive or financial write must be read inside the same transaction **and under the same lock, statement, or version guard** that authorizes the write.
2. A scheduled-loan installment must price debt and rate at its own due-date boundary (`INV-LOAN-006`).
3. Stored P/I splits are derived templates; posting must re-resolve stale derived values at the consumption boundary.
4. A posting may shrink to the remaining debt and must not overpay a retired debt into credit.
5. Financial edge cases require concrete numerical regression coverage and negative controls.

---

## 3. Change-surface overview

The PR's material changes fall into six groups.

### 3.1 Canonical dated ledger balance

`backend/src/common/ledger-balance.sql.ts` introduces:

- `LEDGER_EXCLUDES_VOID`
- `LEDGER_TOP_LEVEL_ONLY`
- `LEDGER_MOVEMENT_PREDICATE`
- `ledgerMovementPredicate`
- `ledgerBalanceJoin`
- `LEDGER_BALANCE_EXPRESSION`
- `ACCOUNT_BALANCE_AS_OF_SQL`

Existing balance readers in accounts, statement-cycle, action history, joint accounts, import post-processing, and net worth were converted from hand-written `(status != VOID) + top-level-only` clauses to the shared predicate.

The replacements inspected are mechanically equivalent to the previous predicates. No material regression was confirmed in this refactor.

### 3.2 Effective loan-rate timeline

`backend/src/accounts/effective-loan-rate.util.ts` provides the backend equivalent of the frontend effective-rate rule, using the latest row effective on or before the requested date and falling back to the account scalar.

A shared truth-table fixture is asserted by both backend and frontend tests. This closes the earlier “backend stale scalar vs frontend timeline” gap for managed installment pricing.

### 3.3 Posting-boundary re-pricing

`ScheduledTransactionLoanService.resolvePostingAllocation()` now calls the common installment resolver inside the posting transaction. `ScheduledTransactionsService.post()` locks the scheduled transaction, re-reads the override/split basis, and re-prices unchanged managed templates before the occurrence is claimed and the financial transaction is written.

Explicit typed amounts and stored override amounts remain user-authoritative and bypass re-pricing.

### 3.4 Projection anchor

`GET /scheduled-transactions/loan-anchor/:accountId` returns the next active scheduled installment date plus the ledger debt through that date.

The controller explicitly supports:

- owned accounts;
- delegated report access;
- jointly shared accounts, after a joint-account read authorization check and re-scope to the owner.

No cross-tenant/joint-account authorization defect was confirmed.

### 3.5 Frontend bill-anchored projections

The Loan Amortization Report and Loan Detail schedule pass the server anchor to `buildLoanProjectionInput()` and `resolveCurrentLoanTerms()`.

A source-scan guard enumerates bill-anchored vs today-anchored call sites.

### 3.6 Tests/docs/i18n

The PR adds/updates unit tests, a real-PostgreSQL dated-balance integration test, documentation/invariants, localized unreadable-ledger errors, and an unrelated test-only budget wall-clock flake fix.

---

# 4. Confirmed findings

## PR-1278-01 — Paid-off debt falls back to the stale scheduled installment

**Severity:** HIGH  
**Confidence:** high  
**Causality:** `PRE_EXISTING_BUT_IN_SCOPE` — the stale-template failure is the defect class this PR explicitly claims to close, but the retirement branch remains intentionally mapped back to legacy posting behavior.

### Evidence

Primary code:

- `backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts:295-323`
  - `resolvePostingAllocation()` throws for `unreadable`;
  - every other non-`ok` result, including `paid-off`, returns `null`.
- `backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts:571-584`
  - `debt <= 0.01` returns `{ kind: "paid-off", debt }`.
- `backend/src/scheduled-transactions/scheduled-transactions.service.ts:3038-3236`
  - the payload is only rewritten when `loanAllocation` is truthy;
  - `null` leaves the previously persisted split and parent amount untouched;
  - the occurrence is still claimed and `transactionsService.create()` is still called.
- `backend/src/scheduled-transactions/scheduled-transaction-loan.service.spec.ts:1600-1705`
  - the unit suite explicitly expects `resolvePostingAllocation()` to return `null` for retired debt, but does not exercise `ScheduledTransactionsService.post()` afterward.

The specification contradicts itself:

- `docs/specs/scheduled-loan-installment-pricing.md:92-101` says a posting may shrink and “retires the loan rather than overpaying it into credit.”
- `docs/specs/scheduled-loan-installment-pricing.md:120-132` says already-retired debt resolves nothing and “the posting proceeds unchanged.”

The implementation follows the second sentence, which violates the first invariant for the final occurrence.

### Violated invariant / expected behavior

A scheduled loan payment derived from debt must not move money after the debt through its posting boundary is already zero or credit.

`null` must not conflate:

- “this is not a managed loan template” and
- “this is a managed loan template whose authoritative debt is paid off.”

Those outcomes require different state transitions.

### Concrete numerical reproduction

Input:

- loan ledger debt before payoff: **$200,000.00**
- annual rate: **6%**
- monthly periodic rate: **0.5%**
- persisted scheduled parent: **$1,500.00**
- stale persisted split:
  - interest: **$1,000.00**
  - principal transfer: **$500.00**
- before the next due date, the user records a principal-only payoff of **$200,000.00**
- therefore `debt(dueDate) = $0.00`

Implementation result:

1. `resolveInstallment()` returns `paid-off`.
2. `resolvePostingAllocation()` converts that to `null`.
3. `post()` keeps the stale payload.
4. The source account is charged **$1,500.00**.
5. The loan receives another **$500.00** of principal, becoming **$500.00 in credit**.
6. **$1,000.00** of interest expense is recorded despite no debt remaining.

Expected result:

- financial movement for this occurrence: **$0.00**
- no new interest expense;
- no principal transfer into a paid-off debt;
- the occurrence must still be consumed deterministically so cron/manual retry does not repeat it;
- non-revolving amortizing schedules may then be deactivated by the existing post-recalculation rule;
- a LINE_OF_CREDIT may advance/consume the occurrence while remaining active.

Difference:

- parent cash movement: **$1,500.00 incorrect**
- interest: **$1,000.00 overstated**
- loan principal movement: **$500.00 overpayment**
- loan balance error immediately after posting: **$500.00 credit**

### User/data impact

This is persistent ledger corruption, not display drift. A user who pays off the debt independently before the scheduled occurrence can still be charged the old bill and can record non-existent interest.

Because the scheduled occurrence is then advanced/deactivated, the extra transaction may look final rather than obviously retryable.

### Recommended correction

Do not use `LoanPostingAllocation | null` to represent both “not applicable” and “paid off.”

Return a discriminated posting decision, for example:

```diff
-type LoanPostingResolution = LoanPostingAllocation | null;
+type LoanPostingResolution =
+  | { kind: "not-applicable" }
+  | { kind: "paid-off" }
+  | {
+      kind: "allocation";
+      amountsBySplitId: Map<string, number>;
+      parentAmount: number;
+    };
```

Then keep occurrence consumption inside the same posting transaction, but skip the financial write for a managed paid-off installment:

```diff
-const loanAllocation = await this.loanService.resolvePostingAllocation(...);
-if (loanAllocation) {
-  rewritePayload(loanAllocation);
-}
+const loanDecision = await this.loanService.resolvePostingAllocation(...);
+if (loanDecision.kind === "allocation") {
+  rewritePayload(loanDecision);
+}
+const skipFinancialWrite = loanDecision.kind === "paid-off";

 // Claim occurrence and consume/advance it atomically.

- await this.transactionsService.create(userId, transactionPayload);
+ if (!skipFinancialWrite) {
+   await this.transactionsService.create(userId, transactionPayload);
+ }
```

The exact amortizing-loan vs LINE_OF_CREDIT state transition should be retained from the existing recalculation semantics: a revolving LOC should not be permanently deactivated merely because its debt is zero today.

### Recommended regression test

Add a real posting-path test, not only a resolver unit test:

```diff
+it("consumes a retired scheduled-loan occurrence without writing money", async () => {
+  // persisted template: -1500 = -1000 interest + -500 principal
+  // authoritative debt through postDate: 0
+  await service.post(userId, scheduleId, unchangedEchoOrBaseOccurrence);
+
+  expect(financialTransactionsForOccurrence()).toHaveLength(0);
+  expect(interestExpenseDelta()).toBe(0);
+  expect(loanBalance()).toBe(0);
+  expect(sourceBalanceDelta()).toBe(0);
+  expect(occurrenceWasConsumed()).toBe(true);
+});
```

Also cover:

- auto-post/base-split path;
- manual Post dialog unchanged echo;
- exactly `$0.01` / just-above-threshold boundary;
- LINE_OF_CREDIT: occurrence consumed but facility remains active.

### Mutation check

Required mutation: convert the paid-off decision back to `null` / re-enable the old unconditional write.

The new regression must fail on:

- transaction count;
- source-account delta;
- loan balance;
- interest amount.

**Execution status:** mutation designed but not executed locally because no checkout was available.

### Verification method

Run the posting regression against PostgreSQL, reload the persisted transaction/split rows and both affected account balances after commit, then retry the same logical occurrence to confirm the durable occurrence claim prevents a second effect.

---

## PR-1278-02 — An overdue anchor discards real post-anchor ledger activity from the projection

**Severity:** MEDIUM  
**Confidence:** high  
**Causality:** `INTRODUCED_BY_PR`

### Evidence

- `frontend/src/lib/loan-history.ts:388-397`
  - any `{ nextDueDate, debt }` pair is accepted as a usable anchor; there is no “anchor must be today/future” restriction.
- `frontend/src/lib/loan-history.ts:589-653`
  - `startingDebt` is replaced with `anchor.debt`;
  - `firstPaymentDate` comes from the anchor;
  - later rate changes are filtered from **today**, not from the (possibly past) anchor date.
- `frontend/src/components/reports/LoanAmortizationReport.tsx:194-242`
  - all real historical events are appended first;
  - then a complete generated projection starting from the anchor is appended;
  - no real ledger movement after the anchor is folded into that generated balance path.
- `frontend/src/components/accounts/loan-detail/LoanDetailView.tsx:120-140`
  - the same anchored projection is used for the per-installment loan-detail schedule.
- `frontend/src/components/reports/LoanAmortizationReport.test.tsx:532-566`
  - the new test allows an August 15 anchor while the audit date is August 29, but its only real transaction is before the anchor; it therefore does not test the problematic “overdue anchor + later real principal movement” interaction.
- `backend/test/integration/scheduled-loan-dated-balance.integration.spec.ts`
  - proves the server returns debt as of the due date, which is correct, but does not test how the frontend reconciles later actual rows.

### Violated invariant / expected behavior

A forward amortization projection must represent one chronological state history.

If the server anchor is in the past, actual transactions and rate changes committed after that boundary cannot disappear from every subsequent projected balance.

A report must not show an actual later repayment and then continue projecting from a balance that never incorporated that repayment.

### Concrete numerical reproduction

Assume today is 2026-08-29.

Input:

- scheduled next due date: **2026-08-01** (overdue, auto-post off)
- authoritative debt through 2026-08-01: **$100,000.00**
- annual rate: **6%**
- monthly payment: **$1,000.00**
- real standalone principal repayment on 2026-08-15: **$20,000.00**
- current debt after that real repayment: **$80,000.00**

Implementation projection:

1. Anchor is accepted because it has both fields.
2. First projected August 1 row:
   - interest = `100,000 × 0.005` = **$500.00**
   - principal = **$500.00**
   - projected balance = **$99,500.00**
3. The real August 15 **$20,000.00** principal payment is displayed as a historical row, but it is never applied to the generated projection state.
4. September 1 projected interest:
   - `99,500 × 0.005` = **$497.50**

Expected chronological projection if the overdue bill is projected and the real August 15 principal movement is then applied:

- after projected August 1 bill: **$99,500.00**
- after real August 15 principal: **$79,500.00**
- September 1 interest:
  - `79,500 × 0.005` = **$397.50**
- after September 1 `$1,000` payment:
  - principal = **$602.50**
  - balance = **$78,897.50**

Implementation after September 1:

- interest: **$497.50**
- principal: **$502.50**
- balance: **$98,997.50**

Difference:

- September interest overstated by **$100.00**
- pre-September projected debt overstated by **$20,000.00**
- post-September projected debt overstated by **$20,100.00**

The report can also become visibly non-chronological: the actual August 15 history row is appended before the newly generated projected August 1 row.

A rate change effective between August 1 and today has the same root problem: the seed can start on August 1 while `futureTimeline` is built from today's date, dropping a rate step that should affect the overdue projected sequence.

### User/data impact

This does not write the database, but it can materially misstate:

- per-installment interest;
- remaining balance;
- payoff date;
- remaining lifetime interest;
- schedule chronology.

It affects exactly the user state created when an active recurring loan schedule is overdue and the user has made real payments, draws, or rate changes after the missed due date.

### Recommended correction

The product needs an explicit overdue-schedule model. Two safe approaches are acceptable:

1. **Full chronological catch-up:** interleave each overdue projected occurrence with actual ledger/rate events after its boundary before continuing to future rows.
2. **Fail closed for long-range projection:** if `nextDueDate < today`, show the overdue next-bill information but withhold payoff/remaining-interest projection until the occurrence is posted/skipped and the schedule advances.

The current behavior — silently generating a full schedule from a past anchor while displaying later actual history separately — should not remain.

A conservative merge-safe correction could classify the anchor:

```diff
 function usableProjectionAnchor(anchor) {
   if (!anchor || anchor.nextDueDate == null || anchor.debt == null) return null;
+  const today = todayYMD();
+  if (anchor.nextDueDate < today) {
+    return { kind: "overdue", ...anchor };
+  }
-  return { nextDueDate: anchor.nextDueDate, debt: anchor.debt };
+  return { kind: "current", ...anchor };
 }
```

and refuse the full projection for `kind: "overdue"` until a chronological catch-up implementation exists. The UI should explain why rather than silently falling back to a different financial answer.

### Recommended regression test

At minimum:

```diff
+it("does not project through an overdue anchor while ignoring later real principal", () => {
+  // today 2026-08-29
+  // anchor 2026-08-01, debt 100000
+  // real principal +20000 on 2026-08-15
+  // 6%, monthly, payment 1000
+
+  const result = build/report(...);
+
+  // If catch-up is implemented:
+  expect(septemberRow.interest).toBe(397.50);
+  expect(septemberRow.balance).toBe(78897.50);
+
+  // Or, for the fail-closed design:
+  // expect(fullProjection).toBeUnavailableBecauseScheduleIsOverdue();
+});
```

Add a companion rate-change case:

- anchor 2026-08-01;
- new rate effective 2026-08-15;
- assert the first installment on/after the effective date uses the new rate, or projection is withheld.

### Mutation check

Mutation: restore the present behavior that treats every non-null past anchor as a normal future seed and drops post-anchor actual events.

The new regression must fail numerically (`$497.50` vs `$397.50`) or fail the explicit overdue-unavailable assertion.

**Execution status:** mutation designed but not executed locally.

### Verification method

Use a fixed clock, render the Loan Amortization Report and Loan Detail schedule with a past anchor plus an actual post-anchor principal row, and assert chronological ordering plus the exact September interest/balance figures.

---

## PR-1278-03 — The posting debt read is not serialized with independent loan-ledger writers

**Severity:** MEDIUM  
**Confidence:** high  
**Causality:** `PRE_EXISTING_BUT_IN_SCOPE` — the PR newly claims that re-resolution happens at the financial consumption boundary, but the lock it names protects the schedule/template, not the ledger aggregate that is consumed.

### Evidence

Normative contract:

- `docs/concurrency-and-idempotency.md:185-195` (`CONC-001`) requires a financial input to be read in the same transaction **and under the same lock, statement, or version guard** that authorizes the write.

Posting path:

- `backend/src/scheduled-transactions/scheduled-transactions.service.ts:2943-2953`
  - locks the `ScheduledTransaction` row with `pessimistic_write`.
- `backend/src/scheduled-transactions/scheduled-transactions.service.ts:3058-3102`
  - calls `resolvePostingAllocation()` after the schedule/template checks.
- `backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts:478-494`
  - `datedLoanDebt()` performs an aggregate ledger query; it does not lock the loan account or otherwise guard the predicate being read.
- `backend/src/common/db/scoped-db.ts:160-240`
  - ordinary `withScopedDb` uses the connection default isolation; the repository documents it as READ COMMITTED.
- `backend/src/transactions/transactions.service.ts:283-431`
  - an independent transaction write uses its own transaction and does not take the scheduled-transaction parent lock.
- `backend/src/accounts/accounts.service.ts:1048-1077` and `1133-1165`
  - the existing balance protocol already serializes balance writers on the account row (`UPDATE` row lock or `lockAccountsForBalanceWrite()` before absolute recomputation).
- `backend/src/common/db/locks.ts:137-174`
  - `lockAccountsForBalanceWrite()` is the repository's existing account-row serialization primitive.

The parent schedule lock is real and useful for occurrence/template races, but it does not protect the aggregate `transactions` predicate that determines `debt(d)`.

### Violated invariant / expected behavior

The P/I allocation written by a scheduled occurrence must be based on a debt value protected by the same serialization domain as other writes that can change that debt.

The PR description says principal-only payments, voids, and imports committed between occurrences reach the amount actually posted. At the consumption boundary, a concurrent committed ledger mutation must not slip between the aggregate read and the financial write without either:

- being ordered after the posting by a shared lock/protocol; or
- forcing the posting to re-read/retry.

### Concrete numerical reproduction

Initial state:

- debt through due date: **$200,000.00**
- rate: **6%**, monthly = **0.5%**
- scheduled parent: **$1,500.00**

Interleaving:

1. T1 starts scheduled `post()`, locks the **schedule row**.
2. T1 reads loan debt = **$200,000.00**.
3. T1 computes:
   - interest = **$1,000.00**
   - principal = **$500.00**
4. T2 independently writes a principal-only repayment of **$50,000.00** dated on/before the same due date and commits. T2 does not need T1's schedule-row lock.
5. T1 continues and commits its already-computed split.

After T2's repayment, the debt basis relevant to the due-date ledger is **$150,000.00**.

Expected scheduled allocation at 6% / 12:

- interest = `150,000 × 0.005` = **$750.00**
- principal = **$750.00**

Implementation allocation:

- interest = **$1,000.00**
- principal = **$500.00**

Difference:

- interest overstated by **$250.00**
- principal understated by **$250.00**
- remaining loan debt after the occurrence is **$250.00 higher** than under the expected allocation.

The parent cash payment is still `$1,500`; the corruption is the financial classification and loan-principal movement.

### User/data impact

The race window is narrower than PR-1278-01, but the result is a persistent P/I misallocation. It can be triggered by two normal operations: posting the scheduled bill while another request records/edits/voids/imports a principal movement.

The current real-PostgreSQL integration test proves SQL predicate semantics but does not run two connections with a controlled interleaving, so it does not test this guarantee.

### Recommended correction

Use a serialization mechanism that protects the **loan account / ledger decision**, not only the scheduled template.

The repository already has `lockAccountsForBalanceWrite()`. A likely direction is to pre-resolve the relevant account ids, acquire the canonical account locks before the ledger read, and hold them through the financial write, while preserving the repository's global lock ordering and then re-validating the schedule under its own lock.

Illustrative direction:

```diff
 const removedAfterOnce = await withScopedDb(this.dataSource, async (m) => {
+  // Resolve the account ids needed for this financial decision from the
+  // pre-lock snapshot, lock them in canonical order, then validate that the
+  // schedule still names the same shape after its own row lock.
+  await lockAccountsForBalanceWrite(
+    m,
+    [scheduled.accountId, preResolvedLoanAccountId].filter(Boolean),
+    userId,
+  );
+
   const current = await m.findOne(ScheduledTransaction, {
     where: { id, userId },
     lock: { mode: "pessimistic_write" },
   });

   // mutation-basis checks...

   const loanDecision = await loanService.resolvePostingAllocation(...);
   // financial write while account lock remains held
 });
```

This is intentionally illustrative rather than a drop-in patch. Before implementation, inventory every path that can create/update/delete/void/import loan-account ledger rows and confirm the chosen account/FK/row-lock protocol serializes them. Do not add a one-sided lock that some ledger writer can bypass.

An equivalent SERIALIZABLE/version-guard design is acceptable if it is demonstrably enforced and retried, but the repository already has a named account-balance lock primitive, so reusing the established protocol is preferable if lock ordering can be maintained safely.

### Recommended regression test

Add a **real PostgreSQL two-connection integration test** with barriers:

```diff
+it("does not post a split from debt that changed concurrently", async () => {
+  // T1: begin scheduled post; stop after debt SELECT.
+  // T2: insert/commit +50000 principal dated <= due date.
+  // Release T1.
+
+  // Acceptable outcomes:
+  // A) T2 blocks behind the pricing lock; T1 linearizes first, then T2.
+  // B) T1 re-reads/retries and posts 750 interest / 750 principal.
+  // C) one side receives an explicit retryable serialization conflict.
+  //
+  // Forbidden: both commit with T1 posting 1000 / 500 while the protocol
+  // claims the later committed principal was inside its pricing boundary.
+});
```

The test must use two physical database connections/query runners. Mocked repository calls do not prove PostgreSQL blocking.

### Mutation check

Mutation: remove/bypass the shared account/ledger serialization guard while retaining the transaction and schedule-row lock.

The two-connection test must then reproduce the stale split or a forbidden interleaving.

**Execution status:** mutation designed but not executed locally.

### Verification method

Observe blocking/retry behavior in PostgreSQL and assert persisted final rows, not only service return values:

- occurrence claim;
- principal transaction;
- interest transaction/split;
- source account balance;
- loan account balance;
- schedule advancement.

---

# 5. Design risks requiring a decision

## DR-1278-01 — Custom loan templates with escrow/tax/insurance remain outside INV-LOAN-006 enforcement

**Status:** design risk, not promoted to a confirmed defect because the new spec explicitly declares the shape unsupported.

`backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts:616-653` only manages:

- principal transfer;
- one identifiable interest category line;
- optional extra-principal transfer.

Any additional unmanaged line causes `kind: "declined"`, which makes posting continue on the persisted template. Therefore a mortgage template with principal + interest + escrow can still carry stale P/I after a principal-only movement.

This is a reasonable fail-safe against rewriting a line whose meaning the service cannot prove, but the invariant language (“scheduled loan installment”) is broader than the enforced shape.

**Decision needed:** either narrow `INV-LOAN-006` and user-facing expectations to “managed P/I(/extra) templates,” or define how fixed escrow/tax/insurance lines participate in the parent total so P/I can still be safely re-priced.

---

# 6. Missing tests

The following test gaps are material to this PR:

1. **Paid-off posting path** — current resolver test stops at `null`; no test proves the financial write is skipped.
2. **Manual unchanged-echo paid-off path** — same root through the Post dialog.
3. **Overdue anchor + post-anchor principal movement** — current anchor tests do not combine these states.
4. **Overdue anchor + rate change between anchor and today.**
5. **Chronological ordering of actual and projected rows when the anchor is in the past.**
6. **Two-connection PostgreSQL serialization test** for a loan-ledger mutation racing scheduled posting.
7. **Mutation test evidence** for the three merge-blocking invariants. The repository `/audit` protocol requires the mutation to demonstrate the regression test is discriminating.
8. **Custom escrow/template decision test** if that shape is intended to be covered by INV-LOAN-006.

The new PostgreSQL integration test is valuable, but it proves the as-of SQL's set semantics, not the complete posting state machine or its concurrency protocol.

---

# 7. Documentation issues

## DOC-1278-01 — Paid-off semantics are contradictory inside the new spec

`docs/specs/scheduled-loan-installment-pricing.md:92-101` says the posting may shrink so an occurrence retires the loan rather than overpaying it into credit.

The same document at `:120-132` says a debt already retired through the boundary resolves nothing and the posting proceeds unchanged.

These cannot both be true for a non-zero persisted template. The code implements the latter and produces PR-1278-01.

## DOC-1278-02 — Backend anchor docstring still describes the old stale-rate gap

`backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts:350-365` states that the service still prices from `accounts.interest_rate` and that rate parity is not fixed.

The same file now calls `datedAnnualRate()` and `effectiveAnnualRateOn()` against `loan_rate_changes`, so the comment is stale and directly contradicts the implementation.

## DOC-1278-03 — Projection-surface description disagrees with the source guard

`docs/specs/scheduled-loan-installment-pricing.md` says the “other projection surfaces (loan detail payoff, Debt Payoff Timeline, Overpayment Simulator)” keep today-anchored semantics.

The new `frontend/src/lib/loan-projection-anchor.guard.test.ts` explicitly places `LoanDetailView.tsx` in `BILL_ANCHORED`, and the account page fetches and passes the anchor to it.

The guard/code should win; update the prose so future reviewers do not “fix” the implementation back toward a stale description.

---

# 8. Rejected hypotheses / false positives investigated

| Hypothesis | Result | Evidence / reason |
| --- | --- | --- |
| New loan-anchor endpoint breaks joint-account reports | **Rejected** | Controller performs a joint-account membership/read authorization check and re-reads in owner scope. |
| Delegated report users cannot access the new endpoint | **Rejected** | Endpoint is annotated with `@AllowDelegate()` and `@DelegateRequiresSection("reports")`. |
| Anchor keeps the original recurrence date after an occurrence override moves it | **Rejected** | Query uses `COALESCE(override_date, next_due_date)`; integration coverage includes the moved date. |
| Backend still prices the bill from stale `accounts.interest_rate` | **Rejected in code** | `datedAnnualRate()` reads `LoanRateChange` rows and calls shared `effectiveAnnualRateOn()`. Only the docstring is stale. |
| Dated ledger balance counts VOID transactions | **Rejected** | Shared predicate excludes VOID; PostgreSQL integration test covers it. |
| Dated ledger balance double-counts split children | **Rejected** | Shared predicate requires `parent_transaction_id IS NULL`; integration test covers it. |
| Posting grows the parent back to `account.paymentAmount` | **Rejected** | Posting-purpose resolver uses the displayed template total and only re-divides/shrinks it; unit coverage exists. |
| User-typed inline/override amount is silently overwritten by re-pricing | **Rejected** | Re-pricing is limited to base templates or unchanged echoed Post-dialog values; explicit changes bypass it. |
| Empty-alias ledger helper emits invalid `t.` references in net-worth queries | **Rejected** | `ledgerMovementPredicate("")` intentionally emits bare column names. |
| The shared ledger predicate refactor changed existing VOID/top-level behavior | **Rejected** | Inspected call-site patches are mechanical replacements of the same two predicates. |
| The budget test change alters production behavior | **Rejected** | Only `budgets.service.spec.ts` changes; it clamps a date used by a wall-clock-sensitive test. |

---

# 9. Adversarial state matrix

| State | Reviewed outcome |
| --- | --- |
| Normal monthly installment | Improved; common resolver produces debt × periodic rate. |
| Principal-only payment before due date | Improved in non-concurrent managed-template path. |
| Principal-only payment exactly on due date | Covered by PostgreSQL integration; included. |
| Principal-only payment after due date | Correctly excluded from that due-date debt, but becomes problematic when the due date is overdue and frontend projects beyond it (PR-1278-02). |
| Extra principal split | Managed by resolver and waterfall; unit coverage present. |
| Final partial payment | Resolver waterfall can shrink parent; coverage present for small remaining debt. |
| **Full payoff before due date** | **Broken: PR-1278-01.** |
| Rate change before/on due date | Backend now uses dated timeline; shared truth-table coverage. |
| Future rate change | Not applied early by `effectiveAnnualRateOn`. |
| Rate change after an overdue anchor but before today | **Projection gap: PR-1278-02.** |
| Override moves occurrence date | Backend prices at `postDate`; anchor query uses override date. |
| Explicit override amount | Preserved as user instruction. |
| Manual Post dialog unchanged echo | Re-priced; intended parity with auto-post. |
| VOID ledger movement | Excluded from debt. |
| Split child | Excluded from debt. |
| Duplicate scheduled posting | Existing durable occurrence claim + schedule lock retained. |
| Concurrent independent principal mutation | **Serialization gap: PR-1278-03.** |
| Joint account | Authorization path reviewed; no defect confirmed. |
| Delegated report access | Annotation and report section guard reviewed; no defect confirmed. |
| Custom escrow/tax/insurance split | Explicitly declined; tracked as design risk DR-1278-01. |
| LINE_OF_CREDIT zero debt | Requires “consume without payment but remain revolving” semantics in PR-1278-01 fix. |

---

# 10. Mutation matrix

The repository's `/audit` protocol requires a counterfactual mutation for each merge-blocking invariant. Local mutations could not be executed because a checkout was unavailable, so the table records the required mutations and the tests that must kill them before merge.

| Finding | Mutation | Test that must fail | Current status |
| --- | --- | --- | --- |
| PR-1278-01 | Collapse `paid-off` back to `null` and allow unconditional `transactionsService.create()` | Paid-off full posting-path PostgreSQL/service test asserting zero financial movement | **Missing / not executed** |
| PR-1278-02 | Accept any past anchor as a normal projection seed while ignoring post-anchor real events | Fixed-clock overdue-anchor + Aug 15 principal test (`$397.50` expected vs `$497.50` stale) | **Missing / not executed** |
| PR-1278-03 | Remove/bypass shared loan-ledger serialization while leaving schedule-row lock | Two-connection PostgreSQL barrier test | **Missing / not executed** |

The existing tests are evidence of the intended happy paths, but they do not kill these mutations.

---

# 11. CI and external review status

## Hosted CI observed at `PR_REVIEW_SHA`

At the final observation:

- PR Checklist: **success**
- Lighthouse CI: **success**
- CI Pipeline: still marked **in progress**
- Backend Lint & Type Check: **success**
- Backend Integration Tests: **success**
- Frontend Unit Tests: **success**
- Backend Unit Tests: **success**
- NPM Audit: **success**
- Frontend Lint / Type Check / i18n: **success**
- Helm lint/render: **success**
- License checks: **success**
- Frontend Bundle Size: **success**
- Bearer Security Scan: **success**
- zizmor workflow scan: **success**
- Schema vs Migrations Drift: **success**
- Dockerfile lint: **success**
- Documentation vs Manifests: **success**
- E2E Browser Cache: **success**
- E2E shards 1/4, 3/4, 4/4: **success**
- E2E shard 2/4: the `Run E2E tests` step was **success**; the job was still finishing post-action cleanup when checked.

No failing hosted check was observed.

## External review ledger

- Human submitted reviews: none observed.
- Inline review threads: none observed.
- PR conversation: deployment/bot activity only; no external technical finding was available to classify.

Therefore no finding in this report is inherited from another reviewer.

---

# 12. Audit ledger

## Directories inspected

Material paths inspected at the pinned PR SHA included:

- `backend/src/accounts/`
- `backend/src/common/`
- `backend/src/common/db/`
- `backend/src/scheduled-transactions/`
- `backend/src/transactions/`
- `backend/src/loan-rate-changes/`
- `backend/src/delegation/`
- `backend/src/import/`
- `backend/src/net-worth/`
- `backend/src/action-history/`
- `frontend/src/lib/`
- `frontend/src/app/accounts/[id]/`
- `frontend/src/components/accounts/loan-detail/`
- `frontend/src/components/reports/`
- `frontend/src/types/`
- `docs/`
- `docs/specs/`
- repository root instructions/docs.

## Files inspected in full or by all material sections

- `.claude/commands/audit.md`
- `CLAUDE.md`
- `backend/CLAUDE.md`
- `frontend/CLAUDE.md`
- `database/CLAUDE.md`
- `backend/src/mcp/CLAUDE.md`
- `README.md`
- `CONTRIBUTING.md`
- `docs/concurrency-and-idempotency.md` (material normative sections)
- `docs/financial-calculation-contract.md` (material financial/write-test sections)
- `docs/financial-semantics.md` (material loan/split/sign sections)
- `docs/testing-contract.md` (material financial/concurrency sections)
- `docs/time-series-contract.md` (material date-boundary sections)
- `docs/row-level-security-contract.md` (material ownership/RLS sections)
- `docs/specs/scheduled-loan-installment-pricing.md`
- `backend/src/common/ledger-balance.sql.ts`
- `backend/src/accounts/effective-loan-rate.util.ts`
- `backend/src/scheduled-transactions/scheduled-transaction-loan.service.ts` (material resolver/anchor/rate/debt sections)
- `backend/src/scheduled-transactions/scheduled-transactions.service.ts` (material posting transaction)
- `backend/src/common/db/scoped-db.ts` (transaction/isolation behavior)
- `backend/src/common/db/locks.ts` (account serialization primitive)
- `backend/src/accounts/accounts.service.ts` (balance writer protocol)
- `backend/src/transactions/transactions.service.ts` (create/write path)
- `frontend/src/lib/loan-history.ts` (anchor/seed/history material sections)
- `frontend/src/lib/loan-schedule.ts` (projection engine material sections)
- `frontend/src/components/reports/LoanAmortizationReport.tsx` (history + projection composition)

## Changed-file patches inspected

Material patches inspected included:

- account balance/report/forecast call sites;
- statement-cycle;
- action-history;
- joint-account balance reads;
- import post-processing;
- net-worth;
- scheduled-loan service;
- scheduled posting service;
- scheduled controller;
- PostgreSQL dated-balance integration test;
- loan resolver/service unit tests;
- frontend account page;
- Loan Detail;
- Loan Amortization Report;
- loan history/tests;
- projection-anchor source guard;
- frontend scheduled-transactions API/type;
- shared ledger guard test;
- effective rate truth table/tests;
- budget flake test.

## Sampled files

- locale updates were sampled in `en` and `pl`; the full locale set was not individually language-reviewed.
- several test files were inspected by relevant changed hunks rather than every pre-existing test case.
- mechanical `LEDGER_MOVEMENT_PREDICATE` replacements were inspected at the changed call sites; unrelated surrounding service code was not treated as fully reviewed.

## Inaccessible / not executed

- local repository checkout;
- local test suite;
- local mutation tests;
- local PostgreSQL two-connection experiment.

Reason: environment could not resolve `github.com` over direct DNS. GitHub connector reads remained available.

## Tests/checks examined

- hosted backend unit suite;
- hosted backend integration suite;
- hosted frontend unit suite;
- hosted lint/type/i18n;
- hosted security/schema/manifests checks;
- hosted E2E job status;
- new real-PostgreSQL `scheduled-loan-dated-balance.integration.spec.ts`;
- changed resolver, posting, report, and anchor tests.

## Areas reviewed without a material finding

- canonical VOID/top-level ledger predicate refactor;
- effective rate timeline parity;
- Canadian/variable-rate plumbing at the changed boundary;
- explicit override preservation;
- manual Post dialog unchanged-echo detection;
- moved override date for the anchor;
- joint-account anchor authorization;
- delegated report access;
- RLS scoping of the new controller/service path;
- frontend API/type wiring;
- i18n key presence in sampled locales;
- unrelated budget test-only date clamp.

---

# 13. Recommended remediation order

1. **PR-1278-01 first.** Split the “not applicable” vs “paid off” state and make the occurrence consumption path capable of advancing/consuming without a financial write.
2. **PR-1278-03 next.** Establish the real serialization boundary around the dated debt read before declaring INV-LOAN-006 “enforced.” Add the two-connection PostgreSQL test at the same time.
3. **PR-1278-02.** Decide and encode overdue projection semantics; either implement chronological catch-up or fail closed until the overdue occurrence advances.
4. Reconcile the contradictory documentation and narrow/clarify the custom-template scope.
5. Run the required mutation checks and preserve the existing green unit/integration/E2E matrix.

---

# 14. Final adversarial challenge

Before issuing this verdict, the three retained findings were re-checked against the pinned PR head rather than inferred from the PR description:

- PR-1278-01 survives because the exact paid-off branch returns `paid-off`, the public posting resolver converts it to `null`, and the posting service only rewrites on a truthy allocation before an unconditional financial create.
- PR-1278-02 survives because the frontend explicitly accepts a past anchor, seeds the full projection from the past debt, appends that projection after all actual history, and has no reconciliation of later real ledger rows.
- PR-1278-03 survives because the only lock protecting the new calculation is the scheduled-row lock, while `CONC-001` requires the financial input itself to be protected by the same serialization guard; the new PostgreSQL test is single-flow SQL semantics, not a two-connection lock test.

No later phase/review evidence or hosted test disproved these scenarios. The current green tests simply do not exercise them.

# Final verdict

## **REQUEST CHANGES**

PR #1278 is directionally strong and fixes the principal #1253 drift for the common managed-template path, but it should not merge while a full payoff can still create an entire stale financial transaction. The overdue-projection and ledger-serialization gaps should also be addressed before INV-LOAN-006 is described as enforced end-to-end.

PR #1299, opened 2026-09-02
Fix elicitation capability check for SDK >= 1.23

kenlasko:
## Linked discussion / issue

Closes #
Approved in: 

## Summary

Fixes a regression where MCP clients that advertise `elicitation` capability but answer `-32601` (method not found) or never respond are incorrectly treated as capable of showing dialogs. This caused write confirmations to be refused or hang until the client's own tool deadline expired.

The root cause: `@modelcontextprotocol/sdk` >= 1.23 normalizes the legacy capability shape `{"elicitation":{}}` into `{"elicitation":{"form":{}}}` before the capability check sees it, so every client advertising elicitation now looks form-capable — including ones that reject the method or drop the connection. The advertised capability alone cannot distinguish them.

**Solution:** Track observed client behavior per session. Only a returned `action` (accept/decline/cancel) proves a human is behind the dialog. Clients that answer for themselves (reject the method, timeout, or never reply) are recorded as "silent" and skip future confirmation rounds, falling back to the client's own per-tool approval. Clients that have already shown a dialog are not demoted by a single failure.

The timeout is also reduced from 5 minutes to 45 seconds to stay under Claude's 60-second tool deadline, preventing opaque client-side timeouts when a dialog goes unanswered.

## Changes

- **`mcp-elicitation-support.ts`** (new): Tracks per-session elicitation behavior using a `WeakMap` keyed on `McpServer`. Exports helpers to record observed behavior and detect when a client answered for itself (specific error codes: `MethodNotFound`, `RequestTimeout`, `ConnectionClosed`, `ParseError`, `InvalidRequest`, `InvalidParams`).
- **`mcp-context.ts`**: Updated `confirmWrite()` to consult observed behavior before asking a "silent" client again, and to distinguish between a user's refusal and a client that never answered. Reduced `CONFIRM_TIMEOUT_MS` from 5 minutes to 45 seconds. Expanded docstring to explain the SDK normalization regression and why behavior, not capability, is the source of truth.
- **`mcp-context.spec.ts`**: Added comprehensive test suite covering:
  - Clients that answer for themselves (6 error codes tested)
  - Session memory (a silent client is not asked again; behavior is per-session)
  - Clients that have shown a dialog (one unanswered dialog refuses the write but does not demote the client)
  - Timeout validation (confirms it stays under 60s but above 20s)
  - Regression pin: verifies the SDK's normalization of `{"elicitation":{}}` to `{"elicitation":{"form":{}}}`
- **`CLAUDE.md`**: Updated tool checklist to clarify that `"unsupported"` means no dialog reached a human, not a lack of capability.

## Checklist

- [x] This PR addresses a **single concern** (elicitation capability regression).
- [x] No user-facing strings added (internal behavior only).
- [x] No shared/core areas refactored without prior agreement.

## AI assistance disclosure

https://claude.ai/code/session_019eMq2jHZSWbbqD3KAJK4b2

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -5919,9 +5919,9 @@
       "license": "MIT"
     },
     "node_modules/fast-uri": {
-      "version": "3.1.5",
-      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.5.tgz",
-      "integrity": "sha512-gHwA1O9LDIcKunMKhObS/HimwtehO1nPUECKAu5TpKgaO19fcWEl4bliWe1jWxVFvIXztJjjQ4L8XQ1EU9f7Jw==",
+      "version": "3.1.7",
+      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.7.tgz",
+      "integrity": "sha512-dOvZVzjdZdz7phd9v6jCbwxrBW3fK6n8Rc0CtdmM4bumzMnxywBYhuph6J819RRw/ku+rLbelwfMunktuzVVHg==",
       "funding": [
         {
           "type": "github",
@@ -8482,6 +8482,8 @@
     },
     "node_modules/object-inspect": {
       "version": "1.13.4",
+      "resolved": "https://registry.npmjs.org/object-inspect/-/object-inspect-1.13.4.tgz",
+      "integrity": "sha512-W67iLl4J2EXEGTbfeHCffrjDfitvLANg0UlX3wFUUSTx92KXRFegMHUVgSqE+wvhAbi4WqjGg9czysTV2Epbew==",
       "license": "MIT",
       "engines": {
         "node": ">= 0.4"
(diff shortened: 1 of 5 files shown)


kenlasko:
**`NPM Audit` is red, and it is not this PR's failure — but it does block the merge, and the fix is a decision for you.**

This PR changes four files (`backend/src/mcp/*`), none of them a manifest: `git diff 1723a96 -- backend/package-lock.json backend/package.json frontend/package-lock.json frontend/package.json` is empty, so the audited tree is byte-identical to the base commit. Two advisories were published against dependencies already on `main` since main's last CI run (run 2946, 15:37 UTC, green):

| Package | Sev | Range | Path |
|---|---|---|---|
| `fast-uri` | high | 3.0.0 – 3.1.5 | transitive (4 advisories: host confusion via IDN canonicalization, SSRF via IPv6 normalization, SSRF via repeated percent-decoding, host confusion via percent-encoded scheme) |
| `qs` | moderate | 2.2.5 – 6.15.3 | transitive (array-limit bypass, DoS via attacker-controlled `isBuffer`) |

`main` would fail the same job today. The frontend half is clean (`npm audit --audit-level=high --omit=dev` → 0 vulnerabilities), and every other completed check on this head is green, including CodeQL, Bearer and Backend Lint & Type Check.

No re-run will clear it — the advisory database is the input, so it is deterministic, and I have not spent one.

**Proposed patch**, verified locally (`cd backend && npm audit fix --package-lock-only --omit=dev` → `found 0 vulnerabilities`). Lockfile only, no `package.json` change, transitive bumps only, +23/-12 lines:

- `fast-uri` 3.1.5 → 3.1.7
- `qs` 6.15.2 → 6.16.0
- `side-channel` 1.1.0 → 1.1.1, `side-channel-list` 1.0.0 → 1.0.1, plus `object-inspect` / `side-channel-map` / `side-channel-weakmap` entries pulled in by the `qs` bump

I have **not** pushed it. A dependency bump is repo-wide rather than scoped to this change — the `qs` minor lands under Express's body parsing for every route — and this PR's own checklist asks for a single concern. It is equally reasonable to want it here so this PR can go green rather than waiting for a separate bump to land on `main`; say the word and I will push it as its own commit on this branch.

---

kenlasko:
Pushed the patch described above as `3e1f5ac`, at the author's request — lockfile only, `backend/package-lock.json` +23/−12, `package.json` untouched, no dependency added or removed.

- `fast-uri` 3.1.5 → 3.1.7
- `qs` 6.15.2 → 6.16.0
- `side-channel` 1.1.0 → 1.1.1, `side-channel-list` 1.0.0 → 1.0.1 (what `qs` 6.16.0 asks for), plus the `resolved`/`integrity` fields npm restored on the entries it rewrote

Verified on the bumped tree before pushing: `npm ci` installs, `npm audit --audit-level=high --omit=dev` reports **0 vulnerabilities**, `tsc --noEmit` clean, `eslint` clean, backend unit suite green (530 suites, 14,322 tests). The integration suite needs a database this environment does not have, so CI covers that half — `qs` sits under Express body parsing, so `Backend Integration Tests` is the run worth watching on this head.

---

PR #1303, opened 2026-09-03
Reorder and group the Settings Preferences section

kenlasko:
## Linked discussion / issue

None. This was requested directly by the repository owner in a Claude Code session; no discussion or issue was opened first. Flagging rather than leaving the section blank — the propose-first workflow is unmet here.

## Summary

The Preferences card rendered sixteen controls as flat siblings in an order that had grown by accretion: Default Currency sat between the colour theme and the exchange pickers, the two investment controls were wedged between the date and number formats, and the two transaction toggles were split apart by the map provider. With no sub-headings it read as one sixteen-row wall.

This puts the controls in the requested order and splits them into six labelled sub-groups, using the `<h3 className="text-sm font-semibold …">` sub-heading shape the other settings cards already use (`AutoBackupSection`, `BackupRestoreSection`, `DangerZoneSection`). The outer container moves from `space-y-4` to `space-y-6` so groups separate more than the rows within them.

| Group | Controls |
|---|---|
| Language & Region | Language, Default Currency |
| Appearance | Theme, Colour theme |
| Dates & Numbers | Date Format, Number Format, Timezone, Week starts on, Show Create Date, Time Format |
| Investments | Default Stock Quote Provider, Preferred Exchanges |
| Transactions | Recent transactions in quick-fill, Lock reconciled transactions |
| Application | Map provider for addresses, Show What's New |

**Time Format stays gated on `showCreatedAt` and directly beneath it**, inside Dates & Numbers, rather than moving several rows above the toggle that drives it — a control that appears and disappears away from its trigger reads as a glitch.

No preference semantics change: the bulk-save payload is untouched, and language, theme and colour theme still self-persist outside it. The What's New tour anchor and the MSN not-configured alert travel with their controls. No group wrapper carries `w-full`, which three timezone tests select on.

**The test finding.** All 36 existing tests stayed green through the reorder, because every one of them asserts by accessible label — the suite had no case for control order at all. Added one: a guard that reads the rendered labels and group headings back out of the DOM in document order and asserts the exact sequence, plus cases pinning Time Format's pairing with its toggle and the localized option labels. Verified red by swapping two controls in the component (exactly one test failed), then reverted.

Also localized the Time Format options, which were hardcoded English literals (`'24-hour (14:30)'`, `'12-hour (2:30 PM)'`) on lines this change was already moving. `buildTimezoneOptions()`'s untranslated fallback labels are the same class of defect but are pre-existing and untouched here.

Two commits: the implementation with the English catalog, then the translation pass across the eighteen full locales.

## Checklist

- [ ] An approved discussion or issue exists and is linked above. — **not met**, see above.
- [x] New behavior has tests, and the existing suite passes. — 760 files / 14,805 tests pass, plus `lint`, `type-check` and `i18n:check`.
- [x] All user-facing strings are translated for **every** locale (i18n parity). — eight new keys across all locales; pseudo-locale regenerated; `messages.parity.test.ts` green.
- [x] No shared/core areas were refactored without prior agreement. — confined to `PreferencesSection.tsx`, its test, and the `settings` catalogs.

## AI assistance disclosure

Written by Claude Code, driven by the repository owner: the target ordering and the decision to keep Time Format paired with its toggle were the owner's; the grouping, the guard test and the i18n fix were proposed by Claude and approved before implementation.

https://claude.ai/code/session_01Uho97RieEnjTopzK8x2VwV

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uho97RieEnjTopzK8x2VwV)_


kenlasko:
**CI status:** everything green except **Workflow Security Scan (zizmor)**, which is not this PR's failure. E2E shards 1-4 are still running.

zizmor never ran. The `astral-sh/setup-uv` step failed fetching its version manifest, and the upload step then failed on the artifact that step never produced:

```
Fetching manifest data from https://raw.githubusercontent.com/astral-sh/versions/main/v1/uv.ndjson ...
##[error]fetch failed
...
##[error]Path does not exist: zizmor.sarif
```

Three things put it outside this change:

- That job scans `.github/workflows/`, and this PR touches no workflow files — `git diff origin/main...HEAD -- .github/` is empty.
- zizmor passed on `main` at `c4d598a`, the exact commit this branch is based on (run 2966).
- The job died in setup, before the scan body ran — a transient network failure reaching raw.githubusercontent.com, not a finding.

There is no fix to port; nothing in the repo is wrong. I tried the one permitted re-run and GitHub refused it with `403 This workflow is already running` because the E2E shards are still in flight. I'll re-run the failed job once the run settles and report back. If it fails a second time on the same fetch, that is an upstream availability problem rather than something this PR can address, and I'll say so rather than retry further.

---

PR #1304, opened 2026-09-03
Notification delivery: Web Push, per-category preferences, reminders

WMP:
## Linked discussion / issue

Closes # 
Approved in: #1291

## Summary

Gives the notification centre a delivery layer and a per-category preference
matrix, and renames the table it all hangs off. The bell keeps showing every
notification; what this adds is **which of them Monize also pushes or emails,
how often, and how a user stops one**.

Seven phases, in the order they were agreed in #1291:

**1. Web Push transport.** `push_instance_config` holds one VAPID key pair per
instance (private half encrypted under `ENCRYPTION_KEY`, generated on a
bootstrap hook, rotatable from `/admin/notifications`); `push_subscriptions`
holds one row per browser profile. `WebPushSender` is the only file in `src/`
that imports `web-push` — `push-secret.guard.spec.ts` fails on a second one — so
budgets, bills and backups ask the notification layer and never a transport
(INV-PUSH-001…005). A push endpoint is a URL the server will POST to, so it is
validated with `IsPushEndpoint` (https floor plus the AI provider's SSRF check,
with the DNS lookup bounded inside the check), and the unique index on
`endpoint_hash` means one endpoint has one owner: a second account subscribing
in the same browser is **refused**, never allowed to take the row over.

**2. `budget_alerts` → `notifications`, with one writer.** The table stopped
being about budgets some time ago (`BACKUP_FAILED`, `SMTP_FAILURE`,
`PROVIDER_OUTAGE` already lived in it). There were three writers with three
opinions about the row's shape; `NotificationService.create` is now the only
one, and `notification-write-door.spec.ts` fails on a second. `category` is
derived (`notificationCategoryOf`), never stored. Migrations 037/112/140/170
that name the old table are made rename-tolerant with a `to_regclass` guard, so
a full replay on a fresh `schema.sql` no longer aborts on a table migration 179
renamed; behaviour on a database that has not reached 179 is unchanged.

**3. Deep links.** Clicking a notification opens the subject rather than the
dashboard. Every target a producer writes is checked against the router tree by
`notification-target.contract.test.ts`.

**4. Reminders.** `notification_reminders` carries the template a fire re-emits
(5 minutes to 7 days, 50 active per account). The firing cron claims due rows
with one conditional `UPDATE ... RETURNING` through a `FOR UPDATE SKIP LOCKED`
CTE, so each row fires exactly once across replicas; a one-shot is stopped in
the same transaction that writes its follow-up, and each fire dismisses the
previous nag so a month of unopened repeats is not a month of unread rows.
Stopping works from the bell, from a **Stop** action on the push, and from
`POST /notifications/reminders/:id/stop`.

**5. The preference matrix and the dispatch seam.** `notification_preferences`
is one row per `(user, category)` with `email` (report), `email_notification`
(alert), `push`, `unifiedpush` and `throttle_minutes`.
`NotificationDispatchService.notify` is the seam every producer calls: it writes
through the write door, then fans out per the matrix. The cooldown gates the
**fan-out only** — never the in-app row, never a report email, never a strict
severity escalation — and when it is active the `(user, category)` advisory lock
is taken **before** the row is written and the decision is made on the same
connection, because a lock taken after the commit serialises nothing. A
reminder's re-emit sits outside the cooldown on both sides: the interval is the
user's own choice, and the cooldown governs producers.

**6. `SYSTEM` in the matrix.** Admin alerts fan out through the same seam, so
`SYSTEM` push is a live per-admin control. It exposes push only (its email is
the severity-driven, SQL-gated admin fan-out, not a user toggle) and its row is
listed and writable **for administrators only** — every `SYSTEM` type is raised
through `raiseAdminAlert`, so for anyone else the row would be a control over
nothing.

**7. UnifiedPush.** A UnifiedPush subscription *is* a Web Push subscription
whose endpoint is a distributor, tagged `transport = 'unifiedpush'` and gated by
its own matrix channel, delivered by the same sender — never a plaintext ntfy
publish, because a notification body can carry an amount or a payee and a
finance app must not hand that to a third-party relay in the clear.

### Privacy of the wire

The payload's `title`/`body` are the **category's** generic copy in the
recipient's stored locale (`push.notification.<category>`), not the row's own
title and message: the wire is encrypted end to end (RFC 8291) but the lock
screen is not, so the screen shows the category and nothing of the money. The
detail is one tap away through `target`. `collapseKey` carries an id or a cause
key, never a name or an amount.

### Database

Five new tables (`notifications` renamed from `budget_alerts`,
`notification_preferences`, `notification_reminders`, `push_subscriptions`,
`push_instance_config`) across migrations 178–184, all idempotent, `schema.sql`
in sync. Every user-owned table ships its RLS policy in the same migration
(three new policies, so the enumeration in `schema.sql` moves from 58 to 61);
new exported columns are classified in the support backup, and
`notification_reminders.source_notification_id` is covered by the
support-backup FK rules. `restore-plan.ts` gains the three tables in dependency
order. "Delete my data" removes push devices and reminders and deliberately
keeps `notification_preferences`, which is a setting rather than data about the
user's money.

### Verification

| Gate | Result |
|---|---|
| Backend `test:unit` | 564 suites, 14 929 tests, 2 skipped |
| Frontend `vitest` | 780 files, 15 160 tests |
| `tsc --noEmit`, both layers | clean |
| ESLint | clean |
| `migration:lint` | every body re-runnable |
| i18n parity, duplicate keys, pseudo-locale | clean, 22 locales |

78 test files touched, 48 of them new. Beyond the per-service specs, the
mechanical rules are held by scanning guards: the write door, the push secret
boundary, the transport list against the CHECK constraint, notification targets
against the router tree, and the occurrence/category coverage of every producer.

**Not verified on the authoring machine** (no local PostgreSQL, no browser):
schema-vs-migrations drift, the support-backup golden test, RLS enforcement, the
real-database dedupe race, E2E and Lighthouse — all CI-owned. Push delivery
itself needs real browsers; the manual pass is scripted separately and covers
Chrome/Edge, Brave (which blocks Google's push service until one privacy switch
is on — the failure now names that switch), Firefox and Safari (no Cookie Store
in the worker, so the Stop action degrades to opening the app), installed iOS
web apps, and Android's OS-level notification block.

### Deliberate decisions, worth a reviewer's eye

1. A reminder's re-emit is exempt from the category cooldown in both
   directions (spec §13.3).
2. Push copy is the category's, not the row's (spec §14) — see above.
3. A UnifiedPush distributor on a private network is refused by the SSRF check
   (spec §15.1); an operator allowlist is future work.
4. `PUT /notifications/preferences/:category` carries no `@DemoRestricted`,
   matching `PATCH /users/preferences`; the daily demo reset covers it.
5. `notification_preferences` survives "delete my data".

### Not in this PR

Per-type preferences inside a group (the matrix has three groups today, and the
preference key is `(user_id, category)`, so that is a schema change) and the
producers for security prices, portfolio value and the GEM report. Both need a
spec first, per the repository's rule for financial features. Also deferred: the
chart in a push payload (`payload.image`, spec §14), a standalone reminders
page, and E2E coverage of the push flow.

## Checklist

- [ ] This PR addresses a **single concern** (large work is split into a series).
      **Not claimed.** This is the seven-phase series from #1291 delivered on one
      branch (62 commits, 381 files). Each phase is a self-contained commit range
      and can be split into its own PR on request — phases 1–2 are the natural
      first cut, since 3–7 build on the renamed table and the write door. Say the
      word and I will split it rather than have you review it as one change.
      One shared-area change: the `budget_alerts` → `notifications` rename
      (agreed in #1291, phase 2a) and the rename-tolerance guards it forced on
      migrations 037/112/140/170.

## AI assistance disclosure

Written with Claude Code. Every phase went through the repository's `/audit`
protocol and five `/code-review` passes, with the findings fixed on the branch;
the round-by-round commits are in the history. The maintainer decisions that
changed the design — the cooldown gating the fan-out rather than the in-app row,
`SCHEDULED_POST_FAILED` as `PAYMENTS`, `SYSTEM` as push-only, encrypted
UnifiedPush rather than a plaintext publish — were put to the maintainer and
taken from their answers, not decided by the tool. I have reviewed the result and
own it; the CI-only checks listed above are the parts not verified on my machine.

---

https://claude.ai/code/session_01VFPRYnmybfdK9PAUqdyp3a


WMP:
### Open inside this PR's scope

Things this PR delivers but knowingly leaves partial. None of them is a
surprise waiting for a reviewer to find — each is named in the code, the spec or
the invariant register, and each is a candidate for a follow-up on this feature
rather than a new one.

1. **`INV-NOTIFY-001` is `partial`, not `enforced`.** The write-door scan cannot
   see the restore's dynamic `INSERT INTO "${table}"`, so a restored
   notification row never passes `boundedTitle` / `boundedDedupeKey` /
   `boundedTarget`. Mitigated rather than closed: `target` is re-validated at
   the consumer (`safeNotificationTarget`, in the app and again in the service
   worker), and the column widths are the database's own — an over-long value
   raises 22001 and fails the restore loudly. Closing it means bounding at the
   restore or accepting the exception explicitly.
2. **A row's stored copy is producer-composed English.** The push body is the
   category's generic copy in the recipient's locale and the immediate email's
   frame is localized, but `title`/`message` are not: the bell composes
   localized copy from `data`, so the gap is visible in the immediate email
   body. Named as a known gap in spec §14.
3. **`notification_browser` is now dormant.** The matrix's `push` column
   supersedes it, yet the legacy preference is still persisted, still writable
   through `PATCH /users/preferences`, still in the frontend auth types — and
   gates nothing. Removing it is a migration plus DTO and type changes, which I
   kept out of a diff this size; leaving it is a preference that lies.
4. **Phase 5 ships without the chart the spec anticipates.** `payload.image`
   (spec §14, with a one-shot-token chart endpoint) has no producer and no
   endpoint here; the payload carries no image.
5. **UnifiedPush has no registering client in this repository.** The transport,
   the per-category gate, the device badge and the admin row all ship, but a
   browser cannot subscribe to an arbitrary endpoint — only a UnifiedPush-capable
   client can create such a row. The channel therefore cannot be exercised end
   to end from this tree; what is testable is the gating, the badge and the
   transport-restricted fan-out.
6. **A distributor on a private network is refused.** The SSRF check rejects
   `https://ntfy.home.lan` at registration and before every send. An operator
   allowlist scoped to `transport = 'unifiedpush'` is the way out and is not
   here (spec §15.1).
7. **The push Stop action degrades on Firefox and Safari.** Those workers expose
   no Cookie Store, so the worker cannot read the CSRF token, the server refuses,
   and the worker opens the app instead of silently failing. A worker-friendly
   CSRF path would close it; the fallback is deliberate, not accidental.
8. **There is no standalone reminders page**, which is why the Stop fallback
   lands on the notification's own target rather than on a list of reminders.
9. **One FK is covered by rule rather than by the harness.**
   `notification_reminders.source_notification_id` is a plain column on the
   entity, so the entity-synced test database has no constraint for the
   support-backup coverage guard to find; it is listed in `SCHEMA_ONLY_FKS` with
   an explicit `REFS` rule instead. Modelling the relation would let the guard
   see it for real.
10. **Delegates can read but not manage.** `GET /notifications` carries
    `@AllowDelegate`; devices, preferences and reminders do not. Fail-closed by
    design — but nobody has yet decided what a delegate *should* be able to
    manage.
11. **The reminder cron is bounded per tick.** At most `CLAIM_BATCH = 100` due
    rows are claimed per minute and re-emitted `REEMIT_CONCURRENCY = 5` at a
    time; a larger backlog waits for the next tick. Fine at today's scale, and a
    knob to revisit rather than a hidden ceiling.
12. **No Playwright coverage of the push flow.** Delivery needs real browsers, so
    the manual pass carries it; parts of it are automatable for Chromium.

kenlasko:
Been playing with this on Android. It's working well so far. When Monize showed the banner, I clicked it and Android asked me to grant the permission. I didn't need to do it manually.

When you look at the notification, the Monize icon shows properly. But the icon on the top-bar is just a white box. Wonder what it takes to make it use the Monize logo.

I

PR #1308, opened 2026-09-03
Serve MCP revision 2026-07-28 beside the 2025-era revisions, on the v2 SDK

kenlasko:
`@modelcontextprotocol/sdk` 1.x is frozen at 1.30 and can never speak the
2026-07-28 protocol revision. This moves the server onto the v2 packages
(`@modelcontextprotocol/server` + `/node`, with `/client` for tests) and then
serves the new revision beside the 2025-era ones from a single server
definition, so no client in the field moves on our schedule.

`/api/v1/mcp` routes on the SDK's own `isLegacyRequest`: a 2026-07-28 request
goes to `createMcpHandler(..., { legacy: "reject" })`, everything else keeps the
sessionful transport. The SDK's stateless legacy fallback is deliberately
unused -- it would serve a 2025-era client from an instance that holds nothing
between rounds, and our 2025 confirmation is a server-side wait.
`docs/adr/0004` records the decision, the alternatives and the one thing the
upgrade makes worse.

## Identity per request

Identity stops coming from the session table. The transport validates the
bearer exactly as before and attaches the result as the SDK's `AuthInfo`; a
handler reads the caller from the request through `resolveUserContext(ctx)`,
which is INV-MCP-001 restated as a property of the request rather than of a
session. A session is still bound to the credential that opened it, and its
scopes are still re-read per request. An OAuth grant with no id is now refused
outright rather than allowed to open one session it could never re-present.

"Which session is this" becomes "which caller is this" (`callerKey`): the
session id on a 2025-era connection, the credential id where there is no
session. Only the relay claim and the observed-elicitation memory ask that
question, and neither changes behaviour on a 2025-era connection. The cost is
that two clients sharing one token share a caller key on the new revision --
recorded in the ADR, and the connect instructions now tell a relay user to mint
the agent its own token.

## A write confirmation that travels instead of waiting

2026-07-28 has no server-initiated requests, so a write is confirmed over two
`tools/call` POSTs: round one returns the question carrying a signed
`requestState`, round two re-derives the same items and reads the answers. The
seal is HMAC-SHA256 over a fingerprint and nothing else, bound to the
credential and the method, with a ten-minute TTL, verified by the SDK seam
before any handler runs. `confirmWrite` moves to its own file and keeps its
outcome-based logic verbatim; the 2025-era path still asks and waits, through
`ctx.mcpReq.elicitInput` so the dialog stays on the tool call's own stream.
`clientAnsweredForItself` now reads both v2 error families, and the pin that
keeps the synthesized `elicitation.form` capability from becoming load-bearing
again stays. INV-MCP-003 and INV-MCP-004 are the new invariants.

**What is fingerprinted is the change, not the round it was built in.** The two
rounds are two tool calls, so each rebuilds its descriptor with a fresh
`actionId`, `expiresAt` and attachment parking slot; `roundStableAction` drops
exactly those at every depth and everything that says what is being approved
still goes in. The envelope half comes from `AI_ACTION_ENVELOPE_FIELDS`,
declared beside the descriptor and typed into `newEnvelope`, so the compiler
refuses a new per-build field that is not on the list. `mcp-confirm.spec.ts`
drives the real action builder twice to prove it -- a builder double that
returns one frozen object agrees with itself whatever is hashed.

The seal bounds who and what, not how many times: the same seal and answer
replayed by the same credential against the same tool inside the TTL commits
again. That is deliberate and written down -- the answer is client-asserted on
this revision, as it already is on the 2025-era one.

## Also in this change

- `server/discover`, `ttlMs`/`cacheScope` cache hints on every cacheable
  listing, the `Mcp-Method`/`Mcp-Name` routing headers through CORS, and RFC
  9207 `iss` on the authorization response. Client ID Metadata Documents are
  deferred (`docs/future-plans/mcp-client-id-metadata-documents.md`).
- The `logging` capability is retired: `logging/setLevel` is gone in this
  revision, the feature is deprecated (SEP-2577), and nothing here ever sent a
  log notification.
- Individual approval mode asks for every card in one round rather than a round
  per card, and a relay card is offered only on the round that asks.
- Mechanical, everywhere under `src/mcp/`: SDK import paths, `(args, extra)` ->
  `(args, ctx)`, raw Zod shapes wrapped in `z.object(...)`, and
  `register(server)` without the resolver. `createServer()` takes no era -- a
  factory that cannot see which revision asked cannot answer them differently.
- Specs read the caller from a built `ServerContext`
  (`testing/mcp-test-context.ts`) instead of mocking a resolver, which is the
  same change the handlers made. `mcp-eras.spec.ts` drives both revisions in
  process against the real SDK.

## `tools/list` shrank ~9%

50,337 bytes, was under a 55,500 cap, because v2's Standard JSON Schema
emission is more compact than the 1.x converter. No tool definition changed, so
every cap comes down with the measurement rather than banking the slack:

    list_accounts                  2300 -> 2050    lookup_securities   1750 -> 1550
    list_transactions              4050 -> 3550    manage_securities   4750 -> 4200
    compare_periods                2150 -> 1900    manage_investment_* 5050 -> 4500
    manage_transactions            6700 -> 5950    list_upcoming_bills 3400 -> 3000
    list_categories                1350 -> 1200    calculate           1400 -> 1200
    list_payees                    2450 -> 2150    get_budget_status   2900 -> 2550
    manage_payees                  3450 -> 3000    get_next_prompt     1600 -> 1400
    generate_report                3150 -> 2800    post_response       1200 -> 1050
    get_portfolio_summary          3500 -> 3100    report_progress     1400 -> 1250
    list_investment_transactions   2750 -> 2450    TOTAL              55500 -> 50500
    list_capital_gains             2300 -> 2050

https://claude.ai/code/session_01DWaDZPVzX8GpRw3HCWSYR1


kenlasko:
**NPM Audit is red, and it is not this PR's failure.** Recording it rather than leaving it unexplained, since this PR is exactly the one that adds dependencies.

The job did not find an advisory. It never got an answer from npm:

```
npm warn audit network timeout at: https://registry.npmjs.org/-/npm/v1/security/advisories/bulk
undefined
npm error audit endpoint returned an error
```

Five minutes between the last install line (23:15:38) and the error (23:20:38), against `registry.npmjs.org` — a service the diff does not touch. The same run's `npm ci` also took five minutes, so that runner had degraded registry connectivity throughout.

**Verified rather than assumed.** I ran the job's exact command against this PR's exact lockfiles:

| | `npm audit --audit-level=high --omit=dev` |
|---|---|
| `backend/` | `found 0 vulnerabilities` |
| `frontend/` | `found 0 vulnerabilities` |

So the check has nothing to report on the packages this PR adds (`@modelcontextprotocol/{server,node,client}`, and `hono` / `@hono/node-server` transitively). No fix exists to port, because there is nothing to fix.

**Re-run: pending, not skipped.** `rerun-failed-jobs` is refused with `403 This workflow is already running` while the four E2E shards are still going. I will re-run it once the workflow finishes — that is the one re-run this failure gets, and a second failure would be treated as real.

Everything else on this head is green: 23 of 28 checks, including Backend Unit, Backend Integration, Frontend Unit, both lint/typecheck jobs, Lighthouse, CodeQL and both License Compliance jobs. The four E2E shards are in flight.

---

kenlasko:
### NPM Audit: an npmjs outage, not this branch

Closing out the re-run I said was pending. It is spent, it failed again, and
the second failure names the cause outright.

**What the re-run's own log shows** ([job](https://github.com/kenlasko/monize/actions/runs/33816421622/job/100859345128)):

```
found 0 vulnerabilities                    <- backend, --audit-level=high --omit=dev
added 717 packages, and audited 718 ...    <- frontend npm ci, fine
npm warn audit 503 Service Unavailable - POST
  https://registry.npmjs.org/-/npm/v1/security/advisories/bulk - Service Unavailable
{ error: 'Service Unavailable' }
npm error audit endpoint returned an error
```

The backend audit answered `0 vulnerabilities` **in that run, on this PR's own
merge commit**. The frontend `npm ci` resolved and downloaded 717 packages, so
the registry was reachable. Only the advisories endpoint failed, and it failed
with a 503 rather than an advisory. The first attempt was the same endpoint
timing out; two symptoms, one endpoint, an hour apart.

Reproduced off GitHub's network entirely -- from an unrelated container at
00:21Z, `npm audit --audit-level=high --omit=dev` in `frontend/`:

```
npm warn audit 503 Service Unavailable - POST
  https://registry.npmjs.org/-/npm/v1/security/audits/quick - Service Unavailable
npm error audit endpoint returned an error
```

Different machine, different network path, same 503. That rules out the runner
and rules out the diff.

**Why it is not this PR's.** This branch changes `backend/package.json` and its
lockfile; `frontend/package.json` is untouched, and the frontend lockfile drift
that had crept in is reverted in f576110. The failing half is the frontend
audit, whose dependency set is byte-identical to `main`.

**No fix exists to port**, so here is the proposed patch instead, unpushed --
it is a CI workflow change with nothing to do with the MCP upgrade, and folding
it in would widen this PR.

The root cause is that `npm audit` exits 1 for a real advisory **and** for an
unreachable endpoint, so the gate cannot tell "you shipped a vulnerability"
from "npmjs is down". `--json` can: an endpoint failure carries a top-level
`error`, a finding carries `metadata.vulnerabilities`. Retry only the first,
and never let a retry swallow the second:

```yaml
      - name: Audit production dependencies (blocking)
        run: |
          # `npm audit` exits 1 for a finding AND for an unreachable advisories
          # endpoint, so the exit code alone cannot separate a vulnerability we
          # shipped from an npmjs outage. `--json` can, and only the outage is
          # worth retrying -- a finding must still fail on the first attempt.
          audit_prod() {
            local attempt out
            for attempt in 1 2 3; do
              if out=$(npm audit --audit-level=high --omit=dev --json); then
                echo "$out" | jq -r '.metadata.vulnerabilities
                  | to_entries[] | "\(.key): \(.value)"'
                return 0
              fi
              if ! echo "$out" | jq -e 'has("error")' >/dev/null 2>&1; then
                echo "$out"          # real advisories: fail now, do not retry
                return 1
              fi
              echo "::warning::npm advisories endpoint unavailable (${attempt}/3)"
              sleep $((attempt * 30))
            done
            echo "::error::npm advisories endpoint did not answer after 3 attempts"
            return 1
          }
          (cd backend  && npm ci && audit_prod)
          (cd frontend && npm ci && audit_prod)
```

It still fails the build on a high-severity advisory in a runtime dependency,
on the first attempt, which is the property the gate exists for. It stops a
registry outage reading as one. Deliberately not `continue-on-error` -- that
would make an outage silently indistinguishable from a clean audit, which is
the opposite trade from the informational dev-dependency step below it.

Happy to open that as its own PR if you want it.

**Where this leaves #1308.** Every other check is green, including all four E2E
shards. f576110 is pushed and a fresh run is under way, which re-runs the audit
against the registry; if npmjs has recovered it should go green on its own. I
am still watching and will report either way.

---

Issue #1193, opened 2026-08-17
Detail density selection is constantly forgotten


bcorob:
On an investment account detail page, click "Normal" to change the density to "Compact" or "Dense"
Observe the transaction rows get smaller and more information fits on the screen.
Refresh the page, or navigate to a different page and navigate back.
Observe the transaction row density.

Actual results:
Transactions constantly switch back to normal density

Expected results:
For a given view, the user's selected density should be sticky and persisted in this browser indefinitely.

Arguably information density should be a profile customization that applies to all views, rather than a per-view customization that needs to be selected at each type of interaction.

PR #1332, opened 2026-09-08
v1.16.0 release notes, a notifications tour, a foldable Browser push block, and favourites-first account pickers

kenlasko:
## Linked discussion / issue

Closes #

Approved in: **no discussion or issue** — each of the five changes below was requested directly by @kenlasko in a working session, in the order they appear. Flagging it rather than ticking the box: the propose-first workflow was not followed here, and if any of these want a discussion first, say which and I will pull it out of the branch.

## Summary

Five changes. They are related by subject (1.16.0's notification and account surfaces) rather than by being one concern — see the checklist.

**1. `docs/release-notes/1.16.0.md`** — the release notes for the 37 PRs merged since the v1.15.1 version bump, indexed in `docs/release-notes/README.md`. Drafted from the specs, plan docs, English catalogs and `.env.example` rather than commit subjects, so the claims are about the tree.

**2. A guided tour of notifications** (`release-1.16.0/notifications`) — nine steps in two halves: the bell and its panel, then the three Settings blocks that decide where a notification goes. Every step advances with Next, so a reader who presses nothing reaches the end, and every anchor is a container that renders in each state of its block — the panel and its filter strip render over an empty inbox, the push block renders its "an administrator has not enabled this" message rather than disappearing — so there is no data gate and no dropped step for a first-run account.

One engine addition: `openNotificationBell` on `TourStep`, the sibling of the introduction tour's `openToolsMenu`. The notification panel closes on a click outside itself, so a `click` advance would open it and the reader's next press — on the tour card — would close it under the step describing its contents. `NotificationBell` ORs the flag over its own state, so it also wins over the click-outside close.

**3. The Browser push block folds away** (`<details>`, matching `PushDiagnostics` in the same card). Collapsed, its heading carries how many devices can be **delivered to**; retired rows are named separately rather than summed in, because counting them would make the summary disagree with the list underneath and including them would claim delivery to endpoints that answer nothing. A device read that failed says "Device list unavailable" rather than reporting the empty array as "No devices registered". The three branches whose whole content is one sentence explaining why push is unavailable do not fold.

**4. The fold is remembered** in `localStorage` under `monize-settings-sections`, via a new `settingsSectionStore` — one store for every foldable Settings section rather than one per section, which is `densityStore`'s lesson (issue #1193) applied before it can be relearned. Browser-local for the same reason row density is: whether a panel is worth its height is a fact about the screen, not the account.

**5. Favourites first in both account switchers**, and the lists properly ordered. The ordering rule `buildAccountDropdownOptions` already had is now `orderAccountsForPicker` / `orderForPicker`, shared by both switchers, the Transactions filter's favourite chips and the CSV transfer-rule picker (which held a hand-copied version of the whole rule). The account detail page's switcher previously had no sorting at all.

### Three decisions worth a reviewer's eye

- **`favourite_sort_order` defaults to 0**, so a user who has starred accounts without ever dragging them into an order has ties, and a stable sort leaves those in whatever order the API answered in — the same accounts reading differently on two screens. The helper breaks ties alphabetically; the drag-to-arrange list writes real indices, so the tiebreak never fires once anyone uses it.
- **The name a picker sorts on is the one it displays.** The detail switcher folds a linked brokerage/cash pair into one row under a stripped `displayName`, so ordering by the stored name reads as unsorted. A pair's star is the *primary's*, because that is the row the accounts list draws and stars.
- **The dashboard's `FavouriteAccounts` is deliberately left alone.** It is the drag-to-arrange surface itself, its order is the thing being edited, and its two sorts must agree with each other for a drag to move the right card.

### Two traps found on the way, both written up in `frontend/CLAUDE.md`

- **jsdom flips `<details>.open` on a summary click and fires no `toggle` event.** So `onToggle` alone would be untestable here and one missed event away from a summary describing the wrong state. `open` is controlled, the summary's `onClick` moves it, and `onToggle` stays wired for the toggles no click produces (Chrome expanding a `<details>` to reveal a find-in-page match).
- **A persisted fold outlives a test.** `setup.ts` clears `localStorage` but cannot reach a store that has already read it, so the first test to collapse a section left it collapsed for the rest of the file. And the fold is asserted on `details.open`, not on the content having gone — jsdom applies no user-agent stylesheet, so "the button is not there" would pass in a browser and fail here for a reason unrelated to the component.

## Checklist

- [ ] An approved discussion or issue exists and is linked above. — **No.** Requested directly in a session; see above.
- [ ] This PR addresses a **single concern** (large work is split into a series). — **No.** Five concerns, each its own commit, in the order they were asked for. Happy to split into a series if you would rather review them separately; the commits are already clean cut points.
- [x] New behavior has tests, and the existing suite passes. — 31 new tests. `npm run lint`, `npx tsc --noEmit`, `npm run i18n:check` and the full Vitest suite are green: **842 files, 16,234 tests**.
- [x] All user-facing strings are translated for **every** locale (i18n parity). — Tour copy, the collapsed-summary strings and the switcher section labels are in all 18 non-English catalogs; pseudo-locale regenerated. The three locales that already said "Favourites" two different ways were aligned with the wording the `accounts` namespace uses, since that is the star this heading reflects.
- [x] No shared/core areas were refactored without prior agreement. — The two shared touches are additive: a new optional `TourStep` field, and `orderForPicker` generalising a rule `account-utils` already owned. No existing call site changes behaviour except the four account pickers, deliberately.
- [x] The branch is rebased on the latest `main` (`f14d0287`, nothing behind).

## AI assistance disclosure

Written by Claude Code (Opus 5) across one session, driven turn by turn by @kenlasko. Every claim in the release notes was checked against the specs, catalogs and source rather than commit messages; the design decisions above were made in the session and are recorded in the commit messages and `frontend/CLAUDE.md` rather than only here.

https://claude.ai/code/session_01Yb1GH1uL461X5EdDTrpWAz

---
_Generated by [Claude Code](https://claude.ai/code/session_01Yb1GH1uL461X5EdDTrpWAz)_


WMP:
Please wait for notifications pr and mobile table part 2

kenlasko:
@WMP, are you all done now?

WMP:
No, we still need to carry out manual testing of the notifications and create some more mobile tables. I’m trying to do them using Codex, but it’s taking a while… Claude won’t be back online until Thursday.

PR #1334, opened 2026-09-09
Add a "Send to AI Assistant" destination to the share review screen

kenlasko:
## Linked discussion / issue

No discussion or issue exists for this one. It was requested directly by the repository owner in a Claude Code session ("Add an option on the web share target modal for 'Send to AI Assistant'. This should only appear if AI is configured in Monize"). The work was already anticipated by `docs/future-plans/pwa-web-share-target.md`, which listed "Ask the assistant" as Phase 2 of the shipped share-target plan.

Closes #
Approved in: not applicable, see above

## Summary

The Web Share Target's review screen offered two destinations, so a shared receipt or CSV the user wanted to *ask about* had to be saved first and re-attached by hand. It now offers the assistant as a third, reusing the hand-off the import wizard already has: `/share` routes to `/ai?share=<id>`, the page reads the bundle as the signed-in reader, and `ChatInterface` stages the files on its composer. Nothing is sent. The user still presses send, so INV-SHARE-002 ("a share is reviewed, never applied") holds for the new destination too.

The row appears only when both halves of "this can work" are true:

- **A provider that can answer** — `useAiConfigured()`, the hook the floating chat bubble already gates on. Absent means the row is not rendered, never a button whose one outcome is "configure a provider first".
- **A chat that accepts these exact files** — a new `assistantAcceptsFiles` in `lib/ai-attachments.ts`, which runs the chat's own `validateFile` and `validateAddition` rather than restating its caps. The caps genuinely differ: the share stash holds 10 files at 10 MB each, the assistant takes 5 at 5 MB and cannot read OFX, QFX or QIF. So a share the wizard imports happily is often one the chat would refuse file by file. It is all-or-nothing, because staging only the readable subset would send the assistant part of what the user shared and say nothing about the rest.

Two mechanics worth calling out in review:

- The stash is discarded only once `ChatInterface` reports the bytes staged (`onInitialFilesStaged`), never on the way in — dropping it while the files were still being read would leave the user with neither the share nor the attachments. The callback fires even when the files are refused, so the caller is never left holding a bundle nothing will come back for.
- The staging pass is guarded by a ref rather than by its dependency list: `addFiles` changes identity with the attachment list, so an effect keyed on it alone re-stages the share on the render its own first file caused. Both `ChatInterface` cases fail without that guard (verified by removing it).

## Checklist

- [ ] An approved discussion or issue exists and is linked above. — **No.** Requested directly by the repository owner; the destination was pre-agreed as Phase 2 in the committed share-target plan.

### Tests

- `src/app/share/page.test.tsx` — the row is withheld with no provider, for a QIF statement and for a six-file share; offered beside the wizard for a CSV; and pressing it routes to `/ai?share=<id>` **without** discarding the bundle.
- `src/app/ai/page.test.tsx` — the hand-off stages the files, discards only on the staged callback, reads nothing until the reader is known, and leaves the chat usable for an expired bundle or a failed stash read.
- `src/components/ai/ChatInterface.test.tsx` — handed-over files are staged with nothing sent, staged exactly once across re-renders, and reported as staged even when refused.
- `src/lib/ai-attachments.test.ts` — `assistantAcceptsFiles` across type, per-file, count, total-size and empty-set refusals.
- `e2e/tests/share-target.spec.ts` — the assistant is absent on an account with no provider.

Full frontend suite green (849 files, 16,393 tests), plus lint, `type-check`, i18n parity and pseudo-locale freshness, and the backend doc-path guards.

### Docs

INV-SHARE-002 now names three destinations and their gating, the plan's destination table and phasing record what shipped, and `frontend/CLAUDE.md` carries the rule that a destination asks the destination's own validators rather than restating its caps.

## AI assistance disclosure

Written by Claude Code (Claude Code on the web), driven by the repository owner. Every file, test and translation in this branch was produced in that session and reviewed before pushing.

https://claude.ai/code/session_01An5GeoHsZSTiNqjzktpv9N

---
_Generated by [Claude Code](https://claude.ai/code/session_01An5GeoHsZSTiNqjzktpv9N)_


kenlasko:
**NPM Audit is red on this head, and the failure is not this PR's.** Standing down on it, with the evidence and a proposed patch below.

**What fails.** The `NPM Audit` job, in both workspaces, on advisories only:

```
backend:  high-severity advisories in production dependencies (moderate: 1, high: 7, total: 8)
frontend: high-severity advisories in production dependencies (high: 1, critical: 1, total: 2)
```

**Why it is not this PR's.**

- The diff is 33 files of frontend source, tests, i18n catalogs, docs and one E2E spec. It touches no `package.json` and no `package-lock.json` (`git diff origin/main...HEAD -- '**/package.json' '**/package-lock.json'` is empty).
- I reproduced the failure locally on this branch, whose lockfiles are byte-identical to `main`'s: `npm audit --omit=dev` reports the same two frontend advisories and the same eight backend ones. The input is the lockfile `main` already carries.
- `main`'s last CI run passed at 19:33 UTC on 8 Sep (run 34269609633). These advisories were published after it, so `main` would fail the same check today. This job reads a live advisory database, which is why nothing in the repository had to change for it to turn red.

I did not spend a re-run on it: the check is deterministic against that database, so it would fail identically, and the local reproduction on the unchanged lockfile is the stronger evidence.

**No fix exists to port.** There is no open Dependabot PR and no other open PR in the repository carrying these bumps.

**Proposed patch, as a PR of its own.** Both workspaces report every fix as in-range, so it is `npm audit fix` (no `--force`, no breaking changes) in `backend/` and `frontend/`, touching only the two lockfiles:

| Workspace | Package | Severity | Fixed in |
| --- | --- | --- | --- |
| frontend | `next` | critical | 16.3.3 (RCE on Windows hosts, GHSA-p293-qw3h-jr36; AVIF image-optimization RCE, GHSA-2xp9-vwfh-vxw4) |
| frontend | `sharp` | high | 0.35.4 (libheif, GHSA-rgj7-g3m4-5g8c) |
| backend | `nodemailer` | high | >9.1.0 (four advisories, including two recipient-domain allow-list bypasses) |
| backend | `multer` (via `@nestjs/platform-express`) | high | >2.2.0 (four denial-of-service advisories) |
| backend | `hono` (transitive) | moderate | >4.13.4 |
| backend | `@nestjs/core`, `@nestjs/platform-express`, `@nestjs/schedule`, `@nestjs/swagger`, `@nestjs/typeorm` | high | pulled along by the `multer` fix |

I have deliberately not folded that into this PR. A NestJS platform and mailer bump is a runtime-behaviour change with its own test obligation, and CONTRIBUTING asks each PR to address a single concern. Happy to open that dependency PR separately on request.

Every other check on this head is being watched, and I will push fixes for anything red that does belong to this change.

---

kenlasko:
Pushed the audit fix into this PR after all, at the owner's request (460fdb2). Lockfiles only; neither `package.json` moves, so every bump is inside an existing range.

| Workspace | Package | Move |
| --- | --- | --- |
| backend | `nodemailer` | 9.0.1 → 9.1.1 |
| backend | `multer` | 2.2.0 → 2.3.0 |
| backend | `hono` | 4.12.34 → 4.13.7 |
| backend | `@nestjs/common`, `/core`, `/platform-express`, `/testing` | 11.1.x → 11.2.3 |
| backend | `@nestjs/schedule`, `/swagger`, `/typeorm` | patch |
| frontend | `next` | 16.3.0 → 16.3.4 |
| frontend | `sharp` | 0.35.3 → 0.35.4 |

**`npm audit fix` on its own was not the fix, and it fails loudly.** It bumps only the packages an advisory names, so `@nestjs/core`, `/platform-express` and `/testing` went to 11.2.3 while `@nestjs/common` stayed at 11.1.17 — and 11.2.3's router imports a decorator that version does not export. Every suite importing `@nestjs/testing` then fails to run: 360 of 606 backend suites red. `@nestjs/common` is bumped with them. When an advisory moves one member of a package family, move the family. (The frontend half is `npm update next sharp` rather than `audit fix`, which crashes on that tree with npm's own `Cannot read properties of null (reading 'edgesOut')`; it reaches the same versions inside the existing carets.)

Validated before pushing:

- Backend: lint, type-check, full unit suite — 606 files, 15,784 tests, the same run that was 360 suites red before the `common` bump.
- Frontend: type-check, lint, a production build on Next 16.3.4, full suite — 849 files, 16,393 tests.
- `npm audit --omit=dev` now reports zero vulnerabilities in both workspaces.

The backend integration suite needs PostgreSQL and this environment has no Docker daemon, so CI is its first run on these versions; I am watching it. The remaining dev-only advisories (vitest's mocker, `js-yaml`) are deliberately untouched — the job gates production dependencies, and bumping the test runner is not this PR's business.

---

PR #1337, opened 2026-09-09
fix(share): remove the redundant share inbox banner

kenlasko:
## Linked discussion / issue

Closes #
Approved in: no discussion or issue — reported and scoped directly by @kenlasko (repo owner). The Web Share Target itself was agreed in discussion #1292 and is unchanged by this PR.

## Summary

Sharing a file into Monize and pressing **Send to AI Assistant** works — and then the share inbox banner, mounted in the shell above the route that had just consumed the stash, announces "1 file was shared with Monize" over the assistant. Its **Review** link leads to **"Nothing here to review"**, because the share it offers a way back to is the one the user has already finished routing.

The banner was built for the case where the worker's redirect does not survive: an OIDC round trip that drops `returnTo`, or opening Monize from the launcher instead of following the redirect. That case is real, but it is rarer than the just-consumed case it collides with, and its cost is paid on the common path. Rather than teach it to tell the two apart — more state about a stash the user is already done with — the banner is removed, and a share whose redirect does not survive is shared again.

**Unchanged:** the OS share sheet, the `/share` review screen, the stash, and the hand-offs to the transaction form, the import wizard and the assistant.

### The sweep outlives the banner

`purgeExpiredSharedBundles()` ran inside the banner and does **not** go with it. It is one of the three mechanisms behind INV-SHARE-003, and the worker's own sweeps (`activate`, and before each new share) leave a gap: a returning visit that neither updates the worker nor shares anything new would never age a stash out, so shared files would sit on the device past the lifetime the review screen promises.

`ShareStashSweeper` is that call and nothing else — a null-rendering component that keeps the `/share` exclusion, since an expired bundle stays readable there so the screen can say "expired" rather than "nothing here". It deliberately does **not** carry the banner's auth gate: the banner waited for a signed-in reader because listing also *claims* ownership, whereas a sweep decides on `createdAt` alone.

Happy to drop the component and let the worker cover expiry instead, if you would rather have the plain deletion.

### Also

- The three now-unused catalog keys (`noticeTitle`, `noticeAction`, `noticeDismiss`) removed from all 20 locales, pseudo-locale regenerated.
- INV-SHARE-003 and INV-SHARE-005 enforcement text, `frontend/CLAUDE.md`, and the plan document updated — the last one states it "keeps describing the tree rather than the intention", so section 10 records why the banner went and why the sweep stayed.

28 files changed, 217 insertions, 405 deletions.

**Verification** (fresh dependency install):

```
tsc --noEmit          clean
eslint .              1 warning, pre-existing on main (sw.js, untouched here)
vitest run            849 files, 16388 tests passed
jest doc-paths        37 passed (doc-paths + source-comment-paths)
npm run i18n:check    clean
```

## Checklist

- [ ] An approved discussion or issue exists and is linked above. — none exists; reported and scoped directly by the repo owner.
- [x] New behavior has tests, and the existing suite passes. — `ShareStashSweeper.test.tsx` covers the sweep, the `/share` exclusion, the absent auth gate, re-sweeping on navigation, and that it renders nothing.
- [x] All user-facing strings are translated for **every** locale (i18n parity). — the three banner keys are removed from all 20 locale catalogs; parity and pseudo-locale checks pass.
- [x] No shared/core areas were refactored without prior agreement. — one import swap in `SwipeShell`; the share target, review screen and stash are untouched.

## AI assistance disclosure

Written by Claude Code in a Claude Code on the web session, driven by @kenlasko. Type-check, lint, the full Vitest suite, the backend doc-path guards and the i18n freshness check were run locally before pushing.

Worth disclosing: a first attempt at this removed the entire Web Share Target (6,570 deletions) on a misreading of the report. That branch was discarded; this PR is the narrow change that was actually asked for.

https://claude.ai/code/session_01SA12sacJ3AYJyQ6dCdP1bN

---
_Generated by [Claude Code](https://claude.ai/code/session_01SA12sacJ3AYJyQ6dCdP1bN)_


kenlasko:
## E2E shard 1/4 is red on a push test this diff cannot reach

**Failing check:** `E2E Tests (shard 1/4)`, on its **Run Chromium push integration** step ([job log](https://github.com/kenlasko/monize/actions/runs/34357335978/job/102490637078)).

```
[chromium] › push/notifications.spec.ts:48:5
  › Stop sends browser session and Cookie Store CSRF without navigating

  await expect.poll(async () => (await shown(h.worker)).length).toBe(1);
  Expected: 1   Received: 0
  Timeout 5000ms exceeded while waiting on the predicate

1 failed, 8 passed (11.8s)
```

The worker showed no notification after `h.push(reminder)`.

### Why this is not this PR's

- **The app shell is never loaded by that test.** `e2e/push/fixture.ts` starts its own HTTP server on a loopback port, navigates to a synthetic `/initial` page and calls `navigator.serviceWorker.register('/sw.js')` directly. `SwipeShell`, `ShareStashSweeper` and every other component this PR touches are absent from the page under test.
- **The diff touches no service worker, no push code and nothing under `e2e/`.** It is one component swap in `SwipeShell`, a deleted banner, a new null-rendering sweeper, three removed catalog keys across 20 locales, and docs.
- **`main` is green on this branch's base commit** (`d7b36a9`, CI run #3046).
- **No E2E spec drives the removed banner.** I grepped `e2e/` for its copy (`shared with Monize`, `noticeTitle`/`noticeAction`/`noticeDismiss`, `ShareInboxNotice`) and for `Review`/`Dismiss` as accessible names — no matches. That was the one real risk here, given a green Vitest run says nothing about `e2e/`.
- **Three tests in the same file run the identical push-then-poll sequence and passed** (lines 13 and 35 x2), as did the five Stop tests after it. One failure out of nine, consuming the full 5s poll, is the shape of a timing flake rather than a defect.

### No fix to port

There is no open fix for this that I could carry into the PR — the test is not failing on `main`, so there is nothing to port. Making the poll more robust would be widening this PR into a test it has no other business touching.

### Re-run

The one sanctioned re-run is **pending, not skipped**: `rerun-failed-jobs` returns `403 This workflow is already running` while shards 2/4, 3/4 and 4/4 are still in flight. I have a check-in scheduled to fire it as soon as the run completes, and I am keeping this PR watched until it is green. If the same test fails a second time I will stop treating it as a flake and root-cause it rather than re-running again.

Note that shard 1 never reached its own E2E tests — the push step failed first and the **Run E2E tests** step was skipped — so that shard's coverage is still unproven either way. Everything else on the run is green: both lint and type-check jobs, both unit suites, backend integration, Lighthouse, Bearer, CodeQL and the schema-drift check.

---

kenlasko:
## Follow-up: it failed again, and it is failing on unrelated PRs too

I said above that if this failed a second time I would stop calling it a flake and root-cause it. It failed again, so here is what the second failure and a look across other runs establish.

**It is not this PR's, and now there is direct evidence rather than an argument from the diff.** PR #1335 — a one-line change to `frontend/src/types/ai.ts` swapping a sample model name — failed the *same job* today, with the *same* failure mode, on a *different* test. That diff cannot affect a service-worker notification test.

| Run | Branch | Failing test | Result |
|---|---|---|---|
| [#3047](https://github.com/kenlasko/monize/actions/runs/34354678047) | `claude/practical-lovelace-hgol8b` (PR #1335, one line in `types/ai.ts`) | `:85` failed Stop opens reminders (401) | 1 failed, 8 passed |
| [#3050 attempt 1](https://github.com/kenlasko/monize/actions/runs/34357335978) | this PR | `:48` Stop sends browser session and Cookie Store CSRF | 1 failed, 8 passed |
| [#3050 attempt 2](https://github.com/kenlasko/monize/actions/runs/34357335978) | this PR | `:13` displays real notifications, replaces the same subject | 1 failed, 8 passed |

Every occurrence is identical in shape: exactly one of the nine tests fails, always on the first `shown(h.worker)` assertion after `h.push(...)`, always `Array []` against the expected notification, always consuming the full 5000 ms poll. **Which** test loses is random. A defect in a diff cannot select a different victim on each run; a race can.

Two further facts. The suite is two days old (`5781390`, 2026-09-07), so it has had very few CI runs to expose this. And PR #1335 was merged through the same failure — main's run #3049 happened to pass, which is what a roughly one-in-nine race looks like.

### Proposed direction, deliberately not applied here

I am not pushing a harness fix into this PR. It would widen a banner removal into the push test infrastructure, and I cannot validate a fix for a race I can only observe through CI.

The thing I would look at first is `e2e/push/fixture.ts` caching a single `Worker` handle from `context.waitForEvent('serviceworker')` at registration and reusing it for every later `shown()` / `worker.evaluate()`. A service worker's lifetime belongs to the browser, which may terminate and restart it between the push delivery and the assertion; resolving the *current* worker at assertion time is a real fix, whereas raising the 5000 ms timeout only widens the window the race has to lose in. The `chromium-push-report` artifact on each failed run carries the Playwright trace, which should confirm or kill that hypothesis quickly.

That is a guess at the mechanism, offered as a starting point rather than a diagnosis — the cross-PR evidence above is the part I am confident in.

### Where that leaves this PR

Shards 2/4, 3/4 and 4/4 pass, and those are the shards that actually drive the application. Every other check is green. Shard 1 never reaches its own E2E tests, because the push step fails ahead of them and skips them.

My one sanctioned re-run is spent. I am not re-running again, and I will not skip or quarantine the test to get green.

---

kenlasko:
Correcting my earlier diagnosis on this thread: the E2E push flake was not a lost delivery.

`dc6278f` assumed the CDP push message was sometimes being dropped. It was not. Instrumenting the service worker (wrapping `registration.showNotification` and recording the `push` event) showed that in a failing run the handler ran, `showNotification` was called, and its promise resolved in 7-11ms, while `registration.getNotifications()` answered `[]` for the whole 5s poll and still after it.

The actual cause is that `registration.getNotifications()` is destructive while a display is in flight. Chromium answers it by reconciling its stored notification records against what the platform reports as displayed, and a record whose display has not landed yet is erased rather than reported "not yet". `showNotification` resolves before the display reaches the platform, so the harness's first read deleted the notification the test was waiting for, and no later poll could recover it. That matches every symptom: one of the nine tests failing per run, a different one each time, the full timeout burned, on branches whose diffs cannot touch push.

There is no non-destructive way to observe a displayed notification, so the harness cannot poll its way out; polling is what does the damage. `043fc8d` changes `push()` to wait for the worker's own `showNotification` promise to resolve, look exactly once, and repair a look that came too early by delivering again rather than by looking again. A repeat is safe because `sw.js` tags every notification with `collapseTag(payload)`, so it replaces rather than stacks, which is the property the first test in the file asserts.

Measured against the Chromium in this container, running the whole suite end to end:

| Code | Suite runs | Runs with a failure |
|---|---|---|
| before (`dc6278f`) | 14 | 3 |
| after (`043fc8d`) | 37 | 0 |

Suite wall time is unchanged at about 4.2s.

The rule is written down in `frontend/CLAUDE.md` and scanned: `frontend/src/test/e2e-conventions.test.ts` now fails a `deliverPushMessage` or `getNotifications()` call anywhere under `e2e/push` outside the fixture, which is how a future spec would re-open this. I checked the guard fails on the banned shape as well as passing on the current tree.

No product code changed. `frontend/public/sw.js` was correct throughout.

---

PR #1342, opened 2026-09-09
feat(oidc): surface back-channel failures in OIDC callback

WMP:
Log error.cause (the failing token-endpoint Response: status, statusText,
  content-type, location, body) in the callback catch, so non-conform token
  responses are diagnosable instead of the opaque 'unexpected HTTP response
  status code' from openid-client.

  Add an OIDC_DEBUG=true customFetch hook that logs every back-channel call
  (discovery, token, userinfo) with real status/body. Off by default; it logs
  token bodies, so only for diagnosis.

  Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

## Linked discussion / issue

Closes #
Approved in: 

## Checklist


kenlasko:
This one failed the .env.example check for the new variable

PR #1380, opened 2026-09-13
fix(ui): move the Table/Calendar switch beside the register's count line

kenlasko:
## Linked discussion / issue

No discussion or issue: requested directly by the maintainer in a Claude Code session (a layout tweak to the Transactions page toolbar). The propose-first gate is bypassed for maintainer-authored PRs.

Approved in: n/a (maintainer request)

## Summary

The register toolbar carried export, density and the Table/Calendar switch as one group left of the pager. Download and density were already immediately left of the page selectors; the switch was the control out of place, because it is not a control over the list as drawn -- it changes which shape the list is.

- `Pagination` gains an `infoAfter` slot rendered immediately right of the "Showing 1-25 of 90 transactions" line. `infoRight`, the slot `ui-conventions.test.ts` keys on, is unchanged.
- `ListTopToolbar` passes `viewToggle` to `infoAfter`; `actions` (Download) and `DensityToggle` stay in `infoRight`, on the same line and immediately left of the page stepper. A bar drawn without paging props -- the empty register -- has no count line, so there the switch stays last of the buttons rather than disappearing.
- Order assertions updated in `ListTopToolbar.test.tsx` and `TransactionList.test.tsx`, plus a new case for the no-pager branch.
- Comments in `ViewModeToggle.tsx` and the two `docs/frontend/*.md` lines that named the old placement now describe the new one.

`ListTopToolbar` is the single shared bar, so the Investments register's table toolbar gets the same arrangement. That is by design -- `ui-conventions.test.ts` fails a second copy of this markup -- not a side effect; scoping it to the Transactions page alone would mean a second variant of the bar.

No user-facing strings were added, changed or removed.

Verified from `frontend/`: `lint` clean apart from the pre-existing `public/sw.js` unused-var warning; `type-check` clean; `i18n:check` clean; `test:cov` exit 0 (statements 91.44 / branches 86.09 / functions 89.01 / lines 92.64); `build` exit 0. `node scripts/check-docs-manifests.mjs` OK. Not verified by eye in a running app -- placement is asserted through DOM order, not a screenshot.

## Invariants touched

None.

## Checklist

- [ ] An approved discussion or issue exists and is linked above. (No discussion: maintainer-requested layout change, see above.)
- [x] All user-facing strings are translated for **every** locale (i18n parity). (No strings changed; `i18n:check` clean.)

## AI assistance disclosure

Written by Claude Code in an agent session, from a maintainer's instruction. The gate commands above were run in that session and their results are quoted as reported.

https://claude.ai/code/session_01FtG8uTHs3h4bqYdoZJAwcg

---
_Generated by [Claude Code](https://claude.ai/code/session_01FtG8uTHs3h4bqYdoZJAwcg)_


kenlasko:
`E2E Tests (shard 4/4)` failed on `74a29bb7`, and the cause is not this PR's change.

All five failures are `page.goto: NS_ERROR_CONNECTION_REFUSED` against `localhost:3001`, not assertions. The job's own diagnostics say why: `docker compose ps` reports `monize-frontend-1  Exited (0)`, and the frontend log ends with

```
thread 'tokio-rt-worker' panicked at turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs:292:17:
Restore of All for task TaskId 1037361 failed in another thread: restoring failed
turbo-tasks: an internal panic occurred outside the per-task panic boundary. This is a bug in turbo-tasks/Turbopack
Aborting.
```

The dev server aborted mid-run on a Turbopack filesystem-cache restore (right after "Finished filesystem cache database compaction"), so every spec that ran after it got connection-refused. The same shard passed on `bb7e4448` earlier in this PR. There is no fix to port: the abort is inside Turbopack's cache layer, which this diff -- a reorder of two JSX slots in `Pagination` and `ListTopToolbar` -- does not reach. No E2E spec drives the controls that moved (I grepped `e2e/` for the density, export and Table/Calendar accessible names).

Rather than spend a re-run on a superseded head, the push to `dfed5f8f` re-runs the E2E shards. If shard 4 fails the same way there, I will treat it as real rather than as this crash.

---

PR #1386, opened 2026-09-15
feat(backup): ship automatic backups off the machine — S3 (append-only) and email, 3-2-1

WMP:
## Linked discussion / issue

Closes #1369
Approved in: https://github.com/kenlasko/monize/discussions/1369 (`approved-to-build`)

## Summary

Copies each **complete** automatic backup off the ephemeral container it was written on, so losing the backup volume is no longer losing every recovery point. The local artifact stays primary (written first, read by restore); the off-machine copy is taken afterward, outside every transaction, and can never precede, replace, or delete it.

Two destination kinds, each enabled per user and usable together (a 3-2-1 arrangement):

- **S3 (append-only).** Conditional, checksummed `PutObject` (multipart above a configurable part size), `IfNoneMatch: "*"` on every completing call, no delete/read command in the uploader — the credential and the code path can add an object but never overwrite or delete one. Bucket is the deployment default (`BACKUP_S3_*`) or the user's own with their own key pair, stored AES-256-GCM encrypted and never returned to the client.
- **Email.** The encrypted artifact attached when it is at or under `BACKUP_EMAIL_MAX_BYTES`, otherwise a notice naming the file and its size (no bytes, no link). Encrypted-only.

A durable per-`(user, destination, key)` ledger records every copy; a `pending/failed → uploading` conditional `UPDATE ... RETURNING` makes exactly one replica perform. A retry reaper (hourly at :30) re-attempts failed copies with exponential backoff and a five-attempt ceiling, and reclaims claims whose replica died mid-upload via a 60-minute lease. Only an **encrypted** artifact ever leaves the machine; a plaintext `.json.gz` is refused, recorded `skipped-unencrypted`, and alerted once. Off-by-default: an unconfigured deployment behaves exactly as before.

Design/config background is in `docs/specs/backup-off-machine.md`, `docs/future-plans/backup-off-machine.md`, and the maintainer-decision table therein.

**Deferred — need a maintainer decision (not implemented here):**
- Email key is the bare filename, so a second same-day backup with changed content is not re-emailed (freshness vs. inbox volume for sub-daily schedules).
- The manual "Back up now" request awaits the off-site copy before responding (blocking vs. fire-and-forget UX).
- Only the `daily` artifact is dispatched off-machine (weekly/monthly promotions are local retention); off-machine retention is the bucket's lifecycle. `.env.example` should note the S3 endpoint must support/echo `x-amz-checksum-sha256`.

## Invariants touched

- **INV-BACKUP-002** — only an encrypted artifact leaves the machine (dispatcher + email-sender gates + source-scanning guard).
- **INV-BACKUP-003** — a complete local copy exists and is recorded before any off-machine copy; a copy failure never fails the backup.
- **INV-BACKUP-004** — the application cannot delete or overwrite an off-machine copy (no delete/read import, `IfNoneMatch` on every completing put; operator IAM half documented).
- **INV-BACKUP-005** — an off-machine copy is verified (checksum / SMTP acceptance) before it is recorded done; stale claims reclaimed by lease.
- Builds on **INV-BACKUP-001** and **EXT-001**..**EXT-004** (`docs/external-side-effects.md`).

## Checklist

- [x] This PR addresses a **single concern** (shipping automatic backups off the machine; delivered as the WP1–F1 series agreed in the discussion, on one branch).
- [x] No shared/core areas were refactored without prior agreement (the shared S3 transport extraction and the `auto-backup.service.ts` dispatch hook were in the approved scope).

**Verification (this branch):** backend `lint` + `tsc` + `typecheck` + `TZ=UTC test:unit --coverage` (639 suites / 16,316 tests; coverage 95.61/87.01/95.96/96.53) + `i18n:check` + `migration:lint` + repo check-scripts; frontend `lint` + `type-check` + `i18n:check` + `test:cov` (903 files / 17,418 tests; 91.45/86.13/88.95/92.66) + `build`. Not run in this environment: `test:integration` (needs PostgreSQL — the two-connection single-winner claim spec is included and runs in CI), `scripts/verify-schema.sh` (needs Docker), and a real S3/MinIO round-trip (the uploader is proven against a local fake-S3 endpoint).

## AI assistance disclosure

Implemented with AI assistance (Claude Code): the plan and spec, the backend/frontend implementation, tests, and translations. Reviewed with the repository's adversarial audit protocol and a high-effort code review; the author has reviewed and owns the correctness, conventions, and tests.

https://claude.ai/code/session_01LXeKrKTFhEtL3hpWctds3o


WMP:
@kenlasko : I havent SMTP server, can i ask you to check this pr? https://github.com/WMP/monize/actions/runs/35000551788

kenlasko:
Running this code now. Will report back tomorrow with results. 

Rather than a separate section, I would roll this under Automatic Backups, and have an icon representing if backups were either successfully uploaded/emailed or not. Don't have a separate list. Might also be a good idea to put the list of backups inside a window of a set size so it doesn't take up too much space.

WMP:
The maintainer asked to roll the off-site destinations UI under **Automatic
Backups**. That collides with the earlier decision #5 (per-user destinations with
their own S3 credentials), because Automatic Backups is an admin-only section
(`AutoBackupController` is `@Roles("admin")`; "automatic backups are an operator
setting, not a user preference"). Two ways to resolve it — which do you want?

**A. Off-site becomes admin-only.** All configuration moves onto the admin page,
the controller gets `@Roles("admin")`, and "per-user S3 credentials" goes away
(one deployment-wide configuration remains). Simpler and consistent with "backup
is an operator setting" — but it reverses decision #5.

**B. Stays per-user, only visually merged.** The controller stays
`AuthGuard('jwt')`, but the destinations UI + status icon move inside the
"Automatic Backups" card where the user sees it (today an admin sees the
auto-backup section; a regular user does not — so a regular user would need a
trimmed "Automatic Backups" card showing only their own destinations). Keeps
decision #5, but complicates who sees what.

kenlasko:
I should have clarified this belongs in the `Automatic Backup` sub-section of the `User Settings - Backup & Restore` section

kenlasko:
Works as expected. Email came through fine with the attachment intact. 

One other change is to move the Save Destinations button to be directly under the S3 settings and add a Test button for verification. Changing the other settings should not require pressing a Save button.

PR #1404, opened 2026-09-18
feat(ai): move the reverse MCP relay onto rows (R3-R5)

kenlasko:
## Linked discussion / issue

Closes #
Approved in: [`docs/future-plans/horizontal-scaling.md`](docs/future-plans/horizontal-scaling.md) (WP5) and its task list [`horizontal-scaling-tasks.md`](docs/future-plans/horizontal-scaling-tasks.md) — the committed plan this series executes. This PR is tasks **R3, R4 and R5**; R1 (the event bus) and R2 (the tables) are already on `main`.

## Summary

The AI relay's queue lived in six `Map`s in `ai-relay.service.ts`. That made the reverse relay a single-process feature: the browser's SSE stream, the agent's long-poll and the agent's answer each had to reach the same replica, and a restart between "enqueued" and "answered" lost the turn with nothing able to notice. This moves all of it onto the tables R2 added.

**R3 — the queue.** The browser inserts a `pending` `ai_relay_prompts` row and parks; `get_next_prompt` claims with `FOR UPDATE SKIP LOCKED` and a conditional `UPDATE`, so two agents polling one user claim different prompts and the loser of a double `post_response` is refused by the database rather than by a read it had already passed. `expires_at` carries the whole deadline — the queue wait, the idle window every liveness signal pushes out, clamped to `claimed_at + HARD_WAIT_MS` — which is what lets a second replica move a browser's deadline. A wake-up on the `EVENT_BUS` only shortens the wait: it carries ids, never the prompt or the answer, and every waiter re-reads on a slow timer in case the message was lost.

**R4 — liveness and cards.** The agent's poll, inactivity clock and disconnect notice become the `ai_relay_agents` row (one upsert, so two replicas serving one agent cannot each start a clock neither finishes). A confirmation card composed after the browser's stream gave up becomes an `ai_relay_actions` row, drained by whichever replica serves the pickup. A new five-minute sweep keeps the tables from being append-only; nothing the relay does depends on it running.

**R5 — attachments.** `ai_relay_attachments` plus a cascading `ai_relay_attachment_blobs`. **Deviation from the plan, stated:** WP5 said route these through `ATTACHMENT_STORAGE_PROVIDER`. That cannot work — the provider's `database` implementation writes `attachment_blobs`, whose primary key is a foreign key to `transaction_attachments` and whose RLS policy reads the owner from that row, so a relay attachment (which has no transaction) fails the foreign key outright. The alternatives were weakening that foreign key or branching on the bound provider's name inside the relay. The table pair mirrors `transaction_attachments`/`attachment_blobs` instead and buys what the provider could not: bytes and metadata commit or roll back together on every deployment, and the cascade reclaims the bytes, so the sweep has nothing outside PostgreSQL to order against. The cost — a deployment keeping attachment bytes out of PostgreSQL still holds a few megabytes of chat scratch there for ≤20 minutes — is recorded in `docs/external-side-effects.md` and in the task's notes.

Only the open SSE sockets stay process-local, in `RelayStreamRegistry`, beside the controller's heartbeat: an `emit` closure writes to a socket this process holds.

### Deliberate behaviour changes in `single`

- A prompt survives a backend restart mid-wait (the plan's named change).
- A second `post_response` returns `delivered:false` — the `UPDATE` matched nothing — instead of an idempotent `true`. The tool's own description already said `delivered:false` means "unknown or already answered".
- A closed browser socket ends its waiter at once instead of parking out the deadline. The row is untouched, so the pickup endpoint still serves a late answer.
- The per-user caps on buffered answers, cards and attachments are gone. They bounded process memory; `expires_at` and the sweep bound a table.

### Review pass

The last commit is the result of a full adversarial review of the first three. It fixes four paths where a write that had already succeeded could be reported as a failure — and an agent's documented response to a failure is to retry: an unpublishable wake-up failing a committed row, the inactivity reset running after the claim had committed, `reportToolActivity` throwing from the MCP wrapper's `finally` (where it replaces the tool's own outcome), and attachment cleanup destroying the answer it was cleaning up after. It also gives the claim's `UPDATE` the `status = 'pending'` its docstring claimed, makes liveness pick this session's newest turn rather than an unordered `RETURNING`'s first row, and gives the sweeper's `DELETE` a cutoff strictly later than the expiry it shares a transaction with — with the same cutoff, a turn was expired and deleted in the same tick.

The same pass found the MCP tool specs mocking the now-async relay methods with `mockReturnValue`, which returns a plain value — so `await x` and `x` were indistinguishable and neither of the two missing awaits found earlier would have failed them. Converted to `mockResolvedValue` and verified by mutation: dropping an `await` on `postResponse` now fails 2 tests where it previously left all 11 green.

## Invariants touched

**INV-HA-005** — a relay prompt is claimed by exactly one agent poll and answered at most once. Enforced by the `FOR UPDATE SKIP LOCKED` CTE plus each transition's conditional `UPDATE ... RETURNING`, proved by a two-connection integration spec. (`INV-HA-001..005` are catalogued in `docs/system-invariants.md` by task G2, not here.)

No financial invariant is touched: the relay routes prompts and answers and never computes a figure.

## Checklist

Notes on two of those boxes:

- **i18n**: no new user-facing string. The relay's four error messages are unchanged; `npm run i18n:check` is clean.
- **Single concern**: three plan tasks in one PR because R4 and R5 only exist to finish what R3 starts — R3 leaves four `Map`s behind that R4 removes, and R5 is the last piece of relay state that a second replica cannot see. They are three commits, reviewable in order, plus the review-fix commit.

### Verification

Run locally against a real PostgreSQL 16 at `ab3db55`:

- `npm run lint`, `npx tsc --noEmit`, `npm run typecheck`, `npm run i18n:check` — clean.
- `TZ=UTC npm run test:unit -- --coverage` — 652 suites, 16501 passed, 2 skipped; 95.64 stmts / 87.01 branch / 96.00 funcs / 96.55 lines against thresholds 94/85/95/95.
- `npm run build && npm run test:integration` — 57 suites, 615 passed.
- `npm run migration:lint`, `check-migration-prefixes.mjs`, `check-env-docs.mjs`, `check-docs-manifests.mjs` — OK.
- `scripts/verify-schema.sh` needs Docker, which the session did not have; its exact procedure (schema.sql baseline, all 193 migrations replayed twice, normalized `pg_dump` diff) was reproduced against a local cluster and reported no drift. **The CI `schema-drift` job is the authoritative run of this one.**

## AI assistance disclosure

Written with Claude Code (Opus 5), including the adversarial review pass and its fixes. Every claim above was verified by running the suites and the check scripts; the mutation results quoted are from actual runs, not inspection.

https://claude.ai/code/session_01GuGN6pCsLfaLM7Mww92gpd

---
_Generated by [Claude Code](https://claude.ai/code/session_01GuGN6pCsLfaLM7Mww92gpd)_


kenlasko:
@WMP can you test the MCP relay with this once the beta is built?

WMP:
I test only on single instance - works like before. 

I wonder if we actually need Redis here.

For the two planned use cases, PostgreSQL already has good equivalents:

* `LISTEN / NOTIFY` could replace Redis Pub/Sub for relay wake-ups, while PostgreSQL rows remain the source of truth and the existing polling fallback still handles missed notifications.
* A small `UNLOGGED` PostgreSQL table with atomic `INSERT ... ON CONFLICT DO UPDATE` could provide shared rate-limit counters across replicas.

That would keep the whole HA design on one shared dependency instead of adding Redis, Redis HA, monitoring, upgrades, credentials, and another failure mode just for two small ephemeral features.

If PostgreSQL-backed throttling ever becomes a measurable bottleneck, Redis could still be added later behind the existing abstractions.

PR #1412, opened 2026-09-19
fix(securities): stop the portfolio summary's periodic CPU storm

### LINKED ISSUE #1409: PortfolioService causing periodic high CPU utilization
This is very preliminary but wanted to bring this to @WMP attention, since this is the area you've been working on.  I built images after PR #1397 and am running this in production.

I see there's a DEBUG log for `PortfolioService`.  I'm noticing that when I see these logs in my backend:
```
[Nest] 7  - 09/18/2026, 8:13:22 PM   DEBUG [PortfolioService] Portfolio summary phase sinceInceptionResult: 47364ms
[Nest] 7  - 09/18/2026, 8:13:22 PM   DEBUG [PortfolioService] Portfolio summary computed in 47961ms
```

 the backend pod CPU utilization gets very high, and the frontend starts throwing:
```
2026-09-19T00:11:17.356Z [Proxy] ERROR: API proxy error: TypeError: fetch failed
```
Both backend and frontend pods become unhealthy on and off until the CPU eventually settles down to baseline.  

Pods with new images were started at approximately 8:00pm. This is CPU for that period:

Image

CPU peaked at 0.5.  Previous peak CPU in past week was a small spike to 0.05 three days ago.

And memory (red line is the requests, which is 140MB). Peaked at 480MB. Previous peak was 300MB at various points in the past week: 

Image

kenlasko:
## Linked discussion / issue

Closes #1409
Approved in: https://github.com/kenlasko/monize/issues/1409#issuecomment-5738034366 (the diagnosis and the ordered fix list), confirmed in https://github.com/kenlasko/monize/issues/1409#issuecomment-5739444118

## Summary

One concern: the portfolio summary's periodic CPU storm, in the four parts the issue asked for -- plus a fifth the `CLUSTER_MODE=multi` E2E shard demanded before this could be green.

**1. Classify Yahoo's 400 correctly** (`yahoo-finance.service.ts`). Yahoo refuses a window it has no history for with a 400 whose `chart.error.description` reads `Data doesn't exist for startDate = ...` -- a statement about the symbol and the window, the same class as the 404/422 the client already reads as "no such series". The status alone was the whole test, so it came back as `null` ("the provider declined"), `fillRateWindow` reported `answered: false`, `EmptyWindowMemory` could never be written, and the next read asked the identical questions again. That is the loop with no convergence behind "periodic": 76 doomed calls on every summary. A blanket 400 would be wrong -- a request this client built wrong is also a 400 -- so the description decides. The body is read with `text()` rather than `readBody`, because `throttledFetch` has already recorded the non-OK response with the breaker.

**2. The summary no longer backfills rate history inline** (`portfolio.service.ts`). `getInvestedResultSinceInception` opens at the scope's first investment transaction (2000-02-10 on the reporter's data), so every month with no stored rate became an `ensureRatesForDate` call inside the GET a person is watching a spinner for. It now passes `fetchMissing: false` and reports the gap through the honest contract it already has (`timeWeightedReturnReasons`, `incompleteRanges`); filling it belongs to `ExchangeRateHistoryService` and the scheduled jobs.

**3. The walk itself, which is the real work.** Not a clamped start date and not a long-TTL cache: profiling found the walk quadratic for a reason neither would fix. `resolveFxRate` read *every* stored observation of a pair on *every* call, at two `Date.parse` calls per row. On the incident's shape -- ~9,700 daily points against an index holding ~9,700 rows per direction -- that is **66.3 s of blocked event loop for one pair**, before a single holding is multiplied by anything. Node serves `/api/v1/health/ready` on the same loop, which is the readiness failures, the dropped Service endpoint and the frontend's `TypeError: fetch failed`.

Each direction's observations are now prepared once (rows with a usable rate and a date, sorted ascending with ties in source order, memoized in a `WeakMap` against the array the caller handed over) and the admissible one is found by binary search over the age window expressed as two calendar dates. The same 9,700 lookups: **83 ms**. Every FX-converting surface gets this, not only the summary.

**4. The memo's TTL runs from settle, not from start** (`portfolio-summary-memo.ts`). A 48 s valuation was servable for the remaining 12 s of its 60 s window, so the next page open paid for the whole walk again. An in-flight entry now has no expiry at all (concurrent callers still share one walk); an invalidation landing mid-walk drops the entry, and the settling value does not resurrect it.

**5. An invalidation reaches every replica** (`portfolio-summary-memo.ts`, `portfolio-summary-invalidation.bridge.ts`). Not planned, and not optional: the cluster shard failed on this branch, deterministically, and what it caught is a defect the memo has had since #1397. Invalidation was a local call, so the replica that served a write dropped its entry and no other replica heard; behind the round-robin LB the next read landed on one that had not, and the Investments page showed the portfolio as it was BEFORE the trade just entered -- `0 positions + Cash $10,000.00` -- for up to the 60 s TTL. You could not read your own write.

`invalidatePortfolioSummary` now drops locally and announces on the existing `EVENT_BUS`; `PortfolioSummaryInvalidationBridge` subscribes every replica and applies what arrives, never answering an announcement with an announcement. What travels is a user id, which is all the bus permits: the recipient drops a cache and recomputes from the database under its own scope. `NOTIFY` can be lost, so the TTL remains the bound rather than the mechanism, and the announcement rides the same seam as the local drop -- including callers that invalidate inside their own transaction, where both happen a moment before the commit is visible. That window is the seam's and predates this channel. The bus is bound `@Optional()`, because `EventBusModule` is `@Global()` rather than imported per feature module and the integration harness assembles a graph without it.

Six commits, each with its regression test and its doc line. Every one fails on the code before it, verified by mutating each fix back -- including the Nest resolution error the `@Optional()` binding fixes.

## Invariants touched

- **INV-FX-001** -- upheld unchanged. The resolver rewrite is an implementation swap behind the same policy, tested against the previous implementation; `fetchMissing: false` means a pair with no stored history stays withheld with its cause rather than converted or defaulted.
- **INV-CACHE-001** -- strengthened, and its catalog entry updated. The memo still hangs off the post-commit balance seam, and the TTL's start point moved to settle; beyond that, the invalidation now crosses replicas, so "process memory with a 60 s TTL, so on a second replica the bound is the TTL rather than the invalidation" is no longer what the code does. The entries stay per replica; the announcement is what is shared.

## Checklist

Notes on two of those boxes, so a reviewer can check rather than trust:

- **i18n:** no user-facing string changed. The withheld-figure copy the summary now shows more often already exists and is already translated.
- **Shared/core:** `fx-rate-resolver.ts` is the one shared door this touches -- its policy is deliberately untouched (no look-ahead, the same `FX_MAX_RATE_AGE_DAYS` bound, the more recent admissible observation in either direction, a tie to the direct pair, the same named gap reasons), and the spec keeps the old scan as `scanForBest` to test the lookup against it over randomized histories. The cross-replica invalidation in part 5 is the one new cross-cutting behaviour: it was put to the maintainer in a comment on this PR before it was written, and it is built on the `EVENT_BUS` that already exists rather than a mechanism of its own.

## Verification

Locally, on the final commit:

- `npm run lint`, `npx tsc --noEmit`, `npm run typecheck` -- clean
- `TZ=UTC npm run test:unit -- --coverage` -- 685 suites, 17,058 passed, 2 skipped, exit 0 (95.82 stmts / 87.25 branches / 96.27 funcs / 96.75 lines)
- `node scripts/check-docs-manifests.mjs`, `node scripts/check-env-docs.mjs` -- OK
- `test:integration` needs PostgreSQL, which this session has no daemon for; CI runs it, and it is green on this head.

In CI, on this head: `Backend Integration Tests` green, and **`E2E Tests (shard 4/4, cluster)` green** -- the shard that was red before part 5, which is what proves the cross-replica hop rather than any unit test.

## AI assistance disclosure

Written with Claude Code, from the diagnosis in the issue, and driven through CI in the same session. The measurements are mine: 66.3 s from a standalone transcription of the previous `bestInDirection`, 83 ms from the committed code under Jest.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017vPcEL55C23DyFg19gFzRX


kenlasko:
**`E2E Tests (shard 4/4, cluster)` is red, and what it caught is real — but it is not this PR's logic.**

The failing test is `investments.spec.ts › records a BUY through the transaction form`, on the cluster shard only (two replicas behind the round-robin LB, no affinity). After the BUY is recorded through the UI, the account header still reads `0 positions + Cash $10,000.00` — the complete pre-write state — for the full 15 s. Everything else on this head is green, including `Backend Integration Tests`, both unit suites and `cluster.spec.ts` itself; the second shard-4 failure (`mobile.spec.ts`) was flaky and passed on retry.

PostgreSQL is shared between the replicas, so nothing but a per-replica cache can serve that page. It is the portfolio summary memo (#1397): the page load memoises the pre-write summary on whichever replica served it, the POST invalidates the memo **only on the replica that handled the write** (`invalidatePortfolioSummary` is a local call), and the post-write GET round-robins to the other replica, which answers from its own entry for up to `PORTFOLIO_SUMMARY_MEMO_TTL_MS`. `docs/backend/securities-and-providers.md` states that staleness as an accepted bargain; the cluster shard, one day old, is the first thing to show it is user-visible — in a two-replica deployment you cannot read your own trade.

Which replica answers the post-write GET is effectively a coin flip, so this test is red roughly half the time on **any** commit: `main` has two green samples, this head has one red (3/3 attempts within the job, because the request sequence is deterministic inside a run). This PR's perf work plausibly shifted the flip — the fresh computation went from seconds to ~30 ms, which reorders concurrent requests at the load balancer — but I could not prove that mechanism and am not claiming it.

**Why I have not pushed a fix here.** The repair is cross-replica invalidation: publish the invalidation on the existing `EVENT_BUS` (`common/events/`, `PostgresEventBus` over each replica's `LISTEN` connection in `CLUSTER_MODE=multi`) and have every replica drop the user's entries, with the TTL remaining the bound because `NOTIFY` can be lost. That is a new cross-replica behaviour belonging to the horizontal-scaling work and `INV-CACHE-001`, not to "the portfolio summary's periodic CPU storm", and `AGENTS.md` is explicit that a new abstraction beyond the task is a separate proposal rather than a drive-by. It is the author's call whether it rides here or lands as its own PR, and I will implement it either way.

A re-run of the run is already in flight (attempt 2), so the one permitted re-run is spent; if shard 4 comes back green on the same commit with no code change, that is the coin flip confirmed rather than the defect fixed.

---

WMP:
I was initially planning a much larger architectural change for this problem: persist the daily investment valuation as derived state, keep a `dirty_from` boundary per account, and incrementally rebuild only the affected tail when transactions or historical prices change.

After looking at this PR, I think this is a better solution for the actual problem we observed. The expensive part was not inherently the ~9,700-day replay, but the FX resolver repeatedly scanning the full FX history for every day. Fixing that lookup, together with `fetchMissing: false`, the Yahoo empty-window handling and the settle-based memo TTL, keeps the current deterministic "ledger + prices + FX -> result" model without introducing another persistent state and all of its invalidation complexity.

One thing from the incremental design may still be worth keeping as a future guard: add a realistic long-history benchmark/work-count test (20–30 years, multiple securities and currencies) and make "portfolio summary must never perform provider I/O" an explicit invariant. If profiling later shows that the remaining daily replay itself becomes material, incremental materialization can still be introduced then, based on measured need rather than as a workaround for the FX hot path.

kenlasko:
**Correction, and the outcome.** My earlier comment called this a coin flip that would be red on roughly half of all runs, `main` included. That was wrong: a second job on the same commit failed identically, so it was deterministic on this branch, and the "~50/50" reasoning should not be relied on.

The cause was what that comment named, though, and the repair is now in this PR as part 5: the memo's invalidation is announced on the `EVENT_BUS` and applied by every replica. `E2E Tests (shard 4/4, cluster)` is green on `7cc7107`, along with `Backend Integration Tests` -- the shard is what proves the cross-replica hop; no unit test can.

One intermediate commit of mine broke the integration job in passing: the bridge required `EVENT_BUS`, which `EventBusModule` provides `@Global()` rather than per feature module, so the harness's graph would not compile. Fixed by binding it `@Optional()` -- a graph with no bus has no other replica to tell -- with a unit case that reproduces the exact Nest resolution error, so that class of breakage cannot reach the integration job again.

---

PR #1425, opened 2026-09-21
fix(investments): measure a portfolio period over the window its range names

kenlasko:
## Linked discussion / issue

No prior discussion or issue exists. Scope was agreed with the maintainer directly, in the session that produced this work, defect by defect as each was diagnosed.

An earlier revision of these commits cited `#1424` as the issue for these defects. It is not — it is a merged PR about filling gaps in stored FX rate history. That reference was fabricated and is removed in `99044949`; no number replaces it, because none exists.

## Summary

**This PR does not address a single concern.** It carries four, because each was found while investigating the one before it. They are separable and could be split on request; the box below is left unticked rather than argued around.

### 1. A portfolio period is measured over the window its range names (`b047fd99`)

The window a price chart **draws** is deliberately not the period its button names: `portfolio-range-window.ts` opens 3M, 1Y, 2Y and 5Y a day early so the first plotted close precedes the period, `resolveRangePreset` widens `1d` to a week so a daily fallback has more than one point, and `all` resolves to an empty start date. Every portfolio surface sent that drawn window to `GET /net-worth/investments-period-result`, so:

- **1D measured a week.** On a Monday its tooltip dated the baseline seven days back.
- **3M disagreed with the Portfolio performance card beside it** — the chart opened a day before the quarter, the card used the server's `today - 90`.
- **All time asked nothing at all** and read n/a for every account, because an empty start date failed the client's own guard.

The window is now **named, not dated**: the range goes out as `period` and the server resolves it from `portfolio-period-presets.util.ts`, the same file the batch route behind Portfolio performance resolves its windows from. Two cards on one page can no longer open a quarter on different days. `getInvestedResultSinceInception` delegates to `period: 'all'`, so the all-time window has one spelling. `mtd` has no preset and still sends its own dates.

**`startPriceDate`** is new on the payload: the trading session `startDate`'s value came from, via `NetWorthService.getLastPricedDays` (the mirror of `getFirstPricedDay`). `startDate` is a calendar day and the series prices every calendar day from the latest close at or before it, so a Monday 1D window opens on Sunday and is measured from **Friday's** close. The Investments chart and the Portfolio Value report print it as "Since the close of trading on ..." beneath the chart title, for every range, replacing a prior-close tooltip that appeared on only three of them. `null` is unknown, never the calendar day substituted.

### 2. A share transfer is valued at the day's close, not its carried basis (`b2beb932`)

Found by triaging why the maintainer's accounts still withheld their figures after (1). A portfolio built by transferring holdings in reported n/a for every window those transfers fell in — every long window, and always All time.

The cause was not which rows were counted but what value they were given. `foldInvestedFlows` valued a share-moving leg at the basis on the row (`quantity × price`, often nothing at all on an imported row) while `IV` valued the same shares at the day's accepted close. The difference had to be withheld, because it would otherwise read as a gain nobody made.

A leg is now valued at `quantity × close`, where the close is `positionCloseAsOf` over the very series the valuation was built from (`shareLegCloseFrom`). `IV` and `K` move by one number for one day's shares and cancel exactly, so the P&L is the market's alone whatever cost the row recorded. The loader carries the security and the shares to value them from; both period routes pass the same resolver, because two price-resolution rules would let the two sides disagree again.

`externalShareTransfers` is split out of `externallySettledTrades`. It still withholds the **account-level** result — shares entering `MV` with no cash crossing the boundary is a subtraction that was never the market's — and no longer withholds the invested figures, which can now measure it. Cash-settled and mixed-split arms are unchanged.

Where nothing priced a leg the window is still withheld, with the cause a reader can act on: the fold names the security on that day, the reason is `incompletePrices` rather than `externallySettledTrade`, and `withFlowUnpricedSecurities` dates it into `incompleteRanges` so the diagnostics send the reader to one price history. A bare count of uncountable movements was a dead end.

This closes both open items of spec section 10.6.

### 3. The performance periods align in columns (`cbf3e0cb`)

Each period was its own flex row, so the amount and the ratio started wherever their own text did: `+$1,016.00 +1.41%` and `+$12.00 +0.02%` put their figures in different places. The list is a grid of label, amount and ratio, so every period's cells share a column. The third column exists only where a caller supplies a second figure — the security card reports a ratio alone — and a row with no second figure of its own renders an empty cell rather than "n/a", because absent is not unknown.

### 4. Unlinked transfer legs are suggested and linkable (`4e9072d6`)

`transferSecurity` writes `linkedTransactionId` on both rows it creates, and the cost-basis replay matches on exactly that: an unpaired `TRANSFER_IN` cannot take the basis its source released, so that position reports `transferred_basis_unknown` and the reader loses the gain over it. Imported rows were written independently and carry no pairing.

Two legs on one day, of one security, for the same shares, in opposite directions, on two accounts of the portfolio, neither already paired, is strong evidence of one transfer recorded twice — and still a person's call. **Finding is a suggestion, linking is a command.** No measure reads the candidates; inferring a pairing inside a valuation would be a guess wearing a figure's clothes. The write locks both rows in id order, re-checks every condition inside the transaction that links them, and refuses before writing anything while naming which condition failed. The basis rebuild is in that same transaction, because the basis it writes is only correct against the link. The card renders nothing when there is nothing to suggest, and stays silent when the look itself failed — a failed request is not a tidy ledger.

## Invariants touched

INV-PORTRESULT-001 and INV-PORTRESULT-002. Neither status changes: nothing here derives a figure from the plotted series, and the invested measure still excludes cash from both the result and the base of its percentage. `docs/specs/portfolio-period-result.md` sections 6.1, 7, 9 and 10.6 are updated, and 10.6's two open items are closed by (2) rather than worked around.

## Checklist

- [ ] An approved discussion or issue exists and is linked above. **Not ticked deliberately.** None exists; approval was given directly by the maintainer, as described above.
- [ ] This PR addresses a **single concern**. **Not ticked deliberately.** It carries four, each found while investigating the one before it. Splitting is offered above.
- [ ] No shared/core areas were refactored without prior agreement. **Not ticked deliberately.** `invested-capital-flow.util.ts`, `unmeasured-flows.util.ts`, `incomplete-data-ranges.util.ts` and the shared `PerformancePeriodsCard` (read by two surfaces) all changed. Each was agreed with the maintainer in-session as the defect was diagnosed, but not through a Discussion.
- [x] The branch is rebased on the latest `main` (`d162a285`, zero commits behind).

### What has been verified, precisely

Passing on this head: backend `lint`, `tsc --noEmit`, `typecheck`, `i18n:check`; the 379-test `src/net-worth` suite and the 116 tests across the new securities specs, the investment-transactions controller and the module graph; frontend `type-check`, `lint`, `i18n:check`, the i18n parity suite over all 20 locales, `ui-conventions`, `balance-cache.guard`, and every affected component and hook suite. The `invested-capital-flow` integration spec passes against a real PostgreSQL 16 loaded with `database/schema.sql` — the check that matters for (2), since a `GROUP BY` change is invisible to mocked queries.

Both full layer gates (`TZ=UTC test:unit --coverage`, `test:cov`) were re-running on this exact head when the PR was opened, and `npm run build` had not yet re-run after the comment-only edit in `99044949`. An earlier head passed both — backend 697 suites / 17352 tests at 95.84/87.34/96.25/96.77, frontend 911 files / 17617 tests at 91.5/86.16/89.06/92.71 — and `99044949` changes only comment text. CI on this head is the authority; I will report the local results in a comment rather than leave this paragraph as the claim.

The wider backend integration suite was not run green here: several unrelated suites (migration path, event bus) need setup the authoring environment lacks, and they fail for that reason rather than for anything in this diff.

## AI assistance disclosure

Written with Claude Code: the diagnosis, the implementation across both layers, the tests, the spec updates and the translations for all 20 locales.

Three corrections worth recording, because each was a wrong claim I made before the evidence arrived:

1. I reported the backend suite green off a focused run; the full suite then failed 38 i18n tests because the new English error key was untranslated (`8336fcc5`).
2. I twice named a likely cause for the withheld figures before looking at the data — first brokerages with no linked cash sleeve, then trades funded from a bank account. Diagnostic queries run against the real schema showed both were absent: all 103 rows were share movements, 74 of them carrying a basis, which is what redirected the fix to (2).
3. I told the maintainer that linking unpaired transfer legs affected no visible figure. That was wrong — I had checked only the period-result path, and `portfolio-calculation.service.ts` matches legs on `linkedTransactionId` to carry cost basis. That correction is why (4) exists.

---

https://claude.ai/code/session_01CbXhSFHzB3Pf9EWEEG3jQ9

---
_Generated by [Claude Code](https://claude.ai/code/session_01CbXhSFHzB3Pf9EWEEG3jQ9)_


WMP:
I18n is broken, in every language it use word "periods"

kenlasko:
Correcting the checklist: **"New behavior has tests, and the existing suite passes" was ticked in error.** The full backend gate finished after I opened this PR and it was red.

`src/securities/investment-void-classification.guard.spec.ts` refused both readers of `investment_transactions` that the transfer-pair work added. Every reader of that table has to declare at its query site whether it reads rows as **records** (VOID included) or as **effects** (VOID excluded), and neither of mine said so in a form the scan recognises:

```
securities/transfer-pair-link.service.ts:100: const repo = manager.getRepository(InvestmentTransaction);
securities/unlinked-transfer-pairs.util.ts:71: FROM investment_transactions o
```

The guard was right on both counts, and the two answers are different:

- The **candidate query** reads both legs as effects. It already renders `status != 'VOID'` through `investmentEffectStatusSql`; only its comment failed to say so in the words the scan looks for.
- The **link path** reads rows as records, and deliberately. A void leg has to be *read* in order to be refused, so `assertPairable` can name it as the reason the pairing was rejected. A query that filtered it out would report the row as simply not found and send a reader looking for a deletion that never happened.

Fixed in `79855e42`. The guard and the two new suites pass (26 tests), and the full backend gate is re-running on that head.

I had also ticked the box on the strength of an earlier green run plus focused suites, on the assumption that a purely additive feature could not break a source-scanning guard. That assumption was wrong, and the guard exists precisely to catch it. I will post the gate numbers here when both layers report, and will not re-tick the box until they do.

---

kenlasko:
@WMP thanks — you're right about how it reads, and I looked into it properly. Three findings, one of which means the fix is bigger than the two strings you spotted.

**1. It is not confined to `periods`, and it is not new here.** `netWorth.invalidPeriods` is already on `main` (commit `79762985`); this PR added only the singular `invalidPeriod`, following the pattern. I scanned `en/errors.json`: **38 messages** open with a bare English identifier, across `accounts`, `ai`, `delegation`, `monteCarlo`, `netWorth`, `securities`, `transactions` and `calendar`:

```
startDate must be YYYY-MM-DD
accountIds must be comma-separated UUIDs
granularity must be 'daily' or 'monthly'
amountFrom must be a number
baseUrl must be a valid HTTP or HTTPS URL
...and 33 more
```

**2. The leading word is deliberate, but the result is still wrong.** That token is the query-parameter name the caller has to fix, so it is not translated. The intent is defensible; the effect is not. In Polish you see

> `periods musi być listą wstępnie zdefiniowanych okresów rozdzielonych przecinkami`

which opens with an English word and reads as a broken string — exactly the conclusion you drew. A translated sentence should not begin with an untranslated identifier.

**3. The obvious fix produces bad grammar, so it needs whole sentences, not a prefix.** Prepending a translated lead-in and quoting the identifier looked mechanical until I tried it:

| locale | today | naive prepend | why it breaks |
|---|---|---|---|
| pl | `dateFormat jest za długi` | `Wartość "dateFormat" jest za długi` | `długi` is masculine, `Wartość` feminine → `długa` |
| ru | `dateFormat слишком длинный` | `Значение "dateFormat" слишком длинный` | needs `длинное` |
| ru | `accountIds должен быть массивом` | `Значение "accountIds" должен быть` | needs `должно быть` |

Grafting a lead-in onto 627 existing predicates would ship grammatically wrong Polish, Russian and Ukrainian, and I could not review that many by eye.

**The fix**, and it is coming as its own PR rather than this one: 19 parameterised templates replacing the 38 messages, each written as a complete sentence per locale with the identifier interpolated as quoted data. It also collapses the nine separate "must be YYYY-MM-DD" strings into one, which is what let them drift apart in the first place.

```
en: The value of "{{ param }}" must be a date in YYYY-MM-DD format
pl: Wartość "{{ param }}" musi być datą w formacie YYYY-MM-DD
ru: Значение "{{ param }}" должно быть датой в формате YYYY-MM-DD
```

It is kept separate because this PR already carries four concerns, and 361 translated strings plus 38 call sites deserve review on their own rather than buried under them. Your report is what found it — thank you.

One note on severity for whoever picks it up: most of these are query-parameter rejections unreachable from the Monize UI, which only ever sends valid values; `invalidPeriod` in particular can only be triggered by a third-party API or MCP caller. That is a reason to fix it calmly, not a reason to leave it.

---

WMP:
About 3, i have little help for you about polish :) 

image

kenlasko:
## Gate results, and a third correction

CI's **Backend Integration Tests** failed on `d0913722`, and it was mine. Fixed in `54eaa9ee`.

`test/integration/unmeasured-flows.integration.spec.ts` pins the exact shape of `loadUnmeasuredFlowRows`. Splitting the share-movement arm out of `externallySettledTradesSql` gave it a third statement and a third key, so both its assertions failed:

```
+   "externalShareTransfers": Array [ Object { "count": 0, "date": null } ],
    "externallySettledTrades": Array [ ...
```

**Why no unit suite caught it, and why I should have:** a mocked `query` accepts any shape and any parameter list. That spec exists *because* of a past bug where a statement was bound five parameters while referencing only four, and PostgreSQL refused it at PARSE — invisible to mocks. I changed the file that spec covers, ran only the `invested-capital-flow` integration spec, and told you that was "the check that matters". It wasn't the only one. I had a PostgreSQL instance available and should have grepped `test/integration/` for every spec touching the files I'd changed.

Beyond the repair: `externalShareTransfersSql` names `$1..$4` and is bound with `settledParams.slice(0, 4)` — the identical hazard the spec was written for. A third case now asserts the binding is exactly four long and ends in the account scope, not merely that the shape matches.

### Both gates, on the current head

| Layer | Result |
|---|---|
| backend `lint`, `tsc --noEmit`, `typecheck`, `i18n:check` | pass |
| backend `TZ=UTC test:unit --coverage` | 699 suites, 17380 passed / 2 skipped; 95.83 / 87.32 / 96.23 / 96.76 |
| backend integration, real PostgreSQL 16 | 5 suites, 16 tests — `unmeasured-flows`, `invested-capital-flow`, `sold-out-security-series`, `daily-cash-series-dates`, `manual-price-snapshot-recovery` |
| frontend `type-check`, `lint`, `i18n:check`, parity (1574) | pass |
| frontend `test:cov` | 912 files, 17624 tests; 91.48 / 86.15 / 88.98 / 92.70 |
| frontend `build` | compiled successfully |

Both full-suite runs completed on `d0913722`. `54eaa9ee` changes only `backend/test/integration/`, which neither the frontend gate nor `test:unit` (`src/**/*.spec.ts`) reads; `typecheck`, which does cover `test/`, passed after that edit. CI on `54eaa9ee` is the authority.

**"New behavior has tests, and the existing suite passes" is now earned** — that box stands, having been retracted twice above.

One earlier failure was not real: `two-factor.service.spec.ts` failed in a run where I had two coverage suites competing for CPU in the same container, and a different suite failed in the other parallel run. It passes alone (66 tests) and passed in the serial gate above. Not a flake claim — a resource-contention artefact of how I ran it, which CI does not reproduce.

---

Issue #1214, opened 2026-08-18
Monize should treat blank payees specially: resolve action at time of display, rather than persisting the action at time of creation


bcorob:
When creating a transfer transaction, it may not make sense to specify the payee.  Often the payee is just "self," but the important thing for the user is not **who** the money is going to but **where** the money is going.  Since the user is already specifying transfer targets, there's not much need to specify an explicit payee.  But it's still helpful and expected to see the resultant action in the payee column when viewing lists of transactions.

When entering a transfer transaction directly, if the payee is left blank, Monize will auto-fill the payee with a description like "Transfer to Savings."  It actually appears to stamp different payees on each leg of the transfer, so one leg will show "Transfer to Savings" and one will show "Transfer from Checking."  This produces an immediately intuitive result.

However, because these fields are written directly, this text information is effectively "stamped" into the transaction forever and it's very difficult for Monize to adjust later.  If the user subsequently renames one of the accounts (for example, from "Savings" to "Long-term Savings"), all the transactions will now show stale information in the payee field, showing "Transfer to Savings" instead of "Transfer to Long-term Savings."

Stamping information into the transaction can also affect situations where a user changes display languages (for example, if spouses sharing information each prefer different primary languages, or one person takes over records from a person who prefers a different language).  In this case, any information stamped into the transaction can't easily be shown in the user's desired language.

All of this is an argument for Monize to persist blank payees as blank when storing data, and to dynamically resolve what text to display when the transaction is rendered.  This allows a blank payee to be resolved reflecting the user's current language, current account names and display preferences, and Monize can smartly determine the best way to render this info in the payee column when in various situations:

- on phone/small screens, it could show "xfer to chk." or similar abbreviated text
- on normal screens, it could show "Transfer to Checking"
- at the user's option or when room allows, more information can be shown, such as the institution or last 4 of the account number: "Transfer to First National Bank Checking *1234"

This same logic can apply to the payee field for investment transactions, where there are several different ways to render the details of the transaction with varying levels of detail.

kenlasko:
Good idea. Will do it for the next release after this one. I'm getting ready to shake'n'bake 1.15.0

PR #1428, opened 2026-09-22
feat(transactions): sort the register by any column, and keep the balance right both ways

### LINKED ISSUE #521: Allow sorting of transaction lists by any column
When looking at transactions, it's normal to show sorted by date, similar to a check register or statement.  However, the default sort mode seems to be with most recent at the top and cannot be toggled to most recent on the bottom.  I suggest allowing to click on a column heading to allow sorting by that field.  This should allow to toggle date sorting ascending/descending or to sort by another field such as payee as desired.

kenlasko:
## Linked discussion / issue

Closes #521
Approved in: https://github.com/kenlasko/monize/issues/521 (labelled `approved-to-build`; the KMyMoney rule -- hide the running balance under a non-date order -- was agreed there)

The design and task list this implements are `docs/future-plans/register-sorting.md` and `register-sorting-tasks.md`, merged in #1427.

## Summary

Every data column of the Transactions register sorts on a click: date, account, payee, category, description, ref #, amount, status. Date starts newest-first and reverses; every other column starts A to Z. The choice is remembered in the browser, like the row density and the table/calendar view. On a phone, where the card layout replaces the column header, the controls come back as a chip strip.

**The running balance is the hard part.** The server already accepted a sort, but the three queries that seed each page's balance hardcoded the newest-first order, and all four callers treated page 1 as the page holding the newest row. Sorting by date ascending would have put a wrong balance beside every row below the first page.

The seed keeps its meaning -- the balance after the page's newest row, walked newest-first -- and only two things became direction-aware, each written once: `restrictToRowsNewerThanPage` on the server and `walkRunningBalances` in the browser. The ascending order is the exact reverse of the descending one, amount tiebreak included, so the two windows name the same rows and the same figure lands beside the same row either way. `isNewestPage` replaces the `page === 1` shortcuts, which is why it takes the total.

Under any order but date the balance is **withheld rather than computed against a meaningless neighbour**, and the response says why (`startingBalanceWithheld: "sort"`), so the register can name the fix instead of leaving a reader to wonder where the column went. The flag is set only where a balance would otherwise have been returned, so an all-accounts unfiltered register is not promised one it never had.

The field list is declared once and mirrored, with a contract spec failing when the two layers differ: a field only the browser knows is a 400 behind a column header, and one only the server knows is a column nobody can sort by.

**Defects found while building it, fixed here** (they were written up as follow-ups first; the maintainer asked for them in this PR):

- The deep-link page counter had the target row counting **itself** as above itself: `created_at` is stored to the microsecond and read back into a millisecond JavaScript date, so the bound value was the row's own timestamp truncated. A deep link landed on the wrong page whenever that one row crossed a page boundary. It also compared on three keys where the register orders on four, and applied only some of the register's filters.
- The running balance ignored the status, currency, attachment and tag-key filters, so a filtered register drew a balance that counted rows the page does not show.
- The account CSV export hand-wrote a register order without the credits-before-debits leg.
- The register's today divider read the local clock instead of the shared financial-today helper.

Everything else is unchanged: six of the seven surfaces that draw this register pass no sort and render exactly as before, and a request without the new query parameters produces the order it always did.

## Invariants touched

- **INV-REGISTER-001** (new, `enforced`): a row's running balance is the same figure whichever way the date register runs, and no balance is shown under any other order.
- **INV-BALANCE-001**: read, not changed -- the seed is still a ledger sum under the shared movement predicate.
- **INV-TRANSFER-001**: read -- a VOID row still takes a line and moves nothing.

## Checklist

### What was run

| Gate | Result |
|---|---|
| Backend lint, `tsc`, `typecheck`, `i18n:check` | pass |
| Backend unit + coverage | 701 suites, 17,420 tests |
| Backend integration (real PostgreSQL 16) | 70 suites, 686 tests |
| Frontend lint, type-check, i18n:check | pass |
| Frontend tests + coverage | 914 files, 17,667 tests, thresholds met |
| Frontend build | compiled |
| `check-env-docs`, `check-docs-manifests`, `check-migration-prefixes` | pass |

Three load-bearing tests were mutation-checked: planting the naive "page 1 is always the newest page" rule, the wrong ascending window, and a dropped date-order gate each turn them red. No guard baseline, allowlist or threshold was widened -- when the new sort values pushed `list_transactions` over its `tools/list` byte budget, the prose restating the enum came out instead of the ceiling going up.

**Not run here:** the Playwright suite, which needs the docker-compose stack (Docker is unavailable in this environment). Its spec is included and CI is its gate.

## AI assistance disclosure

Written with Claude Code (Claude Opus 5), from the committed design and task list in `docs/future-plans/`. Every gate above was run in the session; the reasoning behind each decision is in the commit messages and the invariant entry.

https://claude.ai/code/session_012p9TdKVwJS48kPRVDKQpft

---
_Generated by [Claude Code](https://claude.ai/code/session_012p9TdKVwJS48kPRVDKQpft)_


kenlasko:
## Code review applied

I ran the repository's adversarial review protocol read-only against `acbb8d11`, then fixed what it found. Five commits: `f83c4e33`, `0c5af727`, `287464c3`, `47835648`, `3a6f9b56`.

### The one that mattered

**The register painted a running-balance column that was never true of the ledger.** `TransactionList` derived the walk, the today divider and the Balance column's visibility from the *live* `sort` prop, while `transactions` and `startingBalance` were still the previous request's payload. Click Date on a single account and the newest-first page was reversed under the new direction: a column reading 970/980/990/1000 became 1000/990/980/970, at full opacity with no spinner over it — `setIsLoading(false)` is permanent after the first load — and a reload that failed left it up for good. The page now adopts the order beside the rows and the seed it belongs to, the way it already adopted the withheld reason.

Two more from the same class: clicking Date while sorted by payee turned the column back on over rows whose seed was withheld, a strip of dashes under the line promising a balance; and the Date header's first click was a no-op where a remembered account sort had been resolved away, because the toggle read the stored sort rather than the displayed one.

### Also fixed

- **No request guard at all.** Two header clicks further apart than the 150 ms debounce raced, so amount-ordered rows could land under a Payee-active header. Reads are sequenced now; a superseded reply is discarded whole and raises no toast.
- **A deep link arriving in the URL cost two register reads, one of them wrong.** That path is not a filter change, so nothing coalesced the second read: it went out with the target already consumed, for the page the register was on, racing the page the first read had just resolved.
- **`categoryIds=["uncategorized"]` gave one row two different balances depending on direction.** The listing matches a split parent through its children; the balance's row set did not. That was a short total while the register only ran newest-first — it is an INV-REGISTER-001 breach now that the same set is what the ascending window offsets into while the boundaries come from the listing's total. The subquery's arm is the listing's arm now, both halves, and `computeSplitAwareSum` treats the pseudo-category as split-aware so the figure beside a split parent is the line the register hydrated, not the parent's whole amount. Proved in `register-sort.integration.spec.ts`; both halves fail on a planted revert.
- **A whitespace-only search** left the listing unfiltered but flipped the balance to the zero-based regime, so a 1000 opening balance vanished and the row read `-10` instead of `990`.
- **Balance deltas were never rounded.** Opening 0, `+0.09` then `+8.20`, page 2 → the API emitted `0.08999999999999986`.
- A stale or hand-edited `transactions.register.sort` was sent verbatim: a 400 on every read with no control left to clear it.
- `status` sinks its nulls now; the schema declares it defaulted, not `NOT NULL`.
- `hasContentFilters` is written once instead of twice; `RegisterRowFilters` members are required, so the compile error its comment promises is one a planted field actually produces.
- The sibling investments register still read the browser clock while its comment claimed to mirror `TransactionList`, so two TODAY dividers on one screen could fall on different rows.

### Guards and claims that were not holding what they said

- The hand-written-page-window scan allowed 400 characters after `select("t.id")`; the first of the three windows already sat 412 past its own. It scans the whole file now, and fails on a planted hand-rolled window.
- `AccountExportService` changed its ORDER BY in this branch and its spec asserted nothing about it — the mocked builder returns a fixed array, so the suite was as green before the change as after. It records the order calls now.
- Corrected: "only two things are direction-aware" (four), "any data column sorts" (tags, attachments and balance do not), a renamed helper still named in three doc lines, and a present-tense inventory line that this branch made false.

### Not changed, on purpose

- **The single-account balance windows carry neither the joint scope nor the brokerage exclusion.** Unreachable today — `resolveOwnContextJointScope` collapses the one case that would expose it — and widening `findAll`'s scope handling is a change of its own.
- **Nothing ties `RegisterPageWindow.total` to the row set the ascending offset is applied to.** The coupling is prose; a guard for it is a new invariant test, not a fix to this PR.
- **The Payee sort sinks every transfer to the end**, because a transfer leg stores `payee_name` as NULL by design (#1214) and the displayed label cannot be an ORDER BY key on the DISTINCT-ids pagination path. Documented rather than worked around.
- Widening the layer-wide running-balance scan to match case-insensitively names `buildRunningBalanceMap` in the loan payment detector, which walks a balance over an array it does not order itself. Reported here rather than pulled into this PR.

### Verified

- `backend/`: lint, `tsc --noEmit`, `typecheck`, `TZ=UTC test:unit --coverage` — 701 suites, 17,424 tests, 95.81/87.31/96.19/96.74.
- `backend/` integration against a real PostgreSQL 16: 69 of 70 suites, 686 tests. The one failure is `postgres-event-bus.integration.spec.ts`, untouched by this branch, which opens its own client from `DATABASE_*` env vars this container does not set.
- `frontend/`: lint, type-check, i18n:check, `test:cov` — 914 files, 17,680 tests, 91.49/86.18/89.00/92.71 — and `build`.
- Repo scripts: `check-docs-manifests.mjs`, `check-env-docs.mjs`, plus `doc-paths`, `source-comment-paths` and `instruction-files` with the new files staged.
- Every fix mutation-checked: the fix reverted, the new test seen to fail, the fix restored.

**Not executed here:** the Playwright suite. This environment has no Docker and no `e2e/node_modules`, so CI is its gate. The two cases were changed to stop clicking the Date header at its centre — the year toggle sits there and its wrapper stops the event before it reaches the header, so whether the click sorts depends on how wide the rendered dates are — and to scope their row locators to the register's own table.

---

kenlasko:
## E2E: root cause found, and it was the test

`32cdbab5`. Across the whole CI run on `3a6f9b56`, exactly one test failed, on both browsers — one I wrote in this branch:

`tests/transactions.spec.ts › Register sorting › shows a row the same balance in either date direction`

Shard 2 (chromium): 96 passed, that one failed. Shard 4 (firefox, cluster): 84 passed, that one failed. Everything else is green — backend unit and integration, both lint/typecheck jobs, frontend unit, bundle size, CodeQL, Bearer, zizmor, hadolint, Helm, licences, docs-vs-manifests, schema drift, Lighthouse, and E2E shards 1 and 3.

### What was actually wrong

| Head | The click | Result |
|---|---|---|
| `759b892a`, `acbb8d11` | `getByRole('columnheader', {name: /^date/i}).click()` — Playwright clicks the cell's **centre** | Click landed, `aria-sort` stayed `"descending"`. 3 attempts, both browsers. |
| `3a6f9b56` | `dateHeader.getByText('Date', {exact: true}).click()` | Locator never resolved; `locator.click` timed out at 60s. 3 attempts, both browsers. |

The first row confirms the review finding I had recorded as a width-dependent risk: it is not width-dependent, it is deterministic. The Date header holds the year toggle in `SortableHeader`'s `controls` slot, and that wrapper calls `stopPropagation()` by design so pressing the toggle does not also sort — and the centre of the cell is where the toggle sits. Every click went into it.

The second row is my fix attempt, which was worse: the test did not run at all. I could not work out from here why `getByText('Date', { exact: true })` finds nothing — the label is a plain text node in the header's flex row, and Playwright's `elementText` reads raw node values (no `text-transform` handling; `shouldSkipForTextMatching` skips only `SCRIPT`/`NOSCRIPT`/`STYLE`/`<head>`), so it ought to match. Rather than assert a mechanism I cannot demonstrate, this fix removes the question.

### The fix

`await dateHeader.press('Enter')` — no hit-testing and no text matching, which is the failure mode that bit twice. The `<th>` carries `tabIndex={0}` and `onKeyDown={activateOnKey(sort)}` for exactly this, `activateOnKey` handles Enter and Space, and the controls wrapper's key handler only stops events originating inside it. The mouse path stays covered by the Payee header in the case above, which holds no control, so both gestures are exercised on a header where each is unambiguous.

### Also in this commit

The two lint warnings CI reports, neither from this PR's diff:

- `investment-transactions.controller.spec.ts` captured the `TransferPairLinkService` mock in a variable no case reads; the mock is inlined. Worth noting separately: **no test in that file drives the two endpoints that reach that service** (`findCandidates`, `linkPair`). That is a real coverage gap, but it belongs to its own change, not this PR.
- `sw.js` had one `catch (error)` with the binding unused; renamed to `_error`, which the other four catch sites in that file already use.

### What I could not verify

**Playwright still cannot run here** — no Docker, no `e2e/node_modules` — so CI is the only gate, as it was for the two attempts that failed. This is a third attempt at a test I have never executed, and I would rather say that than imply otherwise.

Two things bound it: the change removes a dependency rather than adding one, and every premise rests on source I re-read (`SortableHeader.tsx:52,55`, `interactive-row.ts:69-77`). Once the activation works, this test's own assertions run for the first time — `aria-sort`, the leading row, and the two whole-row comparisons. If those fail, it is information about the feature rather than the locator; the arithmetic behind them is independently proved by `register-sort.integration.spec.ts`, which is green in CI and locally against a real PostgreSQL 16.

`payees.spec.ts › deletes a payee through the UI` was flaky on shard 4 and passed on retry. It is pre-existing and not this PR's; I have not touched it.

---

kenlasko:
## Backend Integration Tests on `32cdbab5`: a stall, not a slow test — and not a code change

The failure is a **timeout, not an assertion**:

```
register sorting (integration) › the balance counts exactly the rows the register shows
  › and the parity across directions survives those filters
thrown: "Exceeded timeout of 30000 ms for a test."
```

Everything else on this head is green. The four E2E shards are still running.

### It is not a code regression

`git diff 3a6f9b56 32cdbab5 -- backend/` is a single **unit** spec, which the integration config does not even match (`testRegex: "test/.*[.-]spec\\.ts$"` — `backend/test/`, not `backend/src/`). The integration spec and every line of backend source it exercises are **byte-identical** between the head where this job passed (`3a6f9b56`, ✅ 12:34) and the head where it timed out (`32cdbab5`, ❌ 13:19).

I also checked whether this PR's earlier backend work could have slowed *this* test. It cannot: it filters by `statuses` only, so `needsSplitsJoin` is false either way and `computeSplitAwareSum` takes its early-return branch exactly as before.

### My first theory was wrong, and the measurement killed it

I had assumed the file was running close to the 30s ceiling — several property tests that walk every page of a listing in both directions — and planned to trim them and raise the per-test timeout. Then I measured, locally against PostgreSQL 16, with Jest's JSON reporter:

| test | duration |
|---|---|
| under the uncategorised pseudo-category, which matches through splits | 1.01s |
| holds across page boundaries on an unfiltered account | 0.78s |
| lands on the page it is actually on, every row, both ways | 0.73s |
| never shows the account overdrawn on the day a transfer funded it | 0.60s |
| … | … |
| **total, all 26 tests** | **12.4s** |

The test that timed out is not in the top twelve — its body is under 0.45s. **A sub-second test did not creep past 30 000 ms; it stopped dead for thirty seconds.** So trimming work or raising the ceiling would have been cargo-cult: it would not have addressed the cause, and raising a timeout on a 0.3s test is the kind of threshold change that hides a real problem. I did not push it.

A second theory also died: I checked whether the `service.create` my new test performs leaves un-awaited background work that could collide with the next test's `TRUNCATE ... CASCADE` (`cleanTables` needs ACCESS EXCLUSIVE on every listed table, and its own comment warns about exactly this class of race). There is no fire-and-forget dispatch in that path — the create is fully awaited.

### What I have not established

The actual blocking mechanism. The suite is `--runInBand`, 70 spec files in one process against a shared `monize_test`, and the run's PostgreSQL log carries two `unexpected EOF on client connection with an open transaction` entries — a client dying mid-transaction, which holds locks until the backend is reaped. That is consistent with a stall, but I cannot tie it to this test, and I will not write it up as the cause on that basis.

### What I am doing

- **Not pushing a speculative fix.** Identical code, one pass, one stall; nothing I could change here is supported by evidence.
- **One re-run**, which is what the evidence justifies. The API refuses it while the run has jobs in flight (`403 This workflow is already running`), so it is queued behind the E2E shards.
- Keeping the PR watched. If this recurs, it is a genuine flake in the shared integration harness rather than in this PR's diff, and it deserves its own investigation with `--detectOpenHandles` and a `lock_timeout` — not a timeout bump here.

---

kenlasko:
## Integration stall: root-caused and fixed

`9b9ba749`. Third reading of this failure, and this one is measured rather than argued.

### What it actually was

**An un-awaited database write outliving its request, against a pool of ten that waits forever for an eleventh connection.**

1. Every one of the ~20 `actionHistoryService.record(...)` call sites invokes it **without `await`** (`transactions.service.ts:545` → `recordTransactionAction`, declared `: void`, → `:3379`). `await …record(` appears zero times in `src/`. `record` then runs *two* transactions — the redo-stack DELETE plus the INSERT, then `pruneUserHistory`.
2. `INTEGRATION_TYPEORM_OPTIONS` set **no pool options at all**, so node-postgres defaults applied: `max: 10`, and `connectionTimeoutMillis: 0` — *wait forever*. With `testTimeout: 30000`, "forever" surfaces as exactly a 30-second stall with nothing naming the pool. That is why my first two readings were wrong.
3. `register-sort.integration.spec.ts` rebuilds its fixture per test — a user, two accounts, two categories and ~8 `service.create` calls — each firing a history chain that takes a connection twice. The test body then queues behind them.

My own change made this file likelier to hit it: the uncategorised test added in the review commit is the only one that writes inside a test body, and it sits immediately before the test that timed out.

### The evidence, not the argument

The repo had already solved this exact problem for price writes, and `cleanTables` says so: a fetch nobody awaits, whose write lands under a truncate and returns an orphan foreign key — *"Both look like a bug in the test rather than in its housekeeping."* The same race was **already damaging every run**, visible in CI's Postgres log as `action_history … violates foreign key constraint … Key (user_id)=… is not present in table "users"`.

I reproduced it locally and measured the full integration suite before and after, same machine:

| | before | after |
|---|---|---|
| `action_history` FK violations | **5** | **0** |
| all FK violations | 6 | 1 |
| per-test durations (register-sort) | 12.4s / 26 tests | 12.9s / 26 tests |

The one remaining FK violation is a deliberate negative test of an unknown currency code (`RCE`), not an orphan write. The four `unexpected EOF on client connection with an open transaction` entries are unchanged at four — they come from the event-bus suite dropping its own connections on purpose. **Neither is this race and neither is claimed as fixed.**

### The fix

Follows the existing pattern rather than inventing one — a module-level pending set and `settlePendingHistoryWrites()`, mirroring `settlePendingPriceWrites` down to the reasoning in its comment. `record` becomes a thin wrapper registering its promise, so all ~20 call sites are covered without editing one, and **production keeps exactly the fire-and-forget behaviour it had**. `cleanTables` drains it beside the price writes; the register-sort fixture drains it once more after writing.

The pool also gains `max: 20` and a finite `connectionTimeoutMillis`, so a future burst cannot starve it silently: exhaustion now fails as `timeout exceeded when trying to connect`.

A regression test in `action-history.service.spec.ts` holds a `record()` in flight and asserts `settlePendingHistoryWrites()` does not resolve until it settles — mutation-checked by dropping the tracking, which turns it red.

### Not changed

- **The production fire-and-forget.** Deliberate, documented at `action-history.service.ts:322-331`, held by `derived-state-writers.guard.spec.ts`. Awaiting it would change request latency and is a separate proposal.
- **`testTimeout: 30000`.** Never the problem — tests run in about a second.
- No test skipped, weakened, or given a longer budget.

### Honest limit

I could not reproduce the 30-second stall itself on demand, so pool starvation is the mechanism the evidence points to rather than one I watched happen. What *is* proven is the shared cause — history writes landing after their teardown — reproduced locally and driven to zero. And if a stall does recur, the finite acquire timeout means the next failure names the pool instead of being silent.

---

